PR #13, opened 2023-01-23
Support Permisionless warp deployments

jmrossy:


nambrot (approved):
Chatted offline about the harmonization of the config schemas and keying by chain name

PR #14, opened 2023-01-29
Add gas payment to transferRemote and upgrade to SDK 1.2.0

jmrossy:
Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/1680

- Update to SDK 1.2.0
- Use gas payment in warp route
- Fix minor UI bugs


src/features/transfer/useTokenTransfer.ts
```diff
@@ -83,16 +84,20 @@ export function useTokenTransfer(onDone?: () => void) {
...
           ? getHypErc20CollateralContract(tokenRoute.hypCollateralAddress, provider)
           : getHypErc20Contract(tokenRoute.sourceTokenAddress, provider);
+        const gasPayment = await hypErc20.quoteGasPayment(destinationChainId);
         const transferTxRequest = await hypErc20.populateTransaction.transferRemote(
           destinationChainId,
           utils.addressToBytes32(recipientAddress),
           weiAmount,
+          {
+            value: gasPayment,
+          },
```
yorhodes, line 101: should the SDK generally be returning results from `populateTransaction`?

jmrossy: Would be good yeah. Having both options (`prepareTransaction` and `sendTransaction`) would be ideal

src/features/transfer/useTokenTransfer.ts
```diff
@@ -83,16 +84,20 @@ export function useTokenTransfer(onDone?: () => void) {
         const hypErc20 = isNativeToRemote
           ? getHypErc20CollateralContract(tokenRoute.hypCollateralAddress, provider)
           : getHypErc20Contract(tokenRoute.sourceTokenAddress, provider);
+        const gasPayment = await hypErc20.quoteGasPayment(destinationChainId);
```
nambrot, line 94: Flagging here that this method does not exist on the audited method, so we will probably want to make another PR for that version and then tag that as the audit-compliant version of the UI

PR #20, opened 2023-03-24
Validate JSON configs before use

jmrossy:
- Validate chain and token configs with Zod
- Use MultiProvider for provider management

Part of https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/1863


src/features/chains/ChainSelectModal.tsx
```diff
@@ -41,54 +43,3 @@ export function ChainSelectListModal({
     </Modal>
   );
 }
-
-// TODO update to support dynamic chain lists
```
jmrossy, line 45: I was keeping this cause I suspected it would be useful later but it ended up confusing a junior dev who was looking through code, so just cutting

src/features/chains/metadata.ts
```diff
@@ -1,61 +1,41 @@
...
-
-export const chainIdToCustomConfig = Object.values(CustomChainConfig).reduce<
-  Record<number, CustomChainMetadata>
->((result, config) => {
-  result[config.chainId] = config as CustomChainMetadata;
-  return result;
-}, {});
-
-// TODO use MultiProvider here
-export function getChainMetadata(chainId: number): ChainMetadata {
```
jmrossy, line 24: MultiProvider does all this stuff for us now :)

nambrot (approved):
TBH, a bit over my head, but lgtm. Would be nice to be able to run a yarn script so that this doesn't show at runtime, but at build time or something?

PR #23, opened 2023-04-12
Address some feedback from warp deployment

jmrossy:
- Make token logoURI optional
- Rename logoImgSrc to logoURI and make optional for chain config
- Use collateral chain as default from chain in form
- Reorganize routes.ts code

### DIFF
--- a/src/features/tokens/routes.ts
+++ b/src/features/tokens/routes.ts
@@ -10,7 +10,7 @@ import { getHypErc20CollateralContract } from '../contracts/hypErc20';
 import { getProvider } from '../multiProvider';
 
 import { getAllTokens } from './metadata';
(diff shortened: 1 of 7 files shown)


src/features/tokens/routes.ts
```diff
@@ -31,6 +31,70 @@ export interface Route {
 // Source chain to destination chain to Route
 export type RoutesMap = Record<number, Record<number, Route[]>>;
 
+export function useTokenRoutes() {
```
jmrossy, line 34: Just reorganizing here, only `useRouteChains` below has semantic changes

asaj (approved):
Not familiar with this codebase so mostly trusting you :)

PR #27, opened 2023-04-21
HypNative support

jmrossy:
- Add support for HypNative tokens
- Modify token list schema to reduce ambiguity
- Create global ChainId and DomainId type aliases for clarity 

Corresponds with https://github.com/hyperlane-xyz/hyperlane-deploy/pull/49

### DIFF
--- a/src/components/toast/TxSuccessToast.tsx
+++ b/src/components/toast/TxSuccessToast.tsx
@@ -2,7 +2,7 @@ import { toast } from 'react-toastify';
 
 import { getMultiProvider } from '../../features/multiProvider';
 
-export function toastTxSuccess(msg: string, txHash: string, chainId: number) {
+export function toastTxSuccess(msg: string, txHash: string, chainId: ChainId) {
   toast.success(<TxSuccessToast msg={msg} txHash={txHash} chainId={chainId} />, {
     autoClose: 12000,
   });
@@ -15,7 +15,7 @@ export function TxSuccessToast({
 }: {
   msg: string;
   txHash: string;
-  chainId: number;
+  chainId: ChainId;
 }) {
   const url = getMultiProvider().tryGetExplorerTxUrl(chainId, { hash: txHash });
   return (
--- a/src/features/contracts/hypErc20.ts
+++ b/src/features/contracts/hypErc20.ts
@@ -1,17 +1,46 @@
 import { Signer, providers } from 'ethers';
 
-import { HypERC20Collateral__factory, HypERC20__factory } from '@hyperlane-xyz/hyperlane-token';
+import {
+  HypERC20Collateral__factory,
+  HypERC20__factory,
+  HypNative__factory,
+  TokenType,
+} from '@hyperlane-xyz/hyperlane-token';
 
-export function getHypErc20CollateralContract(
+// Get the connected HypERC20Collateral, HypNative, or HypERC20 contract
+export function getHypWrapperContract(
+  type: TokenType,
+  contractAddress: Address,
+  signerOrProvider: Signer | providers.Provider,
+) {
+  if (type === TokenType.collateral) {
+    return getHypErc20CollateralContract(contractAddress, signerOrProvider);
+  } else if (type === TokenType.native) {
+    return getHypNativeContract(contractAddress, signerOrProvider);
+  } else if (type === TokenType.synthetic) {
+    return getHypErc20Contract(contractAddress, signerOrProvider);
+  } else {
+    throw new Error(`Unsupported token type: ${type}}`);
+  }
+}
+
+function getHypErc20CollateralContract(
   contractAddress: Address,
   signerOrProvider: Signer | providers.Provider,
 ) {
   return HypERC20Collateral__factory.connect(contractAddress, signerOrProvider);
 }
 
-export function getHypErc20Contract(
+function getHypErc20Contract(
   contractAddress: Address,
   signerOrProvider: Signer | providers.Provider,
 ) {
   return HypERC20__factory.connect(contractAddress, signerOrProvider);
+}
+
+function getHypNativeContract(
+  contractAddress: Address,
+  signerOrProvider: Signer | providers.Provider,
+) {
+  return HypNative__factory.connect(contractAddress, signerOrProvider);
 }
--- a/src/features/contracts/hypErc20.ts
+++ b/src/features/contracts/hypErc20.ts
@@ -8,7 +8,7 @@ import {
 } from '@hyperlane-xyz/hyperlane-token';
 
 // Get the connected HypERC20Collateral, HypNative, or HypERC20 contract
-export function getHypWrapperContract(
+export function getTokenRouterContract(
   type: TokenType,
   contractAddress: Address,
   signerOrProvider: Signer | providers.Provider,
--- a/src/features/tokens/metadata.ts
+++ b/src/features/tokens/metadata.ts
@@ -1,25 +1,48 @@
+import { ethers } from 'ethers';
+
+import { TokenType } from '@hyperlane-xyz/hyperlane-token';
+
 import { tokenList } from '../../consts/tokens';
-import { areAddressesEqual } from '../../utils/addresses';
 import { logger } from '../../utils/logger';
 
-import { TokenListSchema, TokenMetadata } from './types';
+import { TokenMetadata, WarpTokenConfig, WarpTokenConfigSchema } from './types';
 
 let tokens: TokenMetadata[];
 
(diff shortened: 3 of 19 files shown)


src/components/toast/TxSuccessToast.tsx
```diff
@@ -2,7 +2,7 @@ import { toast } from 'react-toastify';
 
 import { getMultiProvider } from '../../features/multiProvider';
 
-export function toastTxSuccess(msg: string, txHash: string, chainId: number) {
+export function toastTxSuccess(msg: string, txHash: string, chainId: ChainId) {
```
asaj, line 5: Where is this coming from? I don't see it imported

src/features/contracts/hypErc20.ts
```diff
@@ -1,17 +1,46 @@
...
+import {
+  HypERC20Collateral__factory,
+  HypERC20__factory,
+  HypNative__factory,
+  TokenType,
+} from '@hyperlane-xyz/hyperlane-token';
 
-export function getHypErc20CollateralContract(
+// Get the connected HypERC20Collateral, HypNative, or HypERC20 contract
+export function getHypWrapperContract(
```
asaj: Should this be `getTokenRouterContract`? I believe that is what they all have in common

src/features/tokens/metadata.ts
```diff
@@ -1,25 +1,48 @@
...
+        type: TokenType.collateral,
+        hypWrapperAddress: token.hypCollateralAddress,
+        address: token.address,
+      });
+    } else if (type == TokenType.native) {
+      tokenMetadata.push({
+        ...commonFields,
+        type: TokenType.native,
+        hypWrapperAddress: token.hypNativeAddress,
+        // Note, using 0x000... address to help identify native tokens
```
asaj, line 42: Why is this necessary given `type`?

src/features/tokens/routes.ts
```diff
@@ -1,35 +1,36 @@
...
-import { getHypErc20CollateralContract } from '../contracts/hypErc20';
+import { getHypWrapperContract } from '../contracts/hypErc20';
 import { getProvider } from '../multiProvider';
 
 import { getAllTokens } from './metadata';
 import { TokenMetadata, TokenMetadataWithHypTokens } from './types';
 
 export enum RouteType {
-  NativeToRemote = 'nativeToRemote',
+  BaseToRemote = 'baseToRemote',
```
asaj: nit: `base` is new terminology to me, usually we talk about local/remote or origin/destination.

src/features/tokens/routes.ts
```diff
@@ -1,35 +1,36 @@
...
+  baseTokenAddress: Address;
+  hypWrapperAddress: Address;
   sourceTokenAddress: Address;
   destTokenAddress: Address;
   decimals: number;
 }
 
 // Source chain to destination chain to Route
-export type RoutesMap = Record<number, Record<number, Route[]>>;
+export type RoutesMap = Record<ChainId, Record<ChainId, Route[]>>;
```
asaj, line 33: nit: ditto above, s/source/origin

src/features/tokens/routes.ts
```diff
@@ -57,35 +58,36 @@ export function useTokenRoutes() {
...
-    );
-  }
-  const symbolOnChain = await erc20.symbol();
-  if (symbol !== symbolOnChain) {
-    throw new Error(
-      `Token config symbol ${symbol} does not match contract decimals ${symbolOnChain}`,
-    );
+  const wrapperContract = getHypWrapperContract(type, hypWrapperAddress, provider);
+
+  if (type === TokenType.collateral) {
```
asaj, line 66: Is this not necessary for native as well?

src/features/tokens/routes.ts
```diff
@@ -57,35 +58,36 @@ export function useTokenRoutes() {
...
   logger.info('Fetching connected domains');
-  const domains = await collateralContract.domains();
+  const domains = await wrapperContract.domains();
   logger.info(`Found ${domains.length} connected domains:`, domains);
 
   logger.info('Getting domain router address');
-  const hypTokenByteAddresses = await Promise.all(
-    domains.map((d) => collateralContract.routers(d)),
-  );
+  const hypTokenByteAddresses = await Promise.all(domains.map((d) => wrapperContract.routers(d)));
```
asaj: nit: hypTokenAddressesAsBytes32 or st

src/features/tokens/types.ts
```diff
@@ -1,22 +1,83 @@
...
+
+const CollateralTokenSchema = commonTokenFields.extend({
+  type: z.literal(TokenType.collateral),
+  address: z.string(),
+  hypCollateralAddress: z.string(),
+});
 
-export const TokenListSchema = z.array(TokenSchema);
+interface CollateralTokenConfig extends BaseTokenConfig {
+  type: TokenType.collateral | 'collateral';
```
asaj, line 37: I'm surprised we need to do this if the enum assigns `TokenType.collateral` to 'collateral'. Does it?

src/features/tokens/types.ts
```diff
@@ -1,22 +1,83 @@
...
+  address: Address;
+  hypWrapperAddress: Address; // Shared name for hypCollateralAddr and hypNativeAddr
+}
+
+interface CollateralTokenMetadata extends BaseTokenMetadata {
+  type: TokenType.collateral;
+}
+
+type ZeroAddress = `${typeof ethers.constants.AddressZero}`;
+interface NativeTokenMetadata extends BaseTokenMetadata {
```
asaj, line 73: This deserves a comment explaining what's going on here (as mentioned before, I don't understand why this is necessary)

src/features/tokens/useTokenBalance.tsx
```diff
@@ -46,11 +48,18 @@ export function getCachedTokenBalance(
...
-async function fetchTokenBalance(chainId: number, tokenAddress: Address, accountAddress: Address) {
-  logger.debug(
-    `Fetching balance for account ${accountAddress} token ${tokenAddress} on chain ${chainId}`,
-  );
-  const erc20 = getErc20Contract(tokenAddress, getProvider(chainId));
-  const balance = await erc20.balanceOf(accountAddress);
-  return balance.toString();
+async function fetchTokenBalance(chainId: ChainId, tokenAddress: Address, accountAddress: Address) {
+  if (isNativeToken(tokenAddress)) {
+    logger.debug(`Fetching balance for account ${accountAddress} native token on chain ${chainId}`);
```
asaj, line 53: nit: if a MP is available it would be more readable to log the chain name

src/global.d.ts
```diff
@@ -1 +1,3 @@
 declare type Address = string;
+declare type ChainId = number;
```
asaj, line 2: Ah, there we go, I didn't know you could do this.

asaj (approved):
New to this repo so apologies if I'm off the mark on some things

src/features/tokens/routes.ts
jmrossy: Yes but those terms are relative. I need a term that refers to the original source of the token. Totally open to renaming if you have a preferred alternative?

src/features/tokens/metadata.ts
jmrossy: It's not strictly necessary but I found that for other parts of the code (mostly UI code) it was helpful for *all* metadata objects to have an addresss

src/features/contracts/hypErc20.ts
jmrossy: That's what I meant by this yes. I'll rename

src/features/tokens/routes.ts
jmrossy: Okay will rename tmrw

src/features/tokens/routes.ts
jmrossy: What would we validate it against?

src/features/tokens/types.ts
jmrossy: Typescript won't allow a field of type `TokenType` to be the literal value 'collateral' even if the enum's values match up. And I'm determined to let devs use the literals in the configs to avoid them having to import `TokenType`. However I was not okay with loosening the type throughout the whole app. This separation of *Config and *Metadata types is the solution

src/features/tokens/types.ts
jmrossy: See above. I'll add a comment

src/features/tokens/useTokenBalance.tsx
jmrossy: this is just a log statement, not user facing

PR #32, opened 2023-07-11
Update to SDK 1.4.0

jmrossy:
- Update SDK, token, and widgets to 1.4.0
- Use SDK values for Solana chains
- Change format of Solana CAIP2 IDs to use numbers
- Fix bug with tx prep for Zebec chain


yorhodes (approved):
nice

PR #36, opened 2023-07-18
Token select fix for issue: 2514

ashu26jha:
This PR tries to fix the https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2514

Features of the PR
- Grays out not selectable
- Adds a tooltip mentioning the reason (Not suitable for source chain to destination chain) (Need to hover on `i`)
- Search bar works as it is

Video, demonstrating the changes is attested below, have a look

https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/assets/66910385/08f03af4-861e-40ca-b78a-45c796a2e757

### DIFF
--- a/src/features/tokens/TokenListModal.tsx
+++ b/src/features/tokens/TokenListModal.tsx
@@ -1,3 +1,4 @@
+import Image from 'next/image';
 import { useMemo, useState } from 'react';
 
 import { TokenType } from '@hyperlane-xyz/hyperlane-token';
@@ -5,6 +6,7 @@ import { TokenType } from '@hyperlane-xyz/hyperlane-token';
 import { TokenIcon } from '../../components/icons/TokenIcon';
 import { TextInput } from '../../components/input/TextField';
 import { Modal } from '../../components/layout/Modal';
+import InfoIcon from '../../images/icons/info-circle.svg';
 
 import { getAllTokens } from './metadata';
 import { isNativeToken } from './native';
@@ -83,16 +85,12 @@ export function TokenList({
     return getAllTokens()
       .map((t) => {
         const hasRoute = hasTokenRoute(originCaip2Id, destinationCaip2Id, t.address, tokenRoutes);
-        return { ...t, disabled: hasRoute };
+        return { ...t, disabled: !hasRoute };
       })
       .sort((a, b) => {
-        if (!a.disabled && b.disabled) {
-          return 1;
-        } else if (a.disabled && !b.disabled) {
-          return -1;
-        } else {
-          return 0;
-        }
+        if (a.disabled && !b.disabled) return 1;
+        else if (!a.disabled && b.disabled) return -1;
+        else return 0;
       })
       .filter((t) => {
         if (!q) return t;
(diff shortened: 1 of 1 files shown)


src/features/tokens/TokenListModal.tsx
```diff
@@ -79,28 +79,44 @@ export function TokenList({
...
       {tokens.length ? (
         tokens.map((t) => (
           <button
-            className="-mx-2 py-2 px-2 rounded hover:bg-gray-100 active:bg-gray-200 transition-all duration-250"
+            className={`-mx-2 py-2 px-2 rounded  ${
+              t.disabled ? 'hover:bg-gray-200' : 'bg-gray-100 '
+            } transition-all duration-250`}
             key={`${t.caip2Id}-${t.address}`}
             type="button"
+            disabled={!t.disabled}
```
jmrossy: This looks reversed? should the new disabled value be the ! of what is is now?

src/features/tokens/TokenListModal.tsx
```diff
@@ -79,28 +79,44 @@ export function TokenList({
...
   }, [searchQuery, originCaip2Id, destinationCaip2Id, tokenRoutes]);
 
   return (
     <div className="flex flex-col items-stretch divide-y divide-gray-100">
       {tokens.length ? (
         tokens.map((t) => (
           <button
-            className="-mx-2 py-2 px-2 rounded hover:bg-gray-100 active:bg-gray-200 transition-all duration-250"
+            className={`-mx-2 py-2 px-2 rounded  ${
+              t.disabled ? 'hover:bg-gray-200' : 'bg-gray-100 '
```
jmrossy: Let's please use a darker gray for the disabled background and also add a margin in between the token rows so there's a small amount of whitespace in between

src/features/tokens/TokenListModal.tsx
```diff
@@ -121,6 +137,18 @@ export function TokenList({
                   </span>
                 </div>
               </div>
+              {!t.disabled ? (
+                <a
```
jmrossy: Why are you using an anchor here? Please change this to an svg image of an info icon (see `info-circle.svg).

src/features/tokens/TokenListModal.tsx
```diff
@@ -121,6 +137,18 @@ export function TokenList({
...
+                  href="#"
+                  className="transititext-primary text-primary hover:text-primary-600 focus:text-primary-600 active:text-primary-700 dark:text-primary-400 dark:hover:text-primary-500 dark:focus:text-primary-500 dark:active:text-primary-600 ml-auto mr-1 cursor-default border-black"
+                  data-te-toggle="tooltip"
+                  title={`Route not supported for ${originCaip2Id} to ${destinationCaip2Id}`}
+                >
+                  <div className="rounded-full bg-gray-300 w-6">i</div>
+                </a>
+              ) : (
+                <></>
```
jmrossy: There's no need for the ternary operator and an empty fragment here. Change the condition above to ` {!t.disabled && (`

src/features/tokens/TokenListModal.tsx
```diff
@@ -79,28 +79,44 @@ export function TokenList({
...
+      })
+      .sort((a, b) => {
+        if (!a.disabled && b.disabled) {
+          return 1;
+        } else if (a.disabled && !b.disabled) {
+          return -1;
+        } else {
+          return 0;
+        }
+      })
```
jmrossy: Nit, can you tighten this up to be on fewer lines (e.g. `if (!a.disabled && b.disabled) return 1`)

jmrossy: Nit, can you tighten this up to be on fewer lines (e.g. `if (!a.disabled && b.disabled) return 1`)

jmrossy (commented):
Thanks for the PR! Please see comments below 🙂

src/features/tokens/TokenListModal.tsx
ashu26jha: Ah! I got it... It was working but the naming is confusing. Reversed this in new commit

src/features/tokens/TokenListModal.tsx
ashu26jha: Sure also made the disabled rows' text slight gray (Won't hurt eyes :P)

Issue #30, opened 2023-06-24
Persist transfer history to local storage


jmrossy:
**Current behavior:**

The tx history data stored in the zustand store is not persisted.

**Desired behavior:**

Persist and rehydrate the transfer state to browser local storage for a semi-durable history of the user's transfers from the UI. We must also expand the transfer data in the zustand store (and storage) to include the following (which will be required for #31):
1. The form values (chains, token, amount, recipient)
2. The the origin chain tx hash
3. The time + date

PR #37, opened 2023-07-18
Persist transfer history to local storage

iamdanylo:
Issue #30

### DIFF
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -1,5 +1,5 @@
-import { StateCreator, create } from 'zustand';
-import { createJSONStorage, persist } from 'zustand/middleware';
+import { create } from 'zustand';
+import { persist } from 'zustand/middleware';
 
 import type { TransferContext, TransferStatus } from './transfer/types';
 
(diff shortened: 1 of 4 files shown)


src/features/store.ts
```diff
@@ -53,4 +56,11 @@ export const useStore = create<AppState>()((set) => ({
...
   },
-}));
+});
+
+export const useStore = create<AppState>()(
+  persist(persistState, {
+    name: 'app-state',
+    storage: createJSONStorage(() => localStorage),
+  }),
+);
```
jmrossy: A few items here:
1. Can you restructure the code to be organized like here: https://github.com/hyperlane-xyz/hyperlane-explorer/blob/a3e422016988fba77d74dc25ca8e26c9eb1c0ff7/src/store.ts#L37
2. Please add a `partialize` option to only persist the transfers
3. Is the `storage` field necessary? I thought it used JSON by default?

iamdanylo: 1) Sure
2) Ok
3) Yeah I missed that, that is optional field

jmrossy (approved):

Issue #31, opened 2023-06-24
Create new wallet flyout menu


jmrossy:
**Current behavior:**

The transfer history is somewhat accessible from the transfer modal (`TransfersStatusModal`)
The `WalletControlBar` and `TransfersStatusBar` are two separate dropdown components in the `Header`.
The `WalletControlBar` is always visible but the `TransfersStatusBar` shows once a transfer has been initiated.

**Desired behavior:**

Merge `TransfersStatusBar` and `WalletControlBar` into a new flyout menu similar to the wallet menu on app.uniswap.org

See mockups below or the [Figma file here](https://www.figma.com/file/4wCujIAQuunD6NTyvCXadA/Token-Bridge-Mockups?type=design&node-id=204%3A928&mode=design&t=hw21CXTOZsFxOCbO-1).

*Note*: Ignore the fonts in the Figma/mockups, just a bug. Use the existing fonts in the tailwind theme.

![Wallet flyout](https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/assets/710751/b47b2d60-3f4d-4061-96f7-fb11786089da)

![Tx modal](https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/assets/710751/5a0b56d8-d8c7-4171-b249-963bdbdf2d93)

PR #39, opened 2023-07-21
New wallet flyout menu #31

iamdanylo:
#31


src/components/nav/Header.tsx
```diff
@@ -19,6 +21,7 @@ export function Header() {
         <div className="flex flex-col items-end md:flex-row-reverse md:items-start gap-2">
           <WalletControlBar />
           <TransfersStatusBar />
+          <SideBarMenu isOpen={true} />
```
jmrossy: I guess you'll want to manage the open state either with some local state in this component, passing the setter down to the WalletControlBar, or with a flag in the zustand store. Whatever you prefer is fine

src/features/transfer/components/BasicSpinner.tsx
```diff
@@ -0,0 +1,39 @@
+import Image from 'next/image';
+
+import { Spinner } from '../../../components/animation/Spinner';
+import CheckmarkCircleIcon from '../../../images/icons/checkmark-circle.svg';
+import EnvelopeHeartIcon from '../../../images/icons/envelope-heart.svg';
+import ErrorCircleIcon from '../../../images/icons/error-circle.svg';
+import { TransferStatus } from '../types';
+
+export function BasicSpinner({ transferStatus }: { transferStatus: TransferStatus }) {
```
jmrossy: Let's rename this. How about TransferStatusIcon?

src/features/transfer/helpers.ts
```diff
@@ -0,0 +1,59 @@
+import { links } from '../../consts/links';
```
jmrossy: Nit: our convention is typically to name files like this `utils.ts`

jmrossy (commented):
Gave it a quick review, no major issues found. I think you're going in the right direction, keep it up!

jmrossy (approved):
I made a few final fixes and style tweaks, should be good to!

Issue #33, opened 2023-07-14
Chain and Token fields should be smarter


jmrossy:
**Current behavior:**

Chain fields: When a user selects a from/to chain, the other field still allows them to select any other chain, even if there's no warp route with that combination.

Token field: The token field shows only the tokens for the selected chains.

**Desired behavior:**

Chain fields: When one chain field set, the other field shows still shows all chains but invalid options are disabled (greyed out).

Token field: The token field always shows all tokens but invalid options (based on to/from chains) are disabled (greyed out)

ashu26jha:
@jmrossy I guess I worked on the Token fields https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/36 should I go on and work on `chain fields`? Also how would I get around the `Rainbow Kit UI project ID` without it the build fails.

Thank you

jmrossy:
Yes, we can finish your #36 PR first and then do the chain field change in a separate PR :)

Regarding the project id error, create an `.env.local` file using the value in the `.env.example` file

ashu26jha:
@jmrossy Made some changes, have a look. Also why is the build failing? 🥲

jmrossy:
The build was failing because your github account doesn't have permissions to access the secrets required to execute the build. 

I've tested, tweaked, and merged your changes through intermediate PR https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/38. Thanks @ashu26jha for the contribution!

ashu26jha:
Will make a separate PR for `chain fileds` 👍

jmrossy:
@ashu26jha would you still like to handle this task? No problem if you're busy, another dev can pick it up :)

ashu26jha:
Hey really sorry, currently busy with uni exams.
Apologies!

jmrossy:
> Hey really sorry, currently busy with uni exams. Apologies!

Totally understandable. Good luck!

jmrossy:
After some UX discussion, it was decided leave the chain fields as they are. Closing!

PR #41, opened 2023-08-08
Chain selection flow updated, disabled chain item state added

iamdanylo:
https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/issues/33


jmrossy:
Stale

PR #43, opened 2023-08-13
Setup Eclipse tokens and chains

jmrossy:


daniel-savu (approved):

jmrossy:
@daniel-savu FYI this PR isn't intended it merge, it's just to show the config changes we have in place for eclipse. May be best to dismiss your review so no one accidentally merges eclipse-specific stuff into main :)

daniel-savu (changes requested):
Requesting changes to prevent merging since that's not the intention of this PR

PR #51, opened 2023-08-22
Improve multi-collateral support

jmrossy:
- Setup Jest unit testing
- Create new CollateralToCollateral route type
- Refactor route computation into separate file
- Add unit tests for route computation
- Improve hacky fixes collateral balance checking
- Implement approval allowance checking

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2548
Helps with https://github.com/hyperlane-xyz/issues/issues/554


src/features/tokens/routes/fetch.ts
```diff
@@ -0,0 +1,159 @@
+import { ProtocolType } from '@hyperlane-xyz/sdk';
```
jmrossy, line 1: Mostly just moved here from `routes/hooks.ts` with a few tweaks to have it avoid creating duplicate routes between collateral tokens

PR #54, opened 2023-09-05
More nautilus warps

tkporter:
* Adds warps for ETH, USDC, USDT, BTC, and POSE all bridged from BSC
* Tested sending 1 wei of each of these. I can send you tokens for testing
* Fixed a couple bugs I spotted to do with trying to bridge really low values - e.g. 1e-7 would be formatted in scientific notation and things would throw, and then also make it easier to see balances that are super tiny (don't feel strongly about merging that tbh, just was useful for testing things worked)

Still needs some styling of things


tkporter:
related https://github.com/hyperlane-xyz/issues/issues/566

jmrossy (approved):

PR #59, opened 2023-09-26
Add support for a transfer route blacklist

jmrossy:
- Add support for route blacklist via NEXT_PUBLIC_TRANSFER_BLACKLIST env var
- Fix outdated CAIP2 example comment


src/consts/config.ts
```diff
@@ -11,7 +12,8 @@ interface Config {
   showTipBox: boolean; // Show/Hide the blue tip box above the transfer form
   showDisabledTokens: boolean; // Show/Hide invalid token options in the selection modal
   walletConnectProjectId: string; // Project ID provided by walletconnect
-  withdrawalWhitelist: string; // comma-separated list of CAIP2 domains to which transfers are supported
+  withdrawalWhitelist: string; // comma-separated list of CAIP2 chain IDs to which transfers are supported
```
nambrot, line 15: NIt: Can we just make them similar? They feel just different enough

nambrot (approved):

src/consts/config.ts
tkporter: idk I think both serve their purpose well enough as they are, i'm gonna merge so we can ship the change but am not opposed to changing

tkporter (approved):
ty 🙏

Issue #61, opened 2023-09-27
IGP fees displayed in UI


tkporter:
- Atm, there's no indication in the UI what the IGP fees will be / what they are
- Originally, if IGP fees exceeded the user balance, then it would just say that some error occurred when signing. As a hack in https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/60, I put a check when submitting a tx that would show a very ugly toast with some basic info
- We should add a section in the IGP in the review tx details card
- This is most important for the EVM. We should still have the ability to support Sealevel ofc via adapters, but for now we can consider just hardcoding an estimated amount

PR #62, opened 2023-09-30
Implement IGP fee details and validation

jmrossy:
Fixes https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/issues/61

Supersedes https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/60 which should be removed from eclipse-mainnet 

Screenshot 2023-09-30 at 11 40 15 AM
Screenshot 2023-09-30 at 12 41 22 PM

### DIFF
--- a/src/features/transfer/TransferTokenForm.tsx
+++ b/src/features/transfer/TransferTokenForm.tsx
@@ -28,11 +28,13 @@ import { getProtocolType } from '../caip/chains';
 import { getTokenAddress, isNonFungibleToken, parseCaip19Id } from '../caip/tokens';
 import { ChainSelectField } from '../chains/ChainSelectField';
 import { getChainDisplayName } from '../chains/utils';
+import { getChainMetadata } from '../multiProvider';
 import { AppState, useStore } from '../store';
 import { SelectOrInputTokenIds } from '../tokens/SelectOrInputTokenIds';
 import { TokenSelectField } from '../tokens/TokenSelectField';
 import { useIsApproveRequired } from '../tokens/approval';
 import { useDestinationBalance, useOriginBalance } from '../tokens/balances';
+import { getToken } from '../tokens/metadata';
 import { useRouteChains } from '../tokens/routes/hooks';
 import { RoutesMap } from '../tokens/routes/types';
 import { getTokenRoute, isRouteFromNative } from '../tokens/routes/utils';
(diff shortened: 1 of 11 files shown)


src/features/transfer/TransferTokenForm.tsx
```diff
@@ -430,11 +443,26 @@ function ReviewDetails({ visible, tokenRoutes }: { visible: boolean; tokenRoutes
...
+                </p>
               ) : (
-                <p>{`Amount (${originUnitName}): ${sendValue}`}</p>
+                <>
+                  <p className="flex">
+                    <span className="min-w-[7rem]">{`Amount (${originUnitName})`}</span>
+                    <span>{sendValue}</span>
+                  </p>
+                  <p className="flex">
+                    <span className="min-w-[7rem]">{`Interchain Gas (${originUnitName})`}</span>
```
tkporter, line 466: If we're bridging an ERC-20 from Ethereum, I don't think it's super clear that the interchain gas is denominated in ETH but the "Amount" on line 458 is denominated in the ERC-20 token

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -430,11 +443,26 @@ function ReviewDetails({ visible, tokenRoutes }: { visible: boolean; tokenRoutes
...
+                    <span>{igpQuote?.weiAmount || '0'}</span>
```
tkporter: Thoughts on formatting this based off the # of decimals? I think the "Amount" above is fair to have in wei because you already will have typed in the decimal-based amount to get to this point, but because people may not be familiar with interchain gas / what wei means, I wonder if this should be formatted

tkporter (approved):
I was originally imagining between the "RECIPIENT ADDRESS" grey box and the "TRANSACTIONS" header that there would be a new section, like "GAS FEES" with its own grey box indicating the gas fees (could show both interchain gas & origin gas). Happy to merge as is though, I don't feel super strongly though, I think this PR is enough to make it more clear to users how much in fees they'll pay & why a tx may fail if they don't have enough of a balance

src/features/transfer/TransferTokenForm.tsx
jmrossy: I'll add the symbol names to the amounts

src/features/transfer/TransferTokenForm.tsx
jmrossy: I like wei for this because the relative number of digits give you a sense of the scale for gas fees. We could have both wei and units for both though

PR #63, opened 2023-10-02
Update Eclipse mainnet with IGP fee details

jmrossy:


jmrossy:
Tested, looks good, merging

PR #66, opened 2023-10-29
Handle IBC denoms

nambrot:


src/consts/tokens.ts
```diff
@@ -38,12 +38,13 @@ export const tokenList: WarpTokenConfig = [
 
   // Neutron token
   {
-    type: 'native',
+    type: 'collateral',
```
jmrossy, line 41: Does this work? The CwHypCollateral adapter in Yorke's branch is not yet implemented.

src/features/caip/tokens.ts
```diff
@@ -100,11 +106,14 @@ export function resolveAssetNamespace(
...
-    return isSpl2022 ? AssetNamespace.spl2022 : AssetNamespace.spl;
-  } else {
-    throw new Error(`Unsupported protocol: ${protocol}`);
+  switch (protocol) {
+    case ProtocolType.Ethereum:
+      return isNft ? AssetNamespace.erc721 : AssetNamespace.erc20;
+    case ProtocolType.Sealevel:
+      return isSpl2022 ? AssetNamespace.spl2022 : AssetNamespace.spl;
+    case ProtocolType.Cosmos:
+      return AssetNamespace.ibcDenom;
```
jmrossy, line 115: Will we only support ibc denom tokens for cosmos or is this temporary. If so please leave a `TODO cosmos` comment so I can keep track of what needs to change

jmrossy (approved):

PR #69, opened 2023-11-02
IBC Route support

jmrossy:
Fixes https://github.com/hyperlane-xyz/issues/issues/682

- Implement partial support for IBC tokens + routes
- Improve celestia logo
- Improve token select fields


src/features/tokens/CosmNativeTokenAdapter.ts
```diff
@@ -0,0 +1,135 @@
+import { MsgTransferEncodeObject } from '@cosmjs/stargate';
```
jmrossy: I'll move the latest version of all this to the SDK this week

src/features/transfer/utils.ts
```diff
@@ -0,0 +1,57 @@
...
+import { getProtocolType } from '../caip/chains';
+import { isNonFungibleToken } from '../caip/tokens';
+import { AdapterFactory } from '../tokens/AdapterFactory';
+import { Route } from '../tokens/routes/types';
+import { isHypRoute, isRouteToCollateral } from '../tokens/routes/utils';
+
+// In certain cases, like when a synthetic token has >1 collateral tokens
+// it's possible that the collateral contract balance is insufficient to
+// cover the remote transfer. This ensures the balance is sufficient or throws.
+export async function ensureSufficientCollateral(route: Route, weiAmount: string) {
```
jmrossy, line 17: Contents of this file aren't new. Moved here from `useTokenTransfer.ts` b.c. it's too big

src/consts/ibcRoutes.ts
```diff
@@ -0,0 +1,54 @@
...
+    destCaip2Id: 'cosmos:neutron-1',
+    destDecimals: 6,
+    originIbcDenom: 'utia',
+    sourcePort: 'transfer',
+    sourceChannel: 'channel-8',
+    // IBC Tia on Neutron
+    derivedIbcDenom: 'ibc/773B4D0A3CD667B2275D5A4A7A2F0909C0BA0F4059C0B9181E680DDF4965DCC7',
+  },
+  {
+    type: RouteType.IbcNativeToHypNative,
```
nambrot: Why is it `HypNative`?

src/features/tokens/CosmNativeTokenAdapter.ts
```diff
@@ -0,0 +1,135 @@
...
+    return this.multiProvider.getCosmJsProvider(this.chainName);
+  }
+}
+
+export class CosmNativeTokenAdapter extends BaseCosmosAdapter implements ITokenAdapter {
+  constructor(
+    public readonly chainName: ChainName,
+    public readonly multiProvider: MultiProtocolProvider,
+    public readonly addresses: Record<string, Address>,
+    public readonly properties: {
```
nambrot: Nit: Maybe call them `ibcProperties` or something

src/consts/chains.ts
```diff
@@ -47,6 +47,25 @@ export const chains: ChainMap<ChainMetadata & { mailbox?: Address }> = {
     ...chainMetadata.solanadevnet,
     mailbox: '4v25Dz9RccqUrTzmfHzJMsjd1iVoNrWzeJ4o6GYuJrVn',
   },
+  cosmoshub: {
```
yorhodes, line 50: are we using this?

src/consts/values.ts
```diff
@@ -7,3 +7,5 @@ export const COSMOS_ZERO_ADDRESS = 'cosmos10000000000000000000000000000000000000
 export const SOL_SPL_NOOP_ADDRESS = 'noopb9bkMVfRPU8AsbpTUg8AQkHtKwMYZiFUjNRtMmV';
 export const SOL_IGP_QUOTE = '10000';
 export const COSM_IGP_QUOTE = '100';
+// Placeholder chain for use with cosmos-kit wallet hooks that require a chain name
+export const PLACEHOLDER_COSMOS_CHAIN = 'cosmoshub';
```
yorhodes, line 11: maybe celestia for usenexus?

src/features/tokens/CosmNativeTokenAdapter.ts
```diff
@@ -0,0 +1,135 @@
...
+  getMetadata(): Promise<MinimalTokenMetadata> {
+    throw new Error('Metadata not available to native tokens');
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  populateApproveTx(_transferParams: TransferParams): unknown {
+    throw new Error('Approve not required for native tokens');
+  }
+
+  // TODO cosmos consider refactoring this into a new method b.c. the other adapter
```
yorhodes: ahh yeah I was thinking this would be a `CosmosNativeToken implements IHypTokenAdapter `

src/consts/ibcRoutes.ts
```diff
@@ -0,0 +1,54 @@
...
+    // IBC Tia on Neutron
+    derivedIbcDenom: 'ibc/773B4D0A3CD667B2275D5A4A7A2F0909C0BA0F4059C0B9181E680DDF4965DCC7',
+    intermediateCaip2Id: 'cosmos:neutron-1',
+    // Router on Neutron for Manta
+    intermediateRouterAddress: 'neutron1ch7x3xgpnj62weyes8vfada35zff6z59kt2psqhnx9gjnt2ttqdqtva3pa',
+    // Router on Manta
+    destRouterAddress: '0x6Fae4D9935E2fcb11fC79a64e917fb2BF14DaFaa',
+  },
+  {
+    type: RouteType.IbcNativeToHypNative,
```
yorhodes: still think this should be derived from cross product of `ibcRoutes` and `hypRoutes` but not important atm
does `computeTokenRoutes` do something similar?

src/features/tokens/CosmNativeTokenAdapter.ts
```diff
@@ -0,0 +1,135 @@
...
+        token: this.properties.derivedIbcDenom,
+        warpRouter: this.addresses.intermediateRouterAddress,
+      },
+      this.properties.ibcDenom,
+    );
+    const transfer = await cwAdapter.populateTransferRemoteTx({
+      ...transferParams,
+      destination: this.properties.destinationDomainId,
+      // TODO cosmos quote real interchain gas payment
+      txValue: '1',
```
nambrot: No action needed here, but this is probably where we would have the logic that charges some amount of the token that is being transferred as IGP payment on the forwarding chain?

src/features/tokens/balances.ts
```diff
@@ -90,8 +92,10 @@ export function useDestinationBalance(
       const route = getTokenRoute(originCaip2Id, destinationCaip2Id, tokenCaip19Id, tokenRoutes);
       const protocol = getProtocolType(destinationCaip2Id);
       if (!route || !recipientAddress || !isValidAddress(recipientAddress, protocol)) return null;
-      const adapter = AdapterFactory.HypTokenAdapterFromRouteDest(route);
-      const balance = await adapter.getBalance(recipientAddress);
+      const tokenAdapter = isIbcOnlyRoute(route)
```
nambrot, line 95: Not sure if it's relevant, but if we can cut complexity by not supporting the underlying IBC transfer, we probably should. I don't believe folks will be using the UI for that purpose

src/features/tokens/routes/types.ts
```diff
@@ -3,22 +3,52 @@ export enum RouteType {
   CollateralToSynthetic = 'collateralToSynthetic',
   SyntheticToSynthetic = 'syntheticToSynthetic',
   SyntheticToCollateral = 'syntheticToCollateral',
+  IbcNativeToIbcNative = 'ibcNativeToIbcNative',
+  IbcNativeToHypNative = 'ibcNativeToHypNative',
```
nambrot: Still don't quite get what `HypNative` is in this context? Don't we usually call `HypNative` the warp contract that takes the native token on the EVM side?

src/features/tokens/routes/types.ts
```diff
@@ -3,22 +3,52 @@ export enum RouteType {
...
+  sourcePort: string;
+  sourceChannel: string;
+  derivedIbcDenom: string;
+}
+
+export interface IbcRoute extends BaseIbcRoute {
+  type: RouteType.IbcNativeToIbcNative;
+}
+
+export interface IbcToHypRoute extends BaseIbcRoute {
```
nambrot: Nit: Wonder if `IbcToWarpRoute` and `RouteType.IbcNativeToWarp` would be a bit more clear

src/features/tokens/routes/types.ts
```diff
@@ -3,22 +3,52 @@ export enum RouteType {
...
   destCaip2Id: ChainCaip2Id;
-  destRouterAddress: Address;
   destDecimals: number;
   // The underlying token on the destination chain
-  // Only set for CollateralToCollateral routes (b.c. sealevel need it)
+  // Only set for CollateralToCollateral routes (b.c. sealevel needs it)
   destTokenCaip19Id?: TokenCaip19Id;
 }
 
+export interface HypRoute extends BaseRoute {
```
nambrot: IMO confusing name (but I'm obviously not the naming expert)

src/features/wallet/hooks.tsx
```diff
@@ -64,29 +81,50 @@ export function useAccounts(): {
...
+  useEffect(() => {
+    if (cosmWalletStatus !== 'Done' || !cosmWalletClient) return;
+    const multiProvider = getMultiProvider();
+    const cosmChainNames = getCosmosChainNames();
+    const cosmChainCaip2Ids = cosmChainNames.map((c) =>
+      getCaip2Id(ProtocolType.Cosmos, multiProvider.getChainId(c)),
+    );
+    cosmWalletClient.enable?.(cosmChainNames).catch((err) => logger.error(err));
+    const cosmosAccountPromises = cosmChainNames.map((c) => cosmWalletClient.getAccount?.(c));
+    Promise.all(cosmosAccountPromises)
```
nambrot, line 106: Does this not need to be awaited?

nambrot (approved):
Mostly lgtm, obviously you know by now that my naming suggestions should be taken with a grain of salt, but imo Hyp or HypNative to me implies to much that the other side is a HypNative contract (when in facts its a HypERC20 synthetic), but would leave that up to you (or @yorhodes )

src/consts/chains.ts
jmrossy: Yeah, we need a chain that all wallets should already have b.c. the cosmos-kit hooks throw errors in certain cases where there are no cosmos chains provided to them (like on `main` branch where there are no cosmos tokens/chains configured)

src/consts/ibcRoutes.ts
jmrossy: Fair point, I can rename to HypSynthetic

src/features/tokens/CosmNativeTokenAdapter.ts
jmrossy: Actually this was intentionally generic in case we end up needing a general property bag in the 4th param spot for other adapters, that way we can keep the args more similar to each other

src/features/tokens/CosmNativeTokenAdapter.ts
jmrossy: Yes this is a possible place for that

src/features/tokens/balances.ts
jmrossy: Not really, that's why I added both

src/features/tokens/routes/types.ts
jmrossy: I'll call it HypSynthetic.

src/features/wallet/hooks.tsx
jmrossy: You can't await in a useEffect. It has a `then` fn attached to it

src/consts/ibcRoutes.ts
jmrossy: Sure but really this entire file shouldn't exist, not just the ToHypNative route. Ideally these routes would be configured in a way similar to the `consts/tokens.ts` file we already have.

As for whether or not the 2-hop route should be represented as a single object, we definitely want that. The places the UI use that route info (which is almost everywhere), benefit a lot from sustainability of these route details, regardless of their underlying mechanisms.

src/consts/values.ts
jmrossy: It works better to be a chain wallets already have. I don't think this does what you'd expect. It's really just a hack to work around limitations of cosmos-kit being used in a 'sometimes not cosmos' dapp.

PR #73, opened 2023-11-07
Increase cosmos IGP quote and use it in transfer creation

jmrossy:


nambrot (approved):

src/consts/chains.ts
```diff
@@ -61,7 +61,7 @@ export const chains: ChainMap<ChainMetadata & { mailbox?: Address }> = {
     ],
     nativeToken: {
       name: 'Neutron',
```
jmrossy, line 63: Should we change the name too?

PR #88, opened 2023-12-01
Add support for JSON and YAML chain and token configs

jmrossy:
Fixes https://github.com/hyperlane-xyz/issues/issues/732

- Add support for JSON and YAML chain and token configs
- Update some stale docs
- Cleanup default chain and token const files


avious00:
thanks for adding the additional comments at the top of the snippets in `/chains.yaml` + ` /chains.ts`

src/consts/tokens.yaml
```diff
@@ -0,0 +1,14 @@
+# A list of warp token configs
```
nambrot: I think we need to get a bit crisper IMO with our language. Very easy to add the warp token config that is input to the warp deploy command

src/features/chains/metadata.ts
```diff
@@ -12,9 +12,13 @@ import {
 } from '@hyperlane-xyz/sdk';
 import { ProtocolType } from '@hyperlane-xyz/utils';
 
-import { chains } from '../../consts/chains';
+import { chains as ChainsTS } from '../../consts/chains';
+import ChainsJson from '../../consts/chains.json';
```
nambrot, line 16: Do we even have a thing that outputs json chain metadata?

src/features/tokens/metadata.ts
```diff
@@ -1,7 +1,9 @@
 import { EvmTokenAdapter, ITokenAdapter, TokenType } from '@hyperlane-xyz/sdk';
 import { ProtocolType } from '@hyperlane-xyz/utils';
 
-import { tokenList } from '../../consts/tokens';
+import { tokenList as TokensTS } from '../../consts/tokens';
+import TokensJson from '../../consts/tokens.json';
+import TokensYaml from '../../consts/tokens.yaml';
```
nambrot, line 6: For warp token artifacts, they are not currently being output with tokens.yaml as the file name right?

nambrot (approved):
🙏 for fixing this

We are due for a more holistic overhaul of artifacts that are reused across, but one step at a time :)

src/features/chains/metadata.ts
jmrossy: The CLI commands can output either. There's a `format` flag that can be json or yaml.

src/consts/tokens.yaml
jmrossy: Good point, I'll rephrase

src/features/tokens/metadata.ts
jmrossy: The CLI artifact file name is `warp-ui-token-config.json` but I don't recommend we use that long name here

PR #93, opened 2023-12-19
Add support for a wallet blacklist config

jmrossy:
A more robust version of @tkporter's version here: https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/67/commits/2dde5d92ad74b0f7756280cc2a9bb3c63ff02948


yorhodes (changes requested):
honestly I question whether this is useful at all
the holder can just move the tokens?

yorhodes (approved):

Issue #92, opened 2023-12-19
Warp UI: More recipient address validation


jmrossy:
Prevent users from sending to recipient addresses that don't match the expected address format of the destination chain (e.g. an evm address to neutron, different cosmos chains with same address but different prefix)

PR #94, opened 2023-12-19
Validate cosmos recipient address bech32 prefix

jmrossy:
Fixes https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/issues/92


yorhodes (approved):
nice

PR #105, opened 2024-01-05
Fix igp quotes for ibc-based routes

jmrossy:


src/features/transfer/validateForm.ts
```diff
@@ -166,6 +166,9 @@ async function validateTokenBalances({
   if (sendValue.gt(balances.senderTokenBalance)) return { amount: 'Insufficient balance' };
 
   // Next, ensure balances can cover IGP fees
+  // But not for pure IBC routes because IGP is not used
+  if (isIbcOnlyRoute(route)) return null;
```
nambrot, line 170: Don't we need to quote the IGP on neutron if the transfer goes not to Neutron?

jmrossy: `isIbcOnlyRoute` will be true for routes that are purely IBC. Celestia -> Neutron is the only instance of that atm. For Celestia -> Arb (via neutron), this won't be true

nambrot (approved):

PR #120, opened 2024-01-19
Move config management to build-time 

jmrossy:
Fixes https://github.com/hyperlane-xyz/issues/issues/986

- Refactor chain, token, and route config management into build-time script
- Combine chain, token, and route state into single `WarpContext` var
- Fix outdated `publicRpcUrls` field in example chain config


src/context/context.ts
```diff
@@ -0,0 +1,33 @@
...
+  routes: RoutesMap;
+  multiProvider: MultiProtocolProvider<{ mailbox?: Address }>;
+}
+
+let warpContext: WarpContext;
+
+export function getWarpContext() {
+  if (!warpContext) {
+    warpContext = {
+      chains: Chains as any,
```
jmrossy, line 22: `any` cast is required because typescript isn't smart enough to recognize enums in the serialized data as the valid enum values. It's a safe cast because all this data gets run through zod at build time

src/features/routes/hooks.ts
```diff
@@ -0,0 +1,19 @@
+import { useMemo } from 'react';
+
+import { getChainIdFromToken } from '../caip/tokens';
+import { getTokens } from '../tokens/metadata';
+
+import { RoutesMap } from './types';
+
+export function useRouteChains(tokenRoutes: RoutesMap): ChainCaip2Id[] {
```
jmrossy, line 8: Not new, moved here from the now mostly gone `src/features/tokens/routes/hooks`

src/features/transfer/TransferTokenCard.tsx
```diff
@@ -22,22 +19,7 @@ export function TransferTokenCard() {
...
-        {routesError && (
-          <div className="my-32 flex flex-col items-center text-center">
-            <h3 className="text-red-500">Error searching for token routes.</h3>
-            <div className="mt-3 text-sm text-red-500">
-              Please ensure token configs are valid and RPCs are healthy.
-            </div>
-            <div className="mt-4 text-xs text-gray-500">{routesError.toString()}</div>
-          </div>
-        )}
+        <TransferTokenForm tokenRoutes={getWarpContext().routes} />
```
jmrossy, line 22: Tbh we don't _really_ need to pass the routes around as props anymore but I'm gonna punt on refactoring the component props for now. If we ever do need async routes again, we'll be glad to have it structured this way

src/scripts/buildConfigs/chains.ts
```diff
@@ -0,0 +1,33 @@
+import path from 'path';
```
jmrossy, line 1: Not new, moved here

src/scripts/buildConfigs/tokens.ts
```diff
@@ -0,0 +1,132 @@
+import path from 'path';
```
jmrossy, line 1: Mostly not new, moved here from src/features/tokens/metadata

src/scripts/buildConfigs/routes.ts
```diff
@@ -1,15 +1,37 @@
...
+import { TokenMetadata, TokenMetadataWithHypTokens } from '../../features/tokens/types';
+import { logger } from '../../utils/logger';
+
+export async function getRouteConfigs(context: WarpContext): Promise<RoutesMap> {
+  logger.info('Searching for token routes');
+  const processedTokens: TokenMetadataWithHypTokens[] = [];
+  for (const token of context.tokens) {
+    // Skip querying of IBC tokens
+    if (isIbcToken(token)) continue;
+    const tokenWithHypTokens = await fetchRemoteHypTokens(context, token, context.tokens);
```
paulbalaji: can remove the 3rd arg from `fetchRemoteHypTokens` if we're already passing through the full context?

src/context/context.ts
yorhodes: even with `resolveJsonModule` on the tsconfig?

src/scripts/buildConfigs/routes.ts
jmrossy: Good catch, thanks

src/context/context.ts
jmrossy: Yep. There may be a way to force it to work but tbh didn't seem worth the trouble

PR #121, opened 2024-01-22
Eclipsetestnet

daniel-savu:


src/consts/chains.ts
```diff
@@ -40,40 +40,56 @@ export const chains: ChainMap<ChainMetadata & { mailbox?: Address }> = {
     mailbox: 'Ge9atjAc3Ltu91VTbNpJDCjZ9CFxFyck4h3YBcTF9XPq',
   },
   solanatestnet: {
-    ...chainMetadata.solanatestnet,
```
jmrossy, line 43: Why remove the destructure here? Seems like most of the `solanatestnet` values are the same as the SDK const?

jmrossy (commented):
Looks good so far, just one question

PR #131, opened 2024-02-05
Add neutron to celestia IBC route

jmrossy:
Tested locally, transferred successfully.

source channel chosen based on Keplr's default channel choice for this route.


jmrossy:
Done as part of warp core migration

PR #154, opened 2024-04-12
Add Ancient8 USDC to Nexus

tkporter:
Happy to do an SDK release if we'd prefer


src/consts/chains.ts
```diff
@@ -1,4 +1,5 @@
-import { ChainMap, ChainMetadata, ExplorerFamily } from '@hyperlane-xyz/sdk';
+import { ChainMap, ChainMetadata, ChainTechnicalStack, ExplorerFamily } from '@hyperlane-xyz/sdk';
+import { etherToken } from '@hyperlane-xyz/sdk/dist/consts/chainMetadata';
```
jmrossy: How about using another chain's nativeToken (e.g. `ethereum.nativeToken`) instead of reaching into an unexported value like this? Imports into packages should go through the front door, with occasional exceptions for types or code that won't get bundled for production

jmrossy (approved):
Just one comment

src/consts/chains.ts
tkporter: sgtm!

PR #161, opened 2024-05-25
Warp UI Registry Support

jmrossy:
Requires https://github.com/hyperlane-xyz/hyperlane-registry/pull/43 and https://github.com/hyperlane-xyz/hyperlane-registry/pull/46
Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3692
Corresponds with https://github.com/hyperlane-xyz/v3-docs/pull/103

- Migrate from SDK consts and add registry support
- Upgrade Hyperlane package dependencies
- Avoid premature init of wagmi configs

**TODO:**

- [x] Merge and publish registry 43
- [x] Update widgets lib to latest registry
- [x] Decide on fix for token logos
- [x] Docs updates
- [x] Investigate Jon's test results


src/context/warpCoreConfig.ts
yorhodes, line 1: can you reuse local or partial registry here instead?
would be great for the UI not to have this IO code

src/features/chains/utils.ts
```diff
@@ -13,7 +14,10 @@ export function getChainDisplayName(chain: ChainName, shortName = false) {
 
 export function isPermissionlessChain(chain: ChainName) {
   if (!chain) return true;
-  return getChainMetadata(chain).protocol === ProtocolType.Ethereum || !chainMetadata[chain];
+  return (
+    getChainMetadata(chain).protocol !== ProtocolType.Ethereum ||
+    !CoreChains.includes(chain as CoreChain)
+  );
 }
```
yorhodes, line 21: what does "permissionless" mean in the UI? I think over time we should prune these distinctions
imo the registry is the right place to draw these boundaries

src/context/context.ts
```diff
@@ -8,40 +9,41 @@ import {
...
-  multiProvider: MultiProtocolProvider<{ mailbox?: Address }>;
+  registry: IRegistry;
+  chains: ChainMap<ChainMetadata>;
+  multiProvider: MultiProtocolProvider;
   warpCore: WarpCore;
 }
 
+// Note: This was initially static so it was simpler to keep it out of a state store.
+// Now it's somewhat dynamic based on env vars. If more flexibility is needed, it could be moved.
 let warpContext: WarpContext;
```
yorhodes, line 24: im not super familiar with this react query context pattern but this does feel like a code smell

src/consts/config.ts
```diff
@@ -2,6 +2,7 @@ import { ADDRESS_BLACKLIST } from './blacklist';
 
 const isDevMode = process?.env?.NODE_ENV === 'development';
 const version = process?.env?.NEXT_PUBLIC_VERSION || '0.0.0';
+const registryUrl = process?.env?.NEXT_PUBLIC_REGISTRY_URL || undefined;
```
yorhodes, line 5: nice
does NEXT_ prefix really make sense here?

src/consts/warpRouteWhitelist.ts
```diff
@@ -0,0 +1,6 @@
+// A list of warp route config IDs to be included in the app
+// If left empty, all warp routes in the configured registry will be included
```
yorhodes, line 2: love it

yorhodes: or maybe undefined is more intuitive?

src/context/chains.ts
```diff
@@ -1,29 +1,39 @@
...
+  let defaultChainMetadata = chainMetadata;
+  if (config.registryUrl) {
+    logger.debug('Using custom registry', config.registryUrl);
+    defaultChainMetadata = await registry.getMetadata();
+  } else {
+    logger.debug('Using default published registry');
+    // Note: this is an optional optimization to pre-fetch the content list
+    // and avoid repeated request from the chain logos that will use this info
+    await registry.listRegistryContent();
+  }
```
yorhodes, line 35: any way we can do this at build time?

src/components/icons/ChainLogo.tsx
```diff
@@ -1,29 +1,49 @@
...
     };
-  }, [chainName]);
+  }, [chainName, chainId]);
 
-  return <ChainLogoInner {...rest} chainId={chainId} chainName={chainDisplayName} icon={icon} />;
+  return (
+    <ChainLogoInner
+      chainName={name}
+      // TODO
+      // @ts-ignore update registry version in widget lib
```
yorhodes: should we maybe move the widgets lib into the monorepo? seems a little cumbersome atm to juggle across the repos

next.config.js
```diff
@@ -12,7 +12,7 @@ const isDev = process.env.NODE_ENV !== 'production'
 const ENABLE_CSP_HEADER = true;
 const FRAME_SRC_HOSTS = ['https://*.walletconnect.com', 'https://*.walletconnect.org','https://*.solflare.com'];
 const STYLE_SRC_HOSTS = ['https://*.googleapis.com']
-const IMG_SRC_HOSTS = ['https://*.walletconnect.com'];
+const IMG_SRC_HOSTS = ['https://*.walletconnect.com', 'https://*.githubusercontent.com'];
```
yorhodes, line 15: should we allow 
`NEXT_REGISTRY_URL`?

yorhodes: or alternatively force that this CORS stuff happens at build time

yorhodes (commented):
few questions around registry fetching

src/context/warpCoreConfig.ts
jmrossy: Maybe. It may not look like it cause the file was renamed so it's showing all green but this is mostly existing code. I wasn't looking to restructure things too deeply but yeah there's room for improvement

src/features/chains/utils.ts
jmrossy: Agreed. https://github.com/hyperlane-xyz/hyperlane-registry/issues/42 works towards that.

next.config.js
jmrossy: We only support github-based remote registries atm so when would NEXT_PUBLIC_REGISTRY_URL have a different URL than this?

jmrossy: > CORS stuff

CSP !== CORS :)

src/components/icons/ChainLogo.tsx
jmrossy: Yeah we did decide that we would remember? There's a ticket but it was de-prioritized because now that there aren't hard-coded chain logos in it, it doesn't need to change as often. https://github.com/hyperlane-xyz/issues/issues/1111

src/consts/config.ts
jmrossy: The `NEXT_PUBLIC_` prefix isn't for taste. It's what allows this env var to be exposed to client-side :)

src/consts/warpRouteWhitelist.ts
jmrossy: I'll support both

src/context/chains.ts
jmrossy: We could yes. But if we're looking to make this faster, I would start with this: https://github.com/hyperlane-xyz/hyperlane-registry/issues/29#issuecomment-2107820979

src/context/context.ts
jmrossy: This isn't related to react-query. Though you're right it is a smell... now. It wasn't when the context was simple and static. I considered moving it to a state store (hence the comment) but that would mean going through store access hooks and passing the context around through function params everywhere it's used. Not hard, but a verbose and far-reaching refactor. I accepted this tradeoff

next.config.js
yorhodes: > We only support github-based remote registries atm so when would NEXT_PUBLIC_REGISTRY_URL have a different URL than this?

fair enough

yorhodes (approved):

Issue #153, opened 2024-03-27
Research UX approach for address confirmation and make recs


avious00:

AlexBHarley:
@avious00 and @jmrossy interested for your thoughts on the below. Would say that there's no 100% sure fire way to ensure users don't send to the wrong address, it's probably the biggest problem CEXs deal with, but combining the below items together could prevent some of the problems reported.

#### Recommended

- Upgrade to RainbowKit & Wagmi V2
  - Bug fixes
  - Allow detecting users installed wallets rather than having a hardcoded list of recommended wallets
  - More likely to get help from teams behind these products if we can show reproductions of issues with the latest versions
- Visibility into recipient address 
  - Show registered name in addition to address
    1. ENS + ICNS to start, optional follow with Unstoppable Domains, Stargaze Names
    2. Some wallets expose the (locally given) name of the wallet. 
       - Here we can also be tricky with detecting extensions overwriting MetaMask. If we're using an injected provider we can check the `window.ethereum` object and sometimes they set some flags indicating which wallet it is
    4. Some connectors expose the type of wallet, ie MetaMask, OKX
  - Show image in addition to connected address
    1. NFT loaded from name service
    2. Locally derived image, ie MetaMask blockie/identicon, Rainbow icon, CB wallet
- Automatically load connected wallet address instead of making people click "self"
- If we've never bridged to this address before, flag as a new address

As a side note, some things I noted while checking out the warp route UI. If these are captured elsewhere apologies.

- Don't allow connecting wallet type if no chains of this type are configured
- Allow disconnecting each wallet rather than all at once
- Disable bridge button + show info rather than showing an error after clicking "Continue"
  - if insufficient balance
  - no token selected
- Choose first available token for route by default
- Remember selected token when switching network
- Info indicator on disabled tokens doesn't do anything

jmrossy:
@AlexBHarley @avious00 

> Upgrade to RainbowKit & Wagmi V2 

Definitely. Been wanted to, just hasn't been prioritized yet. I would strongly recommend also updating the cosmos and solana wallet libs to latest at the same time.

> Visibility into recipient address 

I'm open to this. I think it won't help in most cases because it's amateur users who fall into these traps most often but it's worth a try.

> Some connectors expose the type of wallet

[...]

PR #170, opened 2024-06-12
Bump dependencies

AlexBHarley:
A couple quick wins, more context here [#3952](https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/issues/153)


package.json
```diff
@@ -6,39 +6,40 @@
...
     "@chakra-ui/react": "^2.8.1",
-    "@cosmjs/cosmwasm-stargate": "^0.31.3",
-    "@cosmjs/stargate": "^0.31.3",
-    "@cosmos-kit/core": "^2.7.2",
-    "@cosmos-kit/cosmostation": "^2.4.4",
-    "@cosmos-kit/keplr": "^2.4.4",
-    "@cosmos-kit/leap": "^2.4.3",
-    "@cosmos-kit/react": "^2.5.3",
+    "@cosmjs/cosmwasm-stargate": "^0.32.3",
+    "@cosmjs/stargate": "^0.32.3",
```
jmrossy, line 10: If you upgrade this, you'll need to make sure the package patch that was applied for the injective blockchain still works. See the `injective` branch

package.json
```diff
@@ -6,39 +6,40 @@
...
+    "@cosmos-kit/leap": "^2.11.0",
+    "@cosmos-kit/react": "^2.15.0",
     "@emotion/react": "^11.11.1",
     "@emotion/styled": "^11.11.0",
     "@headlessui/react": "^1.7.14",
     "@hyperlane-xyz/registry": "2.1.0",
     "@hyperlane-xyz/sdk": "3.13.0",
     "@hyperlane-xyz/utils": "3.13.0",
     "@hyperlane-xyz/widgets": "3.13.1",
+    "@interchain-ui/react": "^1.23.16",
```
jmrossy, line 23: I can't quite remember but I _think_ there may have been a reason I didn't include this, like to avoid it's inclusion in the bundle when it's not really necessary. Can you confirm it's a necessary addition?

jmrossy (commented):
Thanks for making improvements here but I think we'll need to be cautious about re-testing things

PR #172, opened 2024-06-12
Investigation: trying to detect overwritten connector names

AlexBHarley:
We _can sometimes_ detect if an extension is overriding MetaMask, but the implementation is hacky and prone to breaking when wallets update what they inject.


src/features/wallet/hooks/useConnectorName.ts
```diff
@@ -0,0 +1,37 @@
...
+import { AccountInfo } from './types';
+
+function useEvmConnectorName(accountInfo: AccountInfo | undefined) {
+  const account = useAccount();
+
+  if (
+    account.connector instanceof InjectedConnector &&
+    accountInfo?.connectorName === account.connector.name
+  ) {
+    if (window.ethereum?.isOkxWallet || window.ethereum?.isOKExWallet)
```
jmrossy: Did you mean to return something here?

jmrossy (commented):
This PR should be in the `motivation` section for [EIP 6963](https://eips.ethereum.org/EIPS/eip-6963) haha 

It's good to know we have some options, albeit hacky ones. If we do take this approach, I'd prefer if this logic were organized by protocol like the other stuff under `src/wallet/hooks`

src/features/wallet/hooks/useConnectorName.ts
AlexBHarley: Er yes my bad

AlexBHarley:
Yeah, just wanted to provide as reference.

I think the bigger win will be RainbowKit V2 as that does have 6963 support, and many wallets are supporting now. But that will be a much bigger upgrade I believe

PR #177, opened 2024-06-13
Sanctions

AlexBHarley:
fixes https://github.com/hyperlane-xyz/issues/issues/1067

### DIFF
--- a/src/features/sanctions/hooks/useChainalysisSanctioned.ts
+++ b/src/features/sanctions/hooks/useChainalysisSanctioned.ts
@@ -1,0 +1,43 @@
+import { isAddress } from 'viem';
+import { useContractRead } from 'wagmi';
+
+import { useEvmAccount } from '../../wallet/hooks/evm';
+
+// https://go.chainalysis.com/chainalysis-oracle-docs.html
+const ORACLE_ABI = [
+  {
+    inputs: [
+      {
+        internalType: 'address',
+        name: 'addr',
+        type: 'address',
+      },
+    ],
+    name: 'isSanctioned',
+    outputs: [
+      {
+        internalType: 'bool',
+        name: '',
+        type: 'bool',
+      },
+    ],
+    stateMutability: 'view',
+    type: 'function',
+  },
+] as const;
+const ORACLE_ADDRESS = '0x40C57923924B5c5c5455c48D93317139ADDaC8fb';
+
+export const useChainalysisSanctioned = () => {
+  const evmAddress = useEvmAccount().addresses[0]?.address;
+
+  const sanctioned = useContractRead({
+    abi: ORACLE_ABI,
+    functionName: 'isSanctioned',
+    args: [isAddress(evmAddress) ? evmAddress : '0x'],
+    chainId: 1,
+    address: ORACLE_ADDRESS,
+    enabled: !!evmAddress,
+  });
+
+  return !!sanctioned.data;
+};
--- a/src/features/sanctions/hooks/useIsAccountChainalysisSanctioned.ts
+++ b/src/features/sanctions/hooks/useIsAccountChainalysisSanctioned.ts
@@ -27,7 +27,7 @@ const ORACLE_ABI = [
 ] as const;
 const ORACLE_ADDRESS = '0x40C57923924B5c5c5455c48D93317139ADDaC8fb';
 
-export const useIsAccountChainalysisSanctioned = () => {
+export function useIsAccountChainalysisSanctioned() {
   const evmAddress = useEvmAccount().addresses[0]?.address;
 
   const sanctioned = useContractRead({
@@ -40,4 +40,4 @@ export const useIsAccountChainalysisSanctioned = () => {
   });
 
   return !!sanctioned.data;
-};
+}
--- a/src/features/sanctions/hooks/useIsAccountChainalysisSanctioned.ts
+++ b/src/features/sanctions/hooks/useIsAccountChainalysisSanctioned.ts
@@ -27,7 +27,7 @@ const ORACLE_ABI = [
 ] as const;
 const ORACLE_ADDRESS = '0x40C57923924B5c5c5455c48D93317139ADDaC8fb';
 
-export const useChainalysisSanctioned = () => {
+export const useIsAccountChainalysisSanctioned = () => {
   const evmAddress = useEvmAccount().addresses[0]?.address;
 
   const sanctioned = useContractRead({
--- a/src/features/sanctions/hooks/useOfacSanctioned.ts
+++ b/src/features/sanctions/hooks/useOfacSanctioned.ts
@@ -10,7 +10,7 @@ export const useOfacSanctioned = () => {
   const evmAddress = useEvmAccount().addresses[0]?.address;
 
   const sanctionedAddresses = useQuery<string[]>({
-    queryKey: ['ofac sanctioned', evmAddress],
+    queryKey: ['useOfacSanctioned', evmAddress],
     queryFn: () => fetch(OFAC_SANCTIONED_ADDRESSES_ENDPOINT).then((x) => x.json()),
     enabled: !!evmAddress,
   });
--- a/src/features/sanctions/hooks/useOfacSanctioned.ts
+++ b/src/features/sanctions/hooks/useOfacSanctioned.ts
@@ -1,0 +1,24 @@
+import { useQuery } from '@tanstack/react-query';
+import { isAddress, isAddressEqual } from 'viem';
+
+import { useEvmAccount } from '../../wallet/hooks/evm';
+
+const OFAC_SANCTIONED_ADDRESSES_ENDPOINT =
+  'https://raw.githubusercontent.com/0xB10C/ofac-sanctioned-digital-currency-addresses/lists/sanctioned_addresses_ETH.json';
+
+export const useOfacSanctioned = () => {
+  const evmAddress = useEvmAccount().addresses[0]?.address;
+
+  const sanctionedAddresses = useQuery<string[]>({
+    queryKey: ['ofac sanctioned', evmAddress],
+    queryFn: () => fetch(OFAC_SANCTIONED_ADDRESSES_ENDPOINT).then((x) => x.json()),
+    enabled: !!evmAddress,
+  });
+
+  return (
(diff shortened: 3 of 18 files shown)


avious00:
> @AlexBHarley is attempting to deploy a commit to the **Abacus Works** Team on [Vercel](https://vercel.com).
> 
> A member of the Team first needs to [authorize it](https://vercel.com/git/authorize?team=Abacus%20Works&type=github&job=%7B%22headInfo%22%3A%7B%22sha%22%3A%226b9b04d6a7ba1eb0e4f06fa2a7132b3d6dd3d65c%22%7D%2C%22id%22%3A%22QmcvsYSSSAaf1HwuMtBZKmS7rudo5wUim6ysDwajY3NZLH%22%2C%22org%22%3A%22hyperlane-xyz%22%2C%22prId%22%3A177%2C%22repo%22%3A%22hyperlane-warp-ui-template%22%7D).

authorized it

src/features/sanctions/hooks/useChainalysisSanctioned.ts
```diff
@@ -0,0 +1,43 @@
...
+const ORACLE_ADDRESS = '0x40C57923924B5c5c5455c48D93317139ADDaC8fb';
+
+export const useChainalysisSanctioned = () => {
+  const evmAddress = useEvmAccount().addresses[0]?.address;
+
+  const sanctioned = useContractRead({
+    abi: ORACLE_ABI,
+    functionName: 'isSanctioned',
+    args: [isAddress(evmAddress) ? evmAddress : '0x'],
+    chainId: 1,
```
jmrossy: Can you confirm it's sufficient to check only eth mainnet? Do the chainanalysis contracts on other chains have the same data?

src/features/sanctions/hooks/useOfacSanctioned.ts
```diff
@@ -0,0 +1,24 @@
...
+import { useEvmAccount } from '../../wallet/hooks/evm';
+
+const OFAC_SANCTIONED_ADDRESSES_ENDPOINT =
+  'https://raw.githubusercontent.com/0xB10C/ofac-sanctioned-digital-currency-addresses/lists/sanctioned_addresses_ETH.json';
+
+export const useOfacSanctioned = () => {
+  const evmAddress = useEvmAccount().addresses[0]?.address;
+
+  const sanctionedAddresses = useQuery<string[]>({
+    queryKey: ['ofac sanctioned', evmAddress],
```
jmrossy: This key is fine but a convention I've used elsewhere in the app is to use the hook name, so 'useOfacSanctioned'

src/features/sanctions/hooks/useChainalysisSanctioned.ts
```diff
@@ -0,0 +1,43 @@
...
+      },
+    ],
+    stateMutability: 'view',
+    type: 'function',
+  },
+] as const;
+const ORACLE_ADDRESS = '0x40C57923924B5c5c5455c48D93317139ADDaC8fb';
+
+export const useChainalysisSanctioned = () => {
+  const evmAddress = useEvmAccount().addresses[0]?.address;
```
jmrossy: Please confirm with @avious00 that it's sufficient for this task to just check EVM and not Cosmos and/or Solana. Typically new features for the Warp UI should be done in a multi-protocol way

src/features/sanctions/hooks/useOfacSanctioned.ts
```diff
@@ -0,0 +1,24 @@
...
+    queryFn: () => fetch(OFAC_SANCTIONED_ADDRESSES_ENDPOINT).then((x) => x.json()),
+    enabled: !!evmAddress,
+  });
+
+  return (
+    !!sanctionedAddresses.data &&
+    !!evmAddress &&
+    isAddress(evmAddress) &&
+    !!sanctionedAddresses.data.find((x) => isAddress(x) && isAddressEqual(x, evmAddress))
```
jmrossy: Use our own `eqAddress` fns from '@hyperlane-xyz/utils` which handle multiple protocols

src/features/sanctions/hooks/useChainalysisSanctioned.ts
```diff
@@ -0,0 +1,43 @@
...
+        type: 'bool',
+      },
+    ],
+    stateMutability: 'view',
+    type: 'function',
+  },
+] as const;
+const ORACLE_ADDRESS = '0x40C57923924B5c5c5455c48D93317139ADDaC8fb';
+
+export const useChainalysisSanctioned = () => {
```
jmrossy: Nit: suggest `useIsAccountChainanalysisSanctioned`, my first thought was that this returned the list, not the account status

src/features/sanctions/hooks/useOfacSanctioned.ts
```diff
@@ -0,0 +1,24 @@
+import { useQuery } from '@tanstack/react-query';
+import { isAddress, isAddressEqual } from 'viem';
+
+import { useEvmAccount } from '../../wallet/hooks/evm';
+
+const OFAC_SANCTIONED_ADDRESSES_ENDPOINT =
+  'https://raw.githubusercontent.com/0xB10C/ofac-sanctioned-digital-currency-addresses/lists/sanctioned_addresses_ETH.json';
+
+export const useOfacSanctioned = () => {
```
jmrossy: Ditto for name clarity

src/features/sanctions/hooks/useSanctioned.ts
```diff
@@ -0,0 +1,9 @@
+import { useChainalysisSanctioned } from './useChainalysisSanctioned';
+import { useOfacSanctioned } from './useOfacSanctioned';
+
+export const useSanctioned = () => {
```
jmrossy: ditto

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -215,6 +216,7 @@ function ButtonSection({
   setIsReview: (b: boolean) => void;
 }) {
   const { values } = useFormikContext<TransferFormValues>();
+  const sanctioned = useSanctioned();
```
jmrossy: Nit: sanctioned -> isSanctioned

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -228,6 +230,9 @@ function ButtonSection({
   }));
 
   const triggerTransactionsHandler = async () => {
+    if (sanctioned) {
+      return;
```
jmrossy, line 234: Pop an error toast here

src/features/wallet/SideBarMenu.tsx
```diff
@@ -37,6 +38,7 @@ export function SideBarMenu({
   const disconnects = useDisconnectFns();
   const { readyAccounts } = useAccounts();
   const didMountRef = useRef(false);
+  const sanctioned = useSanctioned();
```
jmrossy: Tbh we don't need to check this here. Trevor's first pass PR at a blacklist feature may have misled you.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -107,24 +109,34 @@ export function SideBarMenu({
...
-                ))}
-            </div>
-            {sortedTransfers?.length > 0 && (
-              <button onClick={resetTransfers} className="my-5 mx-2 flex flex-row items-center">
-                <Image className="mr-4" src={ResetIcon} width={17} height={17} alt="" />
-                <span className="text-gray-900 text-sm font-normal">Reset transaction history</span>
-              </button>
+            {sanctioned ? (
+              <div className="text-gray-800 text-sm font-normal text-center">
+                <span>Unable to pull transaction history</span>
```
jmrossy: This sidebar doesn't pull anything. It just surfaces info from the persisted zustand store. A local history of txs.

src/features/sanctions/hooks/useChainalysisSanctioned.ts
AlexBHarley: None of them are very good but yes the Ethereum contract has the most addresses registered and afaict all the other addresses are also registered on the Ethereum contract.

A correct implementation would treat EVM addresses as equivalent, as risk has nothing to do with the chain.

src/features/sanctions/hooks/useChainalysisSanctioned.ts
AlexBHarley: Totally agree but I couldn't find any lists to use for Cosmos or Solana.

This approach blocks all functionality for the app if you have a sanctioned wallet connected which feels like the correct behaviour.

src/features/transfer/TransferTokenForm.tsx
AlexBHarley: See the Discord discussion

src/features/sanctions/hooks/useIsAccountChainalysisSanctioned.ts
```diff
@@ -0,0 +1,43 @@
...
+        type: 'bool',
+      },
+    ],
+    stateMutability: 'view',
+    type: 'function',
+  },
+] as const;
+const ORACLE_ADDRESS = '0x40C57923924B5c5c5455c48D93317139ADDaC8fb';
+
+export const useIsAccountChainalysisSanctioned = () => {
```
jmrossy: Missed this on my first pass but the rest of this app uses `function` syntax instead of `=>` arrow fn syntax for fns defined at the module root scope. Named functions integrate with the debugger better, especially async ones. What you've done here is define an anon fn and then assigned it to a named var. Named fns also support recursion. I recommend we keep using them, if only for consistency.

Now when you're defining a function _inline_, like for a callback, that should always be an arrow fn so that the default `this` bindings are correct.

jmrossy (approved):
@AlexBHarley Approved with one comment

src/features/sanctions/hooks/useIsAccountChainalysisSanctioned.ts
AlexBHarley: Thanks for this, + 1 for consistency.

Not sure I fully understand the recursion comment, arrow functions do support that. 

Totally agree that arrow functions should be used for callbacks though, the days of `.bind` are long behind us.

PR #179, opened 2024-06-17
RPC Overrides

AlexBHarley:
Closes https://github.com/hyperlane-xyz/issues/issues/1160


src/utils/json.ts
```diff
@@ -0,0 +1,8 @@
+export function tryParseJson(input: string): unknown | null {
+  try {
+    return JSON.parse(input);
+  } catch (e) {
+    console.warn('unable to parse JSON', e);
```
jmrossy: Use logger. Atm logger just routes to console but that may not always be the case. I thought there was a lint rule enabled to prevent console but I guess it's not enabled in this project. Bonus points for fixing that :)

src/context/chains.ts
```diff
@@ -1,14 +1,17 @@
...
+import { ChainMap, ChainMetadata, ChainMetadataSchema, RpcUrlSchema } from '@hyperlane-xyz/sdk';
 
 import { chains as ChainsTS } from '../consts/chains.ts';
 import ChainsYaml from '../consts/chains.yaml';
 import { config } from '../consts/config.ts';
 import { cosmosDefaultChain } from '../features/chains/cosmosDefault';
+import { tryParseJson } from '../utils/json.ts';
 import { logger } from '../utils/logger';
 
+const RPC_OVERRIDES = process.env['NEXT_PUBLIC_RPC_OVERRIDES'];
```
jmrossy: access this safely in src/consts/config like the rest of the env vars

src/context/chains.ts
```diff
@@ -34,6 +43,25 @@ export async function assembleChainMetadata() {
...
+    ...defaultChainMetadata,
+    ...customChainMetadata,
+  }).reduce((accum, [name, chain]) => {
+    const privateRpc = rpcOverrides.success ? rpcOverrides.data[name] : null;
+    if (privateRpc) {
+      return {
+        ...accum,
+        [name]: {
+          ...chain,
+          rpurls: chain.rpcUrls,
```
jmrossy: Typo `rpurls`. But also the type looks wrong. Above and in your example it looks like rpcOverrides is a map of chain name -> rpcUrlSchema. However the rpcUrls field is an array.

jmrossy (changes requested):
Needs fixes and more testing.

src/context/chains.ts
AlexBHarley: Nice I didn't see this file

src/context/chains.ts
AlexBHarley: Yep this was half baked, apologies

jmrossy (approved):

Xaroz:
closing PR as this was continued on #504

PR #181, opened 2024-06-18
feat: New sidebar wallet button design

### LINKED ISSUE #180: Improve wallet presentation in sidebar

jmrossy:
Fixes #180 

Group addresses by protocol and show wallet logos

Screenshot 2024-06-18 at 1 01 44 PM


AlexBHarley (approved):
Nice one, IMO this is strictly better than what we have right now with the so we should probably merge.

But, I think we ultimately should show the addresses. Maybe rendering something like the following? I could also see a hover type thing working nicely.



Screenshot 2024-06-18 at 17 53 21

PR #196, opened 2024-07-09
Add new warp route with Fraxtal

nambrot:
Updates the config to the warp route that includes fraxtal

### DIFF
--- a/src/consts/warpRoutes.yaml
+++ b/src/consts/warpRoutes.yaml
@@ -4,6 +4,7 @@ tokens:
     standard: EvmHypXERC20
     decimals: 18
     symbol: ezETH
+    logoURI: https://renzo.hyperlane.xyz/logos/ezeth.svg
     name: Renzo Restaked ETH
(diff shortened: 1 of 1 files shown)


nbayindirli (commented):
Screenshot 2024-07-09 at 4 18 04 PM

confirmed Fraxtal is showing up nicely in the UI but looks like ezeth logo is missing from token selection

nbayindirli (approved):

src/consts/warpRoutes.yaml
```diff
@@ -1,1034 +1,146 @@
+# yaml-language-server: $schema=../schema.json
 tokens:
-  - chainName: blast
+  - chainName: arbitrum
     standard: EvmHypXERC20
-    name: Renzo Restaked ETH
+    decimals: 18
     symbol: ezETH
     logoURI: https://renzo.hyperlane.xyz/logos/ezeth.svg
```
jmrossy, line 7: @nambrot @nbayindirli These should be relative which is less fragile (e.g. domain name change) but nbd

PR #213, opened 2024-08-20
pzETH PoC

tkporter:
don't merge!


next.config.js
```diff
@@ -99,4 +99,16 @@ const sentryWebpackPluginOptions = {
...
+  images: {
+    remotePatterns: [
+      {
+        protocol: 'https',
+        hostname: 's2.coinmarketcap.com',
+        port: '',
+        pathname: '/**',
+      },
+    ],
+  },
```
jmrossy, line 113: We shouldn't allow this. The Immunefi report from a few weeks ago showed that maliciously crafted SVGs on other domains + Next's custom `Image` component can be easily exploited to hijack the page content and phish users. The fix was to lock down this image.remotePatterns setting and we shouldn't open it back up

src/consts/chains.ts
```diff
@@ -1,10 +1,47 @@
-import { ChainMap, ChainMetadata } from '@hyperlane-xyz/sdk';
+import { ChainMap, ChainMetadata, ExplorerFamily } from '@hyperlane-xyz/sdk';
+import { ProtocolType } from '@hyperlane-xyz/utils';
 
 // A map of chain names to ChainMetadata
 // Chains can be defined here, in chains.json, or in chains.yaml
 // Chains already in the SDK need not be included here unless you want to override some fields
 // Schema here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/metadata/chainMetadataTypes.ts
 export const chains: ChainMap<ChainMetadata & { mailbox?: Address }> = {
+  solana: {
```
jmrossy, line 9: Can we DRY this up with the solana from the registry (plus maybe the mailbox field)

src/context/context.ts
```diff
@@ -29,6 +29,10 @@ export function getWarpContext() {
 
 export async function initWarpContext() {
   const { registry, chains } = await assembleChainMetadata();
+  if (chains) {
```
jmrossy, line 32: TODO I assume?

next.config.js
tkporter: Yeah very happy to not commit this - this was just a PoC we hacked together to show a partner what the experience would look like

tkporter:
agree with all your comments @jmrossy  -- this was only intended to be hacked together so closing this PR, more sustainable fixes would come with some work that Dan's been doing on finalizing the deploy

PR #252, opened 2024-09-19
chore: update to registry v4.3.4

paulbalaji:
chore: update to registry v4.3.4


paulbalaji:
negated by force commit https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/commit/fdc08c00132e06739674d4095fec6442400acd2e

PR #256, opened 2024-09-19
chore: enable tETH eclipse<>ethereum

paulbalaji:
chore: enable tETH eclipse<>ethereum
- requires registry update to v4.3.6


daniel-savu (approved):
🚢

PR #284, opened 2024-10-09
new road base_mode , for toby token on base

macfly-base:


jmrossy:
@macfly-base You shouldn't try to add your configs to this template. It's meant to be a starting point for people to fork

macfly-base:
handling error, thank you for your comment, I will take it into account in the future

PR #286, opened 2024-10-12
Update LICENSE.md

v4441:
Just a typo.


jmrossy:
Thanks but this the license is not something we wrote. It's the boilerplate Apache license and we probably should leave it unedited

v4441:
It's just that, as it is, the sentence doesn't end as intended. But I understand. Thank you for the review.

PR #291, opened 2024-10-15
feat: Version update for SDK and Registry

ltyu:
Includes Sei and Taiko for ezEth


package.json
```diff
@@ -17,8 +17,8 @@
     "@emotion/react": "^11.13.0",
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^1.7.14",
-    "@hyperlane-xyz/registry": "4.6.0",
-    "@hyperlane-xyz/sdk": "5.3.0",
+    "@hyperlane-xyz/registry": "4.8.0",
+    "@hyperlane-xyz/sdk": "5.5.0",
     "@hyperlane-xyz/utils": "5.3.0",
     "@hyperlane-xyz/widgets": "5.3.0",
```
paulbalaji: worth updating these two as well?

paulbalaji: think you'll have to

ltyu: fixed

jmrossy (approved):

PR #302, opened 2024-10-23
chore: use warp route whitelist

ltyu:
Looks like this was incorrectly configured previously.

https://discord.com/channels/935678348330434570/1298686381698060441/1298702557148872784
"Part of the reason for the confusion is that the ezeth branch was setup wrong. Routes are supposed to live in the registry and then the warp UI just picks the ones it wants via the warpRouteWhitelist..."


src/consts/warpRoutes.yaml
```diff
@@ -1,264 +1,6 @@
...
-    logoURI: /deployments/warp_routes/EZETH/RedETH-logo.svg
-    name: Restaked ETH on Sei
-    standard: EvmHypXERC20
-    symbol: redETH
+# A list of Warp Route token configs
+# These configs will be merged with the warp routes in the configured registry
+# The input here is typically the output of the Hyperlane CLI warp deploy command
+---
+tokens: []
+options: {}
```
paulbalaji: prettier is probably complaining about the lack of trailing line

src/consts/warpRouteWhitelist.ts
```diff
@@ -2,8 +2,7 @@
 // Warp Route IDs use format `SYMBOL/chainname1-chainname2...` where chains are ordered alphabetically
 // If left null, all warp routes in the configured registry will be included
 // If set to a list (including an empty list), only the specified routes will be included
-export const warpRouteWhitelist: Array<string> | null = [];
-// Example:
-// [
-//   // 'ETH/ethereum-viction'
-// ];
+export const warpRouteWhitelist: Array<string> | null = [
+  'ezETH/arbitrum-base-blast-bsc-ethereum-fraxtal-linea-mode-optimism-sei-taiko-zircuit',
```
jmrossy, line 6: Just confirming, do we need to update the registry version on this branch of it's already good?

jmrossy (approved):
Just one question and like Paul said, you need to make prettier happy :)
Thanks!

src/consts/warpRouteWhitelist.ts
ltyu: its already good

PR #303, opened 2024-10-25
feat: add new Stride & Eclipse TIA and stTIA warp routes

tkporter:
- Adds the TIA and stTIA warp routes between Stride & Eclipse
- Uses newer registry version with these artifacts
- Uses newer monorepo packages with a fix to ensure Sealevel -> Cosmos transfers work


daniel-savu (approved):

package.json
```diff
@@ -17,10 +17,10 @@
...
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^1.7.14",
-    "@hyperlane-xyz/registry": "4.9.0",
-    "@hyperlane-xyz/sdk": "5.5.0",
-    "@hyperlane-xyz/utils": "5.5.0",
-    "@hyperlane-xyz/widgets": "5.5.0",
+    "@hyperlane-xyz/registry": "4.11.1",
+    "@hyperlane-xyz/sdk": "5.6.2",
+    "@hyperlane-xyz/utils": "5.6.2",
+    "@hyperlane-xyz/widgets": "5.6.2",
```
jmrossy, line 23: Strictly speaking, 'functional' changes like dep upgrades should be done against `main` and then merged into the prod branches instead of directly to `nexus`. This is to minimize the divergence of the prod instances. But I understand this is easier

jmrossy (approved):

package.json
tkporter: makes sense - moving this into https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/304. I see there's a pretty big deviation between this branch and main apart from these though - out of caution I'll just update these deps

tkporter: Not certain what the process is for merging https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/279/ or what so just gonna merge this for now

PR #313, opened 2024-11-06
chore: use chain name instead of chainId for ChainLogo

paulbalaji:
chore: use chain name instead of chainId for ChainLogo

pre-empting https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4798


src/features/chains/ChainSelectField.tsx
```diff
@@ -21,13 +21,13 @@ export function ChainSelectField({ name, label, chains, onChange, disabled }: Pr
   const [field, , helpers] = useField<ChainName>(name);
   const { setFieldValue } = useFormikContext<TransferFormValues>();
 
-  const handleChange = (newChainId: ChainName) => {
-    helpers.setValue(newChainId);
+  const handleChange = (chainName: ChainName) => {
```
jmrossy, line 24: thanks! A vestige of the first version of this app which used chainId everywhere

jmrossy (approved):

PR #319, opened 2024-11-09
refactor: Replace global warp context with store data

jmrossy:
This PR moves the Warp Context pieces (warpCore, multiProvider, registry) into the Zustand store. That enables support for dynamic chain metadata / token route overrides at runtime, as required for the integration of the new ChainSearchMenu component from the widgets lib.


src/features/WarpContextInitGate.tsx
```diff
@@ -0,0 +1,28 @@
...
+    );
+  }
+
+  if (!isWarpContextReady)
+    return (
+      <div className="flex h-screen items-center justify-center bg-primary-500">
+        <Spinner classes="opacity-50" white />
+      </div>
+    );
+
```
Xaroz: I don't see where you are canceling the timeout when the warpContext is ready, shouldn't you be calling the returned cancel function if warpContext is ready or maybe I am missing something here?

Xaroz (changes requested):
small comment otherwise LGTM

src/features/WarpContextInitGate.tsx
jmrossy: Derp, yes you're totally right. This part hasn't been tested yet.

Xaroz (approved):

PR #323, opened 2024-11-14
feat: Chain Search Menu

Xaroz:
- Replace current chain menu with `ChainSearchMenu` from widgets
- Create a `chainMetadata` (not taken from the registry) with the tokens that come from the warpCore
- Create a `customListItemField` object that contains all the routes from a single chain to any other warp core token chain
- Chain Search Menu will show the routes from the opposing field. EX: If you open `"To"` chain, it will show all the routes from the `"From"` field chain to all the chains

image

### DIFF
--- a/src/features/chains/hooks.ts
+++ b/src/features/chains/hooks.ts
@@ -1,6 +1,7 @@
 import { ChainMap, ChainMetadata } from '@hyperlane-xyz/sdk';
 import { useMemo } from 'react';
 import { useStore } from '../store';
+import { useWarpCore } from '../tokens/hooks';
 import { getChainDisplayName } from './utils';
 
 export function useMultiProvider() {
@@ -32,10 +33,12 @@ export function useChainDisplayName(chainName: ChainName, shortName = false) {
 }
 
 /**
- * Return a chainMetadata object with given chains
+ * Returns chainMetadata object with warp core token chains
  */
-export function useChainMetadataMap(chains: string[]) {
+export function useTokenChainsMetadata() {
   const multiProvider = useMultiProvider();
+  const warpCore = useWarpCore();
+  const chains = useMemo(() => warpCore.getTokenChains(), [warpCore]);
 
   const chainMetadata = useMemo(() => {
     return chains.reduce<ChainMap<ChainMetadata>>((obj, chain) => {
--- a/src/features/chains/hooks.ts
+++ b/src/features/chains/hooks.ts
@@ -31,6 +31,9 @@ export function useChainDisplayName(chainName: ChainName, shortName = false) {
   return getChainDisplayName(multiProvider, chainName, shortName);
 }
 
+/**
+ * Return a chainMetadata object with given chains
+ */
 export function useChainMetadataMap(chains: string[]) {
   const multiProvider = useMultiProvider();
 
--- a/src/features/chains/hooks.ts
+++ b/src/features/chains/hooks.ts
@@ -1,3 +1,5 @@
+import { ChainMap, ChainMetadata } from '@hyperlane-xyz/sdk';
+import { useMemo } from 'react';
 import { useStore } from '../store';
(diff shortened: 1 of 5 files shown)


src/features/chains/hooks.ts
```diff
@@ -28,3 +30,20 @@ export function useChainDisplayName(chainName: ChainName, shortName = false) {
...
+  const chainMetadata = useMemo(() => {
+    return chains.reduce<ChainMap<ChainMetadata>>((obj, chain) => {
+      const metadata = multiProvider.tryGetChainMetadata(chain);
+      if (metadata) obj[chain] = metadata;
+      return obj;
+    }, {});
+  }, [chains, multiProvider]);
+
+  return chainMetadata;
+}
```
jmrossy: I don't think you need this hook. IIUC, `multiprovider.metadata` is already what you're looking for

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -111,14 +112,37 @@ function SwapChainsButton({ disabled }: { disabled?: boolean }) {
...
+
+  const { values } = useFormikContext<TransferFormValues>();
+
+  const originRoutes = useMemo(() => {
+    return getCustomListItemField(warpCore, values.origin, chainMetadata);
+  }, [values.origin, warpCore, chainMetadata]);
+
+  const destinationRoutes = useMemo(() => {
+    return getCustomListItemField(warpCore, values.destination, chainMetadata);
+  }, [values.destination, warpCore, chainMetadata]);
```
jmrossy: Can we move this into `ChainSelectField`?

src/features/chains/utils.ts
```diff
@@ -31,3 +31,26 @@ export function getChainByRpcUrl(multiProvider: MultiProtocolProvider, url?: str
     (m) => !!m.rpcUrls.find((rpc) => rpc.http.toLowerCase().includes(url.toLowerCase())),
   );
 }
+
+/**
+ * Return a customListItemField object that contains the amount of
+ * routes from a single chain to every other chain
+ */
+export function getCustomListItemField(
```
jmrossy: Please move this to ChainSelectField as that's the only place that it will be used.

src/features/chains/utils.ts
```diff
@@ -31,3 +31,26 @@ export function getChainByRpcUrl(multiProvider: MultiProtocolProvider, url?: str
...
+  warpCore: WarpCore,
+  origin: ChainName,
+  chains: ChainMap<ChainMetadata>,
+): ChainMap<{ display: string; sortValue: number }> {
+  return Object.keys(chains).reduce<ChainMap<{ display: string; sortValue: number }>>(
+    (obj, chainName) => {
```
jmrossy: Please rename chainName -> destination

jmrossy (commented):
Excited to get this in! Def a nice UX improvement

src/features/chains/hooks.ts
Xaroz: This hook is needed because we only need the token chains from the warp core, and that doesn't return the `metadata` we need, so we have to create it from scratch.

I renamed this function to `useTokenChainsMetadata` and also moved the `warpCore` and getting the chains logic inside

src/features/transfer/TransferTokenForm.tsx
Xaroz: Reason I didn't do that it is because I think it looks better to separate the logic here instead of having it inside the `ChainSelectField` by doing something like `name === "origin" ? values.destination : values.origin`

src/features/chains/utils.ts
Xaroz: kept in here but renamed to `getAllTokenRoutesFromChain`, please lmk what you think

src/features/chains/utils.ts
Xaroz: done!

PR #325, opened 2024-11-16
green back color

mitjamob:


jmrossy:
@mitjamob customizations you make for your branding and routes should not be merged back to the template's main branch. No need for PRs like this

PR #328, opened 2024-11-18
feat: Replace chain picker with new ChainSearchMenu from Widgets lib

jmrossy:
Builds on https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/323 but resolves conflicts and improves the picker custom field data

### DIFF
--- a/src/features/chains/hooks.ts
+++ b/src/features/chains/hooks.ts
@@ -1,7 +1,4 @@
-import { ChainMap, ChainMetadata } from '@hyperlane-xyz/sdk';
-import { useMemo } from 'react';
 import { useStore } from '../store';
-import { useWarpCore } from '../tokens/hooks';
 import { getChainDisplayName } from './utils';
 
 export function useMultiProvider() {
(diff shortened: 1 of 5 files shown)


src/features/chains/hooks.ts
```diff
@@ -31,22 +28,3 @@ export function useChainDisplayName(chainName: ChainName, shortName = false) {
   const multiProvider = useMultiProvider();
   return getChainDisplayName(multiProvider, chainName, shortName);
 }
-
-/**
- * Returns chainMetadata object with warp core token chains
- */
-export function useTokenChainsMetadata() {
```
jmrossy: @Xaroz Note I made a change after you wrote this and now the ChainMetadata map in the store's multiprovider only includes the token chain anyway. So this hooks is no longer needed :)

src/features/chains/utils.ts
```diff
@@ -36,21 +37,36 @@ export function getChainByRpcUrl(multiProvider: MultiProtocolProvider, url?: str
  * Returns an object that contains the amount of
  * routes from a single chain to every other chain
  */
-export function getAllTokenRoutesFromChain(
+export function getNumRoutesWithSelectedChain(
   warpCore: WarpCore,
-  origin: ChainName,
```
jmrossy: @Xaroz your first version was close but technically num routes _from_ a chain doesn't _necessarily_ equal the num routes _to_ that chain, so we need to treat it as an origin/destination depending on which field it's coming from

PR #356, opened 2024-12-16
feat: added jitoSOL and kySOL warp routes to nexus

xeno097:
Adds the newly deployed SVM routes from Solana to Eclipse for jitoSOL and kySOL

### DIFF
--- a/src/consts/warpRoutes.yaml
+++ b/src/consts/warpRoutes.yaml
@@ -2,5 +2,47 @@
 # These configs will be merged with the warp routes in the configured registry
 # The input here is typically the output of the Hyperlane CLI warp deploy command
 ---
-tokens: []
+tokens:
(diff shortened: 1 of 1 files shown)


src/consts/warpRoutes.yaml
```diff
@@ -2,5 +2,47 @@
 # These configs will be merged with the warp routes in the configured registry
 # The input here is typically the output of the Hyperlane CLI warp deploy command
 ---
-tokens: []
+tokens:
+  - addressOrDenom: G3hWozA3c4iK5oSX9rTWFFDo5rS9W49bjhWDgg3gDPk8
```
jmrossy, line 6: The 'right' way to add these routes is to put them in the registry and then include them in the nexus branch's route whitelist. Do you intend to merge this or was it just for a preview build to test with?

xeno097: preview build to allow testing of the new route as it is not in the registry yet

PR #381, opened 2025-01-20
feat: More chains for TRUMP

nambrot:


src/consts/warpRoutes.yaml
```diff
@@ -2,21 +2,97 @@
 # These configs will be merged with the warp routes in the configured registry
 # The input here is typically the output of the Hyperlane CLI warp deploy command
 ---
+# yaml-language-server: $schema=../schema.json
```
jmrossy, line 5: Don't think this belongs here. There is no schema.json file at /src/

src/consts/warpRoutes.yaml
```diff
@@ -2,21 +2,97 @@
...
-    decimals: 18
     connections:
       - token: sealevel|solanamainnet|21tAY4poz2VXvghqdSQpn9j7gYravQmGpuQi8pHPx9DS
+      - token: ethereum|arbitrum|0x5155EB1bcD30189915cF84717550Acfa537068bF
+      - token: ethereum|flowmainnet|0xD3378b419feae4e3A4Bb4f3349DBa43a1B511760
+      - token: ethereum|form|0x8528bAa7d1d386E7967603e480fa2B558a23644c
+      - token: ethereum|avalanche|0x0e2A546a53678ee8F8605748193a8c114fA0317F
+      - token: ethereum|worldchain|0x0fC7b3518C03BfA5e01995285b1eF3c4B55c8922
+    decimals: 18
+    logoURI: /deployments/warp_routes/TRUMP/logo.png
```
jmrossy, line 17: Does this work when used in this warpRoutes.yaml file? I can't quite remember the exact logic but I'd expect configs in _this_ file to reference paths relative to _this_ app (e.g. something in the `/public` folder)

PR #394, opened 2025-01-22
feat: move to new TONY warp route

tkporter:
- Moves to a new version of the registry that uses the newer version of the TONY/base-solanamainnet warp route.
- PR here intended for unwinding any existing bridged TONY https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/393

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "@emotion/react": "^11.13.3",
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^2.2.0",
-    "@hyperlane-xyz/registry": "6.20.0",
+    "@hyperlane-xyz/registry": "7.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "@emotion/react": "^11.13.3",
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^2.2.0",
-    "@hyperlane-xyz/registry": "6.20.0",
+    "@hyperlane-xyz/registry": "7.0.0",
```
Mo-Hussain, line 20: Was this supposed to be a major release?

tkporter: I chose to yeah as it's technically a breaking change

tkporter: though i guess it depends what we mean by breaking

PR #398, opened 2025-01-23
feat: sync chains and token with url

### LINKED ISSUE #397: Persist chain/token in url params in Warp UI

Xaroz:
- Sync chains select with URL so that there is persistence
- Sync token with URL for persistence
- Added util functions to handle special cases so that app does not crash

fixes #397


src/features/transfer/TransferTokenForm.tsx
```diff
@@ -447,19 +462,28 @@ function WarningBanners() {
...
+  const [defaultOriginQuery, defaultDestinationQuery] = useMultipleQueryParams([
+    WARP_QUERY_PARAMS.ORIGIN,
+    WARP_QUERY_PARAMS.DESTINATION,
+  ]);
+
   return useMemo(() => {
     const firstToken = warpCore.tokens[0];
     const connectedToken = firstToken.connections?.[0];
+
+    if (!isReady) return null;
```
Xaroz: router needs to be ready before we can actually check if any of the URL queries are available, if this is not done, it will override the URL query using the fallback

src/features/tokens/TokenSelectField.tsx
```diff
@@ -26,23 +29,34 @@ export function TokenSelectField({ name, disabled, setIsNft }: Props) {
     const tokensWithRoute = warpCore.getTokensForRoute(origin, destination);
     let newFieldValue: number | undefined;
     let newIsAutomatic: boolean;
+
+    // Use token from persistance
```
jmrossy: typo persistance. I recommend the code Spell checker vscode extension

src/features/tokens/TokenSelectField.tsx
```diff
@@ -26,23 +29,34 @@ export function TokenSelectField({ name, disabled, setIsNft }: Props) {
...
+    if (query.token && typeof query.token === 'string') {
```
jmrossy: Will this re-trigger if a user chooses a different chain but there's still an unrelated token value in the URL

src/features/tokens/TokenSelectField.tsx
```diff
@@ -26,23 +29,34 @@ export function TokenSelectField({ name, disabled, setIsNft }: Props) {
...
       newFieldValue = getIndexForToken(warpCore, tokensWithRoute[0]);
-      newIsAutomatic = true;
       // Multiple possibilities
     } else {
       newFieldValue = undefined;
-      newIsAutomatic = false;
     }
+
+    if (tokensWithRoute.length <= 1) newIsAutomatic = true;
+    else newIsAutomatic = false;
```
jmrossy: newIsAutomatic = (tokensWithRoute.length <= 1)

src/features/tokens/TokenSelectField.tsx
```diff
@@ -26,23 +29,34 @@ export function TokenSelectField({ name, disabled, setIsNft }: Props) {
     const tokensWithRoute = warpCore.getTokensForRoute(origin, destination);
     let newFieldValue: number | undefined;
     let newIsAutomatic: boolean;
+
+    // Use token from persistance
+    if (query.token && typeof query.token === 'string') {
+      // Check if tokenIndex exists
+      const tokenIndex = tryParseAmount(query.token)?.toNumber();
+      if (tokenIndex !== undefined) {
+        const token = warpCore.tokens[tokenIndex];
```
jmrossy: The problem with using token index as the query param value is that the index of a token can change between page loads - like when new routes are added or changed. That means someone could get a token that's different than what the link sender expected. Links should be more durable so I recommend you use token address instead and then search for the token index using the address and chains. If the chains aren't set but the token is, ignore it.

src/features/transfer/TransferTokenForm.tsx
jmrossy: Just curious, why does it need to be ready? Does the URL not have the params in it regardless of the router status?

jmrossy: Also what happens if you return the default values here instead of null (i.e. the ones that we use when the param isn't set)?

Xaroz: removed dependency from next router and instead default to browser native URL handling

src/consts/app.ts
```diff
@@ -13,3 +13,9 @@ export const APP_URL = 'hyperlane-warp-template.vercel.app';
 export const BRAND_COLOR = Color.primary['500'];
 export const BACKGROUND_COLOR = Color.primary['500'];
 export const BACKGROUND_IMAGE = 'url(/backgrounds/main.svg)';
+
+export enum WARP_QUERY_PARAMS {
```
jmrossy: Move this somewhere else. These are consts that we expect people to change to customize the app. Consts for core functionality should probably be somewhere else

src/features/chains/utils.ts
```diff
@@ -62,3 +62,13 @@ export function getNumRoutesWithSelectedChain(
     data,
   };
 }
+
+/**
+ * Check if given chainName has valid chain metadata and return chainName if chain is valid
+ */
+export function getValidChain(
```
jmrossy: Your former `isValidChain` was more clear IMO

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -86,7 +93,13 @@ export function TransferTokenForm() {
   );
 }
 
-function SwapChainsButton({ disabled }: { disabled?: boolean }) {
+function SwapChainsButton({
+  disabled,
+  handleSwapChain,
+}: {
+  disabled?: boolean;
+  handleSwapChain(origin: string, destination: string): void;
```
jmrossy: I don't see this syntax often. To clarify, is this equivalent to:
`handleSwapChain: (origin: string, destination: string)=> void`?

jmrossy: Also nit but I think `onSwapChain` would be more consistent with the nomenclature we use elsewhere

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -126,22 +139,52 @@ function ChainSelectSection({ isReview }: { isReview: boolean }) {
...
+
+  const handleChange = (chainName: string, fieldName: string) => {
+    if (fieldName === 'origin') handleGetTokensRoute(chainName, values.destination);
+    else handleGetTokensRoute(values.origin, chainName);
+    updateQueryParam(fieldName, chainName);
+  };
+
+  const handleSwapChain = (origin: string, destination: string) => {
+    updateQueryParam(WARP_QUERY_PARAMS.ORIGIN, origin);
+    updateQueryParam(WARP_QUERY_PARAMS.DESTINATION, destination);
```
jmrossy, line 193: For the TokenSelectField, you handle the query param setting in the component. But for these others, you're doing it in TransferTokenForm. Let's be consistent to keep code cohesion high

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -126,22 +139,52 @@ function ChainSelectSection({ isReview }: { isReview: boolean }) {
...
+      updateQueryParam(WARP_QUERY_PARAMS.TOKEN, undefined);
+    }
+
+    setFieldValue('tokenIndex', newFieldValue);
+  };
+
+  const handleChange = (chainName: string, fieldName: string) => {
+    if (fieldName === 'origin') handleGetTokensRoute(chainName, values.destination);
+    else handleGetTokensRoute(values.origin, chainName);
+    updateQueryParam(fieldName, chainName);
```
jmrossy, line 188: You've already decided to make a separate `WARP_QUERY_PARAMS` enum which defines the param name, but here you're assuming that fieldName === WARP_QUERY_PARAMS. If you can assume they're always the same, remove WARP_QUERY_PARAMS. If you can't, use WARP_QUERY_PARAMS

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -126,22 +139,52 @@ function ChainSelectSection({ isReview }: { isReview: boolean }) {
     return getNumRoutesWithSelectedChain(warpCore, values.destination, false);
   }, [values.destination, warpCore]);
 
+  const handleGetTokensRoute = (origin: string, destination: string) => {
```
jmrossy: If a function has `get` near the beginning of it's name, that generally implies it's a pure function (i.e. no side-effects). But this does have side-effects: it changes the query param and field value. Suggest renaming, maybe something like `setTokenOnChainChange`

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -449,17 +492,36 @@ function WarningBanners() {
 
 function useFormInitialValues(): TransferFormValues {
   const warpCore = useWarpCore();
+  const parameters = new URLSearchParams(window.location.search);
```
jmrossy: Consider maybe moving the interactions with URLSearchParams to your query params file since you're already doing that with the `update` function and it's better to be consistent. When you introduce a layer of abstraction, it's better to always use it as much as possible instead of creating two paths to the same thing.

jmrossy (commented):
Your general approach to solving the problem here is good but there IMO there are some subtle code smells here (see comments). These aren't obvious issues, just subtleties related to code quality and organization that may possibly hurt maintainability in the long run. But don't be discouraged. You'll naturally get better at nuances like naming, organization, and consistency with more experience.

src/consts/app.ts
Xaroz: Moved under a new file inside const called `core.ts`, I was debating to put it in `common.ts` too, but felt like it didn't really match what the enum meant to do

src/features/chains/utils.ts
Xaroz: The reason for this change is that I want to make sure to get the `chainName` or `undefined` if the chain does not exist, instead of a `boolean`, but I am happy to make adjustments if needed

src/features/transfer/TransferTokenForm.tsx
Xaroz: Kinda derped here, fixed

src/features/transfer/TransferTokenForm.tsx
Xaroz: Moved the logic inside TransferTokenForm for consistency

src/features/transfer/TransferTokenForm.tsx
Xaroz: created a function called `getQueryParams` for this, also use this function in the `updateQueryParams` function

src/features/transfer/TransferTokenForm.tsx
Xaroz: done, did a check for fieldName === WARP_QUERY_PARAMS

src/features/chains/utils.ts
jmrossy: Yeah I can see why this pattern was useful for the places you call it but looking only at this fn, a name like `getValidChain` would make me think this turns ChainMetadata or something.

jmrossy: But it's nbd either way

src/consts/core.ts
```diff
@@ -0,0 +1,5 @@
+export enum WARP_QUERY_PARAMS {
```
jmrossy: Sorry to keep nitpicking but not sure `core` is the winner either. In most projects that would be totally fine but `core` in Hyperlane context almost always refers to a core contract deployment, so this could cause brief confusion. How about `params`,`args`, or `queries`?

jmrossy (approved):

src/consts/core.ts
Xaroz: I went with args, felt a little bit more general and then we can add more things in the file if needed

src/features/chains/utils.ts
Xaroz: So I leveraged the `multiProvider.tryGetChainName()` and used that instead of the `getMetadata` function, but since that function does not allow `null` values as `params`, this `getValidChain` function was renamed to `tryGetValidChainName`, this way it suits my needs more

PR #436, opened 2025-02-14
feat: add CDX/base-solanamainnet to nexus

tkporter:
- Adds CDX/base-solanamainnet to Nexus


paulbalaji (changes requested):
holding off til go-live

paulbalaji (approved):

PR #453, opened 2025-03-05
feat: add warp deploy config

Xaroz:
This PR adds the warp deploy config to the application store, this will be used later on for some additional checking of warp routes.
The warp deploy config map is of type: `Record<ChainName, Record<string, HypTokenRouterConfig>>` (where the string is the token `addressOrDenom`)

### DIFF
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -14,7 +14,7 @@ import { config } from '../consts/config';
 import { logger } from '../utils/logger';
 import { assembleChainMetadata } from './chains/metadata';
 import { FinalTransferStatuses, TransferContext, TransferStatus } from './transfer/types';
-import { assembleWarpCoreConfig } from './warpCore/warpCoreConfig';
+import { assembleWarpCoreConfig, WarpDeployConfigChainAddressMap } from './warpCore/warpCoreConfig';
 
 // Increment this when persist state has breaking changes
 const PERSIST_STATE_VERSION = 2;
@@ -33,11 +33,13 @@ export interface AppState {
   multiProvider: MultiProtocolProvider;
   registry: IRegistry;
   warpCore: WarpCore;
+  warpDeployConfig: WarpDeployConfigChainAddressMap;
   setWarpContext: (context: {
     registry: IRegistry;
     chainMetadata: ChainMap<ChainMetadata>;
     multiProvider: MultiProtocolProvider;
     warpCore: WarpCore;
+    warpDeployConfig: WarpDeployConfigChainAddressMap;
   }) => void;
 
   // User history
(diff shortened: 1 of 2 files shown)


src/features/store.ts
```diff
@@ -177,20 +188,24 @@ async function initWarpContext({
...
   chainMetadataOverrides: ChainMap<Partial<ChainMetadata> | undefined>;
   warpCoreConfigOverrides: WarpCoreConfig[];
+  warpDeployConfig: WarpDeployConfigChainAddressMap;
 }) {
   try {
-    const coreConfig = await assembleWarpCoreConfig(warpCoreConfigOverrides);
-    const chainsInTokens = Array.from(new Set(coreConfig.tokens.map((t) => t.chainName)));
     // Pre-load registry content to avoid repeated requests
     await registry.listRegistryContent();
+    const { warpCoreConfig: coreConfig, warpDeployConfig } = await assembleWarpCoreConfig(
```
nambrot, line 196: Do we need to move the assembly of the warpDeployConfig inside `assembleWarpCoreConfig`? Imo the name is not reflective

src/features/store.ts
```diff
@@ -33,11 +33,13 @@ export interface AppState {
   multiProvider: MultiProtocolProvider;
   registry: IRegistry;
   warpCore: WarpCore;
+  warpDeployConfig: WarpDeployConfigChainAddressMap;
```
nambrot, line 36: Let's not name things that are not `WarpDeployConfig` `warpDeployConfig`. Consider using `WarpDeployConfigIndex` or `WarpDeployConfigMap`

src/features/warpCore/warpCoreConfig.ts
```diff
@@ -18,6 +35,9 @@ export function assembleWarpCoreConfig(storeOverrides: WarpCoreConfig[]): WarpCo
   const filteredRegistryTokens = filteredRegistryConfigValues.map((c) => c.tokens).flat();
   const filteredRegistryOptions = filteredRegistryConfigValues.map((c) => c.options).flat();
 
+  const warpDeployConfigMap = await registry.getWarpDeployConfigs();
+  const warpDeployConfig = assembleWarpDeployConfig(filteredRegistryConfigMap, warpDeployConfigMap);
```
nambrot, line 39: Can't we move this out of this function and pass `tokens` as a parameter?

Xaroz:
Closing this PR as upcoming feature from @yorhodes in this [PR](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5492) will address the reason why warp deploy config was needed (to do limit checking) and leveraging our SDK is probably a better solution.

PR #458, opened 2025-03-12
chore: add rpc url for sonicsvm to env variable

Xaroz:
Add new environment variable for NEXT_PUBLIC_SONICSVM_RPC_URL override because main RPC is down

### DIFF
--- a/src/consts/chains.ts
+++ b/src/consts/chains.ts
@@ -37,6 +37,10 @@ export const chains: ChainMap<ChainMetadata & { mailbox?: Address }> = {
   sonicsvm: {
     ...sonicsvm,
     mailbox: sonicsvmAddresses.mailbox,
+    // Including rpc override because the main Sonic public RPC is down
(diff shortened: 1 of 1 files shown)


src/consts/chains.ts
```diff
@@ -37,6 +37,10 @@ export const chains: ChainMap<ChainMetadata & { mailbox?: Address }> = {
   sonicsvm: {
     ...sonicsvm,
     mailbox: sonicsvmAddresses.mailbox,
+    // Including rpc override because the main Sonic public RPC is down
+    rpcUrls: process.env.NEXT_PUBLIC_SONIC_RPC_URL
```
tkporter: can you change this to `SONICSVM`? as we have a chain `sonic` that is a different EVM chain

tkporter (approved):

src/consts/chains.ts
Xaroz: updated

PR #463, opened 2025-03-12
feat: lorenzo routes

Xaroz:
- Add enzoBTC/bsc-hyperevm route
- Add stBTC/bsc-hyperevm route


Xaroz:
Tested bsc <> hyperevm for both enzoBTC and stBTC

PR #473, opened 2025-03-14
feat: cosmos-module

troykessler:
This PR imports the new typescript sdk to connect to cosmos chains which have the new hyperlane cosmos module enabled

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "viem": "^2.21.41",
     "zustand": "^4.4",
     "@hyperlane-xyz/cosmos-sdk": "file:/Users/troykessler/work/hyperlane/hyperlane-monorepo/typescript/cosmos-sdk",
+    "@hyperlane-xyz/sdk": "file:/Users/troykessler/work/hyperlane/hyperlane-monorepo/typescript/sdk",
     "@hyperlane-xyz/utils": "file:/Users/troykessler/work/hyperlane/hyperlane-monorepo/typescript/utils"
   }
 }
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
   "scripts": {
     "clean": "rm -rf dist cache .next",
     "dev": "next dev --port 8080",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -18,9 +18,9 @@
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^2.2.0",
     "@hyperlane-xyz/registry": "10.15.0",
-    "@hyperlane-xyz/sdk": "8.8.0",
+    "@hyperlane-xyz/sdk": "file:/Users/troykessler/work/hyperlane/hyperlane-monorepo/typescript/sdk",
     "@hyperlane-xyz/utils": "8.8.0",
-    "@hyperlane-xyz/widgets": "8.8.0",
+    "@hyperlane-xyz/widgets": "file:/Users/troykessler/work/hyperlane/hyperlane-monorepo/typescript/widgets",
```
paulbalaji: maybe ask @yorhodes to publish preview packages that we can reuse here

tokens:.yaml.yml
```diff
@@ -0,0 +1,14 @@
+tokens:
```
Xaroz, line 1: this file is probably not needed right?

package.json
```diff
@@ -18,9 +18,9 @@
     "@emotion/styled": "^11.13.0",
```
Xaroz, line 18: would appreciate if we can have a beta/preview package we can test with

src/features/wallet/context/StarknetWalletContext.tsx
```diff
@@ -0,0 +1,17 @@
+// import { getStarknetChains } from "@hyperlane-xyz/widgets";
+import { PropsWithChildren } from 'react';
+
+export function StarknetWalletContext({ children }: PropsWithChildren<unknown>) {
```
Xaroz, line 4: merging with main will fix this

PR #476, opened 2025-03-17
feat: add default destination chain to config

### LINKED ISSUE #475: Add defaultDestinationChain to config
allow setting default destination chain

Xaroz:
Include `defaultDestinationChain` to config allowing the UI to have both default origin and destination chain

fixes #475

### DIFF
--- a/src/consts/config.ts
+++ b/src/consts/config.ts
@@ -15,6 +15,7 @@ interface Config {
   addressBlacklist: string[]; // A list of addresses that are blacklisted and cannot be used in the app
   chainWalletWhitelists: ChainMap<string[]>; // A map of chain names to a list of wallet names that work for it
   defaultOriginChain: string | undefined; // The initial origin chain to show when app first loads
+  defaultDestinationChain: string | undefined; // The initial destination chain to show when app first loads
   enableExplorerLink: boolean; // Include a link to the hyperlane explorer in the transfer modal
   isDevMode: boolean; // Enables some debug features in the app
   registryUrl: string | undefined; // Optional URL to use a custom registry instead of the published canonical version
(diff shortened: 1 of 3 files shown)


src/features/tokens/hooks.ts
```diff
@@ -69,22 +69,24 @@ export function getInitialTokenIndex(
   originQuery?: string,
   destinationQuery?: string,
   defaultOriginToken?: Token,
+  defaultDestinationChain?: string,
 ): number | undefined {
   const firstToken = defaultOriginToken || warpCore.tokens[0];
-  const connectedToken = firstToken.connections?.[0];
+  const connectedToken = firstToken.connections?.[0].token;
```
cmcewen, line 75: very nit but i find this diff a little hard to review because while you are correct that probably the old name was not specific, i am looking for functionality changes but all that i can see if making the statement below slightly simpler

cmcewen (approved):

src/consts/config.ts
```diff
@@ -15,6 +15,7 @@ interface Config {
   addressBlacklist: string[]; // A list of addresses that are blacklisted and cannot be used in the app
   chainWalletWhitelists: ChainMap<string[]>; // A map of chain names to a list of wallet names that work for it
   defaultOriginChain: string | undefined; // The initial origin chain to show when app first loads
+  defaultDestinationChain: string | undefined; // The initial destination chain to show when app first loads
```
nambrot, line 18: Should we generalize this more to a specific combo of origin/destination/token?

troykessler (approved):
I think this is a good improvement to the config, no additions to the code from my side

src/consts/config.ts
Xaroz: I personally like the idea of having the freedom of setting them individually

PR #494, opened 2025-03-31
chore: ousdt filter out chains with no connections

Xaroz:
Filter out chains with no connections in the config


src/features/chains/metadata.ts
```diff
@@ -53,6 +56,7 @@ export async function assembleChainMetadata(
     ),
   );
 
+  console.log('registry', registryChainMetadata);
```
paulbalaji:
```suggestion
```

nambrot (approved):

PR #517, opened 2025-04-17
feat: added revoke tx before approval to avoid usdt approval issues

xeno097:
Updates the logic for sending a transferRemote transaction to revoke any approval before sending the real approval. Some tokens might require approvals to be 0 before setting a different value (this is the case with USDT), making any other approval tx fail until the allowance is reset.

Note:
A follow-up PR will be done in the SDK to move this logic there

### DIFF
--- a/src/features/transfer/useTokenTransfer.ts
+++ b/src/features/transfer/useTokenTransfer.ts
@@ -1,4 +1,10 @@
-import { TypedTransactionReceipt, WarpCore, WarpTxCategory } from '@hyperlane-xyz/sdk';
+import {
+  TOKEN_STANDARD_TO_PROVIDER_TYPE,
+  TypedTransactionReceipt,
+  WarpCore,
+  WarpTxCategory,
+  WarpTypedTransaction,
+} from '@hyperlane-xyz/sdk';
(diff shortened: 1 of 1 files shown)


src/features/transfer/useTokenTransfer.ts
```diff
@@ -141,12 +147,35 @@ async function executeTransfer({
...
+        recipient: originTokenAmount.token.addressOrDenom,
+        weiAmountOrId: 0,
+      });
+
+      txs.unshift({
+        category: WarpTxCategory.Approval,
+        type: TOKEN_STANDARD_TO_PROVIDER_TYPE[originTokenAmount.token.standard],
+        transaction: revokeApprovalTx,
+      } as WarpTypedTransaction);
+    }
```
Xaroz, line 178: with this logic, users will require to sign 3 transactions each time an approval transaction is required even if the allowance is already 0, for now I think it is okay but probably need to change it in the future

Xaroz (approved):
approved with a comment

xeno097:
closing because updating the sdk once https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5973 is merged will fix it

PR #518, opened 2025-04-18
feat: check for all RPC healths in ChainConnectionWarning for EVM chains

Xaroz:
Now will only display the `ChainConnectionWarning` banner when all of the chains are unhealthy for EVM chains. This is because now we have fallback transports for EVM chains. Other VMs behavior remain unchanged and only use the first RPC


cmcewen (approved):
imo this function is a good candidate for a unit test

PR #521, opened 2025-04-21
chore: partial registry and change image and RPC url logic

Xaroz:
- Change `GithubRegistry` to use `PartialRegistry`
- Change `Token` and `Chain` logo link to hard-coded link path
- Modify RPC URL overrides logic to: if `EVM` chain, add override at the end, otherwise in the front
- Upgrade registry

### DIFF
--- a/src/components/icons/TokenIcon.tsx
+++ b/src/components/icons/TokenIcon.tsx
@@ -1,9 +1,8 @@
-import { IRegistry } from '@hyperlane-xyz/registry';
 import { IToken } from '@hyperlane-xyz/sdk';
 import { isHttpsUrl, isRelativeUrl } from '@hyperlane-xyz/utils';
 import { Circle } from '@hyperlane-xyz/widgets';
 import { useState } from 'react';
-import { useStore } from '../../features/store';
+import { links } from '../../consts/links';
 
 interface Props {
   token?: IToken | null;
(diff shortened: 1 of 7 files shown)


src/components/icons/TokenIcon.tsx
```diff
@@ -40,11 +38,11 @@ export function TokenIcon({ token, size = 32 }: Props) {
...
 }
 
-function getImageSrc(registry: IRegistry, token?: IToken | null) {
+function getImageSrc(token?: IToken | null) {
   if (!token?.logoURI) return null;
   // If it's a valid, direct URL, return it
   if (isHttpsUrl(token.logoURI)) return token.logoURI;
   // Otherwise assume it's a relative URL to the registry base
-  if (isRelativeUrl(token.logoURI)) return registry.getUri(token.logoURI);
+  if (isRelativeUrl(token.logoURI)) return `${links.imgPath}${token.logoURI}`;
```
nambrot, line 46: I wonder if we should test if `getUri` exists, if so use that, otherwise use the hardcoded one. I was originally thinking that this hardcode would only exist for TGE

src/consts/links.ts
```diff
@@ -13,4 +13,5 @@ export const links = {
   privacyPolicy: 'https://hyperlane.xyz/privacy-policy',
   bounty:
     'https://github.com/search?q=org:hyperlane-xyz+label:bounty+is:open+is:issue&type=issues&s=&o=desc',
+  imgPath: 'https://raw.githubusercontent.com/hyperlane-xyz/hyperlane-registry/main',
```
ltyu, line 16: does this mean that we won't see the logo when local testing an unmerged route/logo?

Xaroz: yes, this is the same even for now, unless you had a registry release or pointed to a specific URL you wouldn't see the logo

src/components/icons/TokenIcon.tsx
Xaroz: `getUri` does exist for PartialRegistry but it returns:
`_partial_registry__/deployments/warp_routes/<ChainName>/<logo>`

nambrot (approved):

ltyu (approved):

PR #535, opened 2025-04-24
feat: updated sdk package version to fix approval issues

xeno097:
Updates the SDK to the latest version to fix approval issues for tokens that require allowance to be set to 0 before setting a new approval value. context https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5973. 

Adds support for revoke transactions and related messages to display to the user


paulbalaji (approved):

Xaroz (approved):
I am wondering if we should include some sort of information explaining to users what is a revoke transaction or maybe link to somewhere that explains it

PR #536, opened 2025-04-25
chore: cosmos addresses

troykessler:
This PR improves the way cosmos addresses are displayed. In cosmos there is like in Ethereum a private public key pair, but for every cosmos chain there is a different address prefix. Example Osmosis osmo1sduu..., Cosmos Hub cosmos1j48f33.., KYVE kyve1jhf738... and so on. Right now always the address of the first chain gets displayed which often has nothing to do with the chain the user has funds on and wants to swap from/to. This has caused confusion in the past.

This PR addresses this issue by showing the Cosmos Hub address by default if the user is connected with a cosmos wallet since Cosmos Hub is still the standard in the cosmos ecosystem. Once the user selects a cosmos origin chain that address is displayed instead.

TODO: once the monorepo PR https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6022 is merged and released we can switch out the dependencies to the newest hyperlane widgets version to make it ready for review.

### DIFF
--- a/src/features/transfer/TransferTokenForm.tsx
+++ b/src/features/transfer/TransferTokenForm.tsx
@@ -198,6 +198,7 @@ function ChainSelectSection({ isReview }: { isReview: boolean }) {
     updateQueryParam(WARP_QUERY_PARAMS.ORIGIN, origin);
     updateQueryParam(WARP_QUERY_PARAMS.DESTINATION, destination);
     setTokenOnChainChange(origin, destination);
+    setOriginChainName(origin);
   };
 
   return (
--- a/src/features/transfer/TransferTokenForm.tsx
+++ b/src/features/transfer/TransferTokenForm.tsx
@@ -55,8 +55,17 @@ export function TransferTokenForm() {
   const multiProvider = useMultiProvider();
   const warpCore = useWarpCore();
 
+  const { originChainName, setOriginChainName } = useStore((s) => ({
+    originChainName: s.originChainName,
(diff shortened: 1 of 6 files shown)


src/features/transfer/TransferTokenForm.tsx
```diff
@@ -172,10 +185,12 @@ function ChainSelectSection({ isReview }: { isReview: boolean }) {
   };
 
   const handleChange = (chainName: string, fieldName: string) => {
-    if (fieldName === WARP_QUERY_PARAMS.ORIGIN)
+    if (fieldName === WARP_QUERY_PARAMS.ORIGIN) {
       setTokenOnChainChange(chainName, values.destination);
-    else if (fieldName === WARP_QUERY_PARAMS.DESTINATION)
+      setOriginChainName(chainName);
```
Xaroz, line 190: you also have to add for when chains are swapped (onSwapChain below)

Xaroz (commented):
small comment, mostly look good, although not a fan keeping two different state that does the same thing, ex: `formik.values.origin` and `store.originChainName` but I understand it is necessary without needing to lift formik up

troykessler:
> small comment, mostly look good, although not a fan keeping two different state that does the same thing, ex: `formik.values.origin` and `store.originChainName` but I understand it is necessary without needing to lift formik up

Yes tried it to do it with Formik first but it is way more breaking unfortunately

cmcewen (approved):

PR #544, opened 2025-04-28
feat: search full list token

Xaroz:
This PR adds the ability for users to search tokens that have no routes from the current origin and destination, closes #search tokens by route

closes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5708

- Now shows a list of tokens with a slighty grayed out color, this will also be searchable
- When clicking one of the unsupported tokens, a list of available chains will show up
- When clicking on one of the chain, it will set the current token and the origin/destination in the form
- Create a map of symbols and chain map of tokens to accomplish this
- Use [jsdelivr](https://www.jsdelivr.com/) to serve imgs from the github registry
- Minor refactor and drive by changes


troykessler (approved):
UI works perfectly on my machine, no additional comments regarding the implementation

src/features/tokens/TokenListModal.tsx
```diff
@@ -121,51 +140,56 @@ export function TokenList({
...
+          </div>
+        </button>
+      ))}
+      <UnsupportedRouteTokenList
+        unsupportedRouteTokensBySymbolMap={unsupportedRouteTokensBySymbolMap}
+        origin={origin}
+        destination={destination}
+        onSelectUnsuportedRoute={onSelectUnsuportedRoute}
+      />
+      {tokens.length === 0 && isObjEmpty(unsupportedRouteTokensBySymbolMap) && (
```
ljankovic-txfusion: Because `unsupportedRouteTokensBySymbolMap` is derived by filtering `tokens`, there's a potential case where `tokens.length > 0` but `isObjEmpty(unsupportedRouteTokensBySymbolMap) `is true (if all tokens are filtered out). This could lead to both `UnsupportedRouteTokenList` and` "No tokens found"` components being rendered at the same time. Since `UnsupportedRouteTokenList` takes `unsupportedRouteTokensBySymbolMap` as props, checking if that map is empty seems sufficient on its own.

src/features/tokens/TokenListModal.tsx
```diff
@@ -174,3 +198,86 @@ export function TokenList({
...
+          <div className="shrink-0">
+            <TokenIcon token={tokenInformation} size={30} />
+          </div>
+          <div className="ml-2 shrink-0 text-left">
+            <div className="text-sm">{tokenInformation.symbol || 'Unknown'}</div>
+            <div className="text-xs text-gray-500">{tokenInformation.name || 'Unknown'}</div>
+          </div>
+          <Image
+            src={InfoIcon}
+            alt=""
```
ljankovic-txfusion: can we maybe say `alt={tokenInformation.symbol}`

src/features/tokens/TokenListModal.tsx
Xaroz: how would "No tokens found" be rendered if `tokens.length > 0`?

src/features/tokens/TokenListModal.tsx
Xaroz: This is not exactly the symbol but an icon talking about unsupported route, so I did that instead

xeno097 (approved):

src/features/store.ts
```diff
@@ -96,9 +99,15 @@ export const useStore = create<AppState>()(
         chainMetadata: chainMetadata,
       }),
       warpCore: new WarpCore(new MultiProtocolProvider({}), []),
-      setWarpContext: ({ registry, chainMetadata, multiProvider, warpCore }) => {
+      setWarpContext: ({
```
antigremlin: Due to typing, this could be just

```typescript
setWarpContext: (context) => {
  logger.debug(...);
  set(context);
}
```

src/features/store.ts
```diff
@@ -165,10 +175,18 @@ export const useStore = create<AppState>()(
             logger.error('Error during hydration', error);
             return;
           }
-          initWarpContext(state).then(({ registry, chainMetadata, multiProvider, warpCore }) => {
-            state.setWarpContext({ registry, chainMetadata, multiProvider, warpCore });
-            logger.debug('Rehydration complete');
-          });
+          initWarpContext(state).then(
+            ({ registry, chainMetadata, multiProvider, warpCore, tokensBySymbolChainMap }) => {
```
antigremlin: Same here, this could be

```typescript
initWarpContext(state).then((context) => {
  state.setWarpContext(context);
  logger.debug(...);
});
```

src/features/store.ts
```diff
@@ -196,7 +214,9 @@ async function initWarpContext({
     );
     const multiProvider = new MultiProtocolProvider(chainMetadataWithOverrides);
     const warpCore = WarpCore.FromConfig(multiProvider, coreConfig);
-    return { registry, chainMetadata, multiProvider, warpCore };
+
+    const tokensBySymbolChainMap = assembleTokensBySymbolChainMap(warpCore.tokens, multiProvider);
+    return { registry, chainMetadata, multiProvider, warpCore, tokensBySymbolChainMap };
```
antigremlin, line 207: Could this file benefit from declaring a type for the warp context?

```typescript
interface WarpContext {
  registry: IRegistry;
  chainMetadata: ChainMap<ChainMetadata>;
  warpCore: WarpCore;
  tokensBySymbolChainMap: Record<string, TokenChainMap>;
}
```

This we can do

```typescript
setWarpContext: (context: WarpContext) => void;
...
async function initWarpContext({
  registry,
  chainMetadataOverrides,
  warpCoreConfigOverrides,
}: AppState): Promise<WarpContext> {}
```

src/features/tokens/TokenListModal.tsx
```diff
@@ -83,19 +96,25 @@ export function TokenList({
...
 }: {
   origin: ChainName;
   destination: ChainName;
   searchQuery: string;
   onSelect: (token: IToken) => void;
+  onSelectUnsupportedRoute: (token: Token, origin: string) => void;
 }) {
   const multiProvider = useMultiProvider();
   const warpCore = useWarpCore();
+  const { tokensBySymbolChainMap } = useStore((s) => ({
```
antigremlin: Why not `const tokensBySymbolChainMap = useStore((s) => s.tokenBySymbolChainMap);`?

src/features/tokens/TokenListModal.tsx
ljankovic-txfusion: Yeah, you are right at the closer look, maybe just check `isObjEmpty(unsupportedRouteTokensBySymbolMap)` since `unsupportedRouteTokensBySymbolMap` is derived state from `tokens`.

btw this is a bit too verbose for inline code for my taste, you could extract the content of `if` statement in separate `const hasUnsupportedRouteTokensBySymbolMap`, as you wish

ljankovic-txfusion (approved):

src/features/tokens/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+// Map of token symbols and token chain map
+// Symbols are not duplicated to avoid the same symbol from being shown
+// TokenChainMap: An object containing token information and a map
+// chain names with its metadata and the related token
+export function assembleTokensBySymbolChainMap(
+  tokens: Token[],
+  multiProvider: MultiProtocolProvider,
+): Record<string, TokenChainMap> {
+  const multiChainTokens = tokens.filter((t) => t.isMultiChainToken());
+  return multiChainTokens.reduce((acc, token) => {
```
antigremlin: Reduce needs explicit typing here in order to type-check the reducer:

```typescript
multiChainTokens.reduce<Record<string, TokenChainmap>>((acc, token) => {});
```

src/utils/queryParams.ts
```diff
@@ -17,3 +17,7 @@ export function updateQueryParam(key: string, value?: string | number) {
   const newUrl = `${window.location.pathname}?${params.toString()}`;
   window.history.replaceState({}, '', newUrl);
 }
+
+export function updateMultipleQueryParam(params: Array<{ key: string; value?: string | number }>) {
```
antigremlin: Why not

```typescript
export function updateQueryParams(params: Record<string, string | number>) {
  for (const [key, value] of Object.entries(params)) {
    updateQueryParam(key, value);
  }
}
```

and then the usage becomes

```typescript
updateQueryParams({
  [WARP_QUERY_PARAMS.ORIGIN]: origin,
  [WARP_QUERY_PARAMS.DESTINATION]: destination,
  [WARP_QUERY_PARAMS.TOKEN]: token.addressOrDenom,
});
```

antigremlin (commented):
Please see my comments

src/features/store.ts
Xaroz: I like it! Improves the typing interface a lot

src/features/tokens/TokenListModal.tsx
```diff
@@ -121,51 +138,56 @@ export function TokenList({
...
+    const tokenSymbols = tokens.map((item) => item.token.symbol);
+    const q = searchQuery?.trim().toLowerCase();
+    return objFilter(tokensBySymbolChainMap, (symbol, value): value is TokenChainMap => {
+      return (
+        !tokenSymbols.includes(symbol) &&
+        (value.tokenInformation.name.toLowerCase().includes(q) ||
+          value.tokenInformation.symbol.toLowerCase().includes(q))
+      );
+    });
+  }, [tokens, tokensBySymbolChainMap, searchQuery]);
```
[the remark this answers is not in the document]
antigremlin: Makes sense, to my mind

antigremlin (approved):

src/features/tokens/TokenListModal.tsx
Xaroz: > Yeah, you are right at the closer look, maybe just check isObjEmpty(unsupportedRouteTokensBySymbolMap) since unsupportedRouteTokensBySymbolMap is derived state from tokens.

I don't think this is gonna work how you think, since unsupportedRouteTokensBySymbolMap filter outs the symbols that are inside the supported token, ex:
image

> btw this is a bit too verbose for inline code for my taste, you could extract the content of if statement in separate const hasUnsupportedRouteTokensBySymbolMap, as you wish

good point, done

PR #547, opened 2025-04-30
chore: add `wfragSOL` & `wfragJTO`

christopherbrumm:


Xaroz (commented):
you need to bump the version of the registry in main and then it will show up in this [PR](#546) (hopefully) and after that you can merge this

Xaroz (approved):

PR #569, opened 2025-05-12
feat: bump registry+new coti routes

xeno097:


src/consts/warpRouteWhitelist.ts
```diff
@@ -214,4 +214,8 @@ export const warpRouteWhitelist: Array<string> | null = [
   'PEPE/apechain-arbitrum',
   'GG/apechain-arbitrum',
   'Boop/apechain-arbitrum',
+
+  // Coti batch
+  'WBTC/coti-ethereum',
+  'ETH/coti-ethereum',
```
Xaroz: I suggest adding these to the ETH and wbtc sections if they exist

Xaroz (approved):

PR #571, opened 2025-05-15
feat: mirai route

cmcewen:


Xaroz (approved):

Xaroz:
would prefer we update registry on main first

PR #599, opened 2025-05-26
chore: prevent transfer from being sent to router addresses

### LINKED ISSUE #595: prevent transfer to warp route addresses

Xaroz:
Fixes #595, this PR prevents the user from sending funds to a warp route address from the tokens list

- Create a `routerAddressesByChainMap` variable in store that host all the known `addressOrDenom` in a given chain
- Check if recipient is one of these addresses and if so, prevent transfer from being made

### DIFF
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -7,7 +7,7 @@ import {
   WarpCore,
   WarpCoreConfig,
 } from '@hyperlane-xyz/sdk';
-import { objFilter } from '@hyperlane-xyz/utils';
+import { isNullish, objFilter } from '@hyperlane-xyz/utils';
 import { toast } from 'react-toastify';
 import { create } from 'zustand';
 import { persist } from 'zustand/middleware';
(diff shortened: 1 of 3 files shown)


src/features/store.ts
```diff
@@ -26,6 +26,7 @@ interface WarpContext {
   multiProvider: MultiProtocolProvider;
   warpCore: WarpCore;
   tokensBySymbolChainMap: Record<string, TokenChainMap>;
+  routerAddresses: Set<string>;
```
nambrot: I would maybe use a more descriptive name like `warpRouterAddressesToPreventTransfers` and then leave a comment for it's intended use here

nambrot (approved):

src/features/store.ts
xeno097: Wouldn't it be better to have a mapping/record where the key is the chain name and the value is a set with the warp route token addresses deployed on that chain instead of a single set? it might not be true that an address is a warp token on all chains

Xaroz: I feel like this is already a very descriptive name of what it actually is, but I added a comment above it for its current uses

PR #611, opened 2025-05-30
feat: multi-collateral token setup

Xaroz:
This PR features an initial setup for Multi-Collateral tokens for transfers
fixes [ENG-1683](https://linear.app/hyperlane-xyz/issue/ENG-1683/warp-ui-should-handle-multiple-available-routes-for-a)

- Filter out tokens with the same collateral address in origin and destination and only show one of them
- A map containing the collateral address of origin and destination with an array of the same tokens is added, but not used yet
- When submitting, it will check for the same logic and then return the highest balance token represented by the state `multiCollateralToken`
- Transfers will be done with this state otherwise uses the one in the form

You can test this by using `Arbitrum` <> `Optimism` with `USDC`


src/consts/warpRoutes.yaml
```diff
@@ -2,5 +2,23 @@
...
+    symbol: USDC
+  - addressOrDenom: '0x296B0c551c65CB83452f10EA3922824E7b9b3F4C'
+    chainName: arbitrum
+    collateralAddressOrDenom: '0xaf88d065e77c8cC2239327C5EDb3A432268e5831'
+    connections:
+      - token: ethereum|optimism|0x296B0c551c65CB83452f10EA3922824E7b9b3F4C
+    decimals: 6
+    name: USD Coin
+    standard: EvmHypCollateral
+    symbol: USDC
```
xeno097: Should this be here?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -619,3 +671,51 @@ async function validateForm(
...
+  );
+
+  results.forEach((result, i) => {
+    if (result.status === 'fulfilled') {
+      tokenBalances.push({
+        token: tokensWithSameCollateralAddresses[i].originToken,
+        balance: result.value,
+      });
+    }
+  });
```
xeno097: why not `tokenBalances = results .filter(...).map(...)`

src/features/tokens/utils.ts
```diff
@@ -31,3 +39,56 @@ export function assembleTokensBySymbolChainMap(
...
+  )
+    return false;
+
+  return true;
+}
+
+export function getTokensWithSameCollateralAddresses(
+  warpCore: WarpCore,
+  origin: Token,
+  destination: IToken | Token,
```
xeno097: Why is this needed here?

src/features/tokens/utils.ts
```diff
@@ -31,3 +39,56 @@ export function assembleTokensBySymbolChainMap(
...
+) {
+  const originCollateralAddress = origin.collateralAddressOrDenom?.toLowerCase();
+  const destinationCollateralAddress = destination.collateralAddressOrDenom?.toLowerCase();
```
xeno097: I think that is would be better to use normalizeAddress here from the utils package because not all addresses are equivalent when lowercased (is svm the address might be invalid)

src/features/tokens/utils.ts
```diff
@@ -31,3 +39,56 @@ export function assembleTokensBySymbolChainMap(
...
+
+      if (!destinationToken) return false;
+      const isMultiCollateralToken = isValidMultiCollateralToken(originToken, destinationToken);
+      if (!destinationToken || !isMultiCollateralToken) return false;
+
+      // asserting because isValidMultiCollateralToken already checks for existence of collateralAddressOrDenom
+      if (
+        originToken.collateralAddressOrDenom!.toLowerCase() !== originCollateralAddress ||
+        destinationToken.collateralAddressOrDenom!.toLowerCase() !== destinationCollateralAddress
+      )
```
xeno097: I'd use the eqAddress function from the utils package here

src/consts/warpRoutes.yaml
Xaroz: nope, will remove after approval, this is more for testing

src/features/transfer/TransferTokenForm.tsx
Xaroz: I need a way to get the `token` that was used for the `balance` check, I wouldn't be able to do that with filter since I don't have the `index` to match the `tokensWithSameCollateralAddress`

Xaroz: filtering will mess up the indexing to get the token in `tokensWithSameCollateralAddresses`, I could map it first, but this is way shorter and faster

xeno097 (approved):

src/features/tokens/TokenListModal.tsx
```diff
@@ -13,7 +13,10 @@ import { useMultiProvider } from '../chains/hooks';
 import { getChainDisplayName } from '../chains/utils';
 import { useStore } from '../store';
 import { useWarpCore } from './hooks';
-import { TokenChainMap } from './utils';
+import { MultiCollateralTokenMap } from './types';
+import { isValidMultiCollateralToken, TokenChainMap } from './utils';
+
+// type MultiCollateralToken = { baseToken: Token; tokens: Token[] };
```
cmcewen: unneeded comment?

src/features/tokens/TokenListModal.tsx
```diff
@@ -135,6 +137,41 @@ export function TokenList({
         })
         // Hide/show disabled tokens
         .filter((t) => (config.showDisabledTokens ? true : !t.disabled))
+        // Remove the tokens that have the same collateral addresses
```
cmcewen: if you find yourself leaving a comment like this - often nice to make this a named function: `deduplicateMulticollateralTokens` or something

src/features/tokens/utils.ts
```diff
@@ -31,3 +40,66 @@ export function assembleTokensBySymbolChainMap(
     return acc;
   }, {});
 }
+
+export function isValidMultiCollateralToken(originToken: Token, destination: ChainName | IToken) {
```
cmcewen, line 38: could you add some unit tests for this? please AI it

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -69,15 +71,28 @@ export function TransferTokenForm() {
   const [isReview, setIsReview] = useState(false);
   // Flag for check current type of token
   const [isNft, setIsNft] = useState(false);
+  // This state is used for when the formik token is different from
+  // the token with highest collateral in a multi-collateral token setup
+  const [multiCollateralToken, setMultiCollateralToken] = useState<Token | null>(null);
```
cmcewen: maybe something more generic like `routeOverrideToken`?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -69,15 +71,28 @@ export function TransferTokenForm() {
...
     close: closeConfirmationModal,
     isOpen: isConfirmationModalOpen,
   } = useModal();
 
-  const validate = (values: TransferFormValues) =>
-    validateForm(warpCore, values, accounts, routerAddressesByChainMap);
+  const validate = async (values: TransferFormValues) => {
+    const result = await validateForm(warpCore, values, accounts, routerAddressesByChainMap);
+
+    // Unless this is done, the review and the transfer would contain
```
cmcewen, line 92: should this not live inside the validateForm function itself?

cmcewen: you could return a tuple or something to not pass the setter

Xaroz: great idea!

src/features/tokens/TokenListModal.tsx
```diff
@@ -108,34 +108,34 @@ export function TokenList({
...
-          return (
-            t.token.name.toLowerCase().includes(q) ||
-            t.token.symbol.toLowerCase().includes(q) ||
-            t.token.addressOrDenom.toLowerCase().includes(q)
-          );
-        })
-        // Hide/show disabled tokens
-        .filter((t) => (config.showDisabledTokens ? true : !t.disabled))
-    );
+    const tokens = multiChainTokens
```
cmcewen, line 117: nit and you don't have to do this, but similar with chaining these filter functions  as named ones to make this more readable

src/features/tokens/utils.ts
```diff
@@ -1,10 +1,22 @@
...
+} from '@hyperlane-xyz/sdk';
+import { eqAddress, normalizeAddress } from '@hyperlane-xyz/utils';
+import { MultiCollateralTokenMap } from './types';
 
 export type TokenChainMap = {
   chains: ChainMap<{ token: Token; metadata: ChainMetadata | null }>;
   tokenInformation: Token;
 };
 
+type Tokens = Array<{ token: Token; disabled: boolean }>;
```
cmcewen: nit: since you created a separate file do you want to move all the types to it? otherwise don't need a separate file?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -582,40 +625,101 @@ async function validateForm(
...
+    'Multiple multi-collateral tokens found for same collateral address, retrieving balances...',
+  );
+  const tokenBalances: Array<{ token: Token; balance: bigint }> = [];
+
+  // fetch each destination token balance
+  const results = await Promise.allSettled(
+    tokensWithSameCollateralAddresses.map(({ destinationToken }) =>
+      warpCore.getTokenCollateral(destinationToken),
+    ),
+  );
```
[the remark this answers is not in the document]
Xaroz: failed promises will end up as status === rejected, and the arrays are not being mutated so using index is not wrong here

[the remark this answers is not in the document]
Xaroz: but will clean this up to not use index-based regardless

src/features/tokens/utils.ts
```diff
@@ -1,9 +1,12 @@
...
-};
+import {
+  IToken,
+  MultiProtocolProvider,
+  Token,
+  TOKEN_COLLATERALIZED_STANDARDS,
+  WarpCore,
+} from '@hyperlane-xyz/sdk';
+import { eqAddress, normalizeAddress } from '@hyperlane-xyz/utils';
+import { MultiCollateralTokenMap, TokenChainMap, Tokens } from './types';
```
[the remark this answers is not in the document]
Xaroz: interesting, ChainName is also declared in the `global.d.ts`, hence no warning from compiler

PR #676, opened 2025-07-17
chore: bump registry

antigremlin:
Bump registry to 19.4.0


Xaroz (approved):
I don't mind but what is the reason for this bump?

PR #680, opened 2025-07-21
feat: Add usdc/matchain

ltyu:
This PR adds usdc/matchain


src/consts/warpRouteWhitelist.ts
```diff
@@ -293,4 +293,7 @@ export const warpRouteWhitelist: Array<string> | null = [
 
   // PUMP routes
   'PUMP/eclipsemainnet',
+
+  // Matchain
```
Xaroz: suggest moving below all the USDC routes

Xaroz (approved):

PR #694, opened 2025-07-31
feat: add token button

Xaroz:
fixes [ENG-1965](https://linear.app/hyperlane-xyz/issue/ENG-1965/add-token-to-wallet-button)

Implement feature to request to add a token/asset to an user wallet. Currently only works for EVM
- New button under chains to add token
- Wallet must be connected or be a supported protocol (currently only EVM) for the button to show

image

## Summary by CodeRabbit

* **New Features**
  * Added support for users to add tokens directly from the chain selection interface when supported by the protocol and account readiness.
  * Introduced a new "Add token" button in the chain selection UI for eligible tokens.
  * Users can now add assets to their wallet for supported protocols (Ethereum) via an integrated workflow.

* **Chores**
  * Updated several dependencies to newer versions for improved stability and compatibility.

### DIFF
--- a/src/consts/args.ts
+++ b/src/consts/args.ts
@@ -6,7 +6,4 @@ export enum WARP_QUERY_PARAMS {
   TOKEN = 'token',
 }
 
-export const ADD_ASSET_SUPPORTED_PROTOCOLS: ProtocolType[] = [
-  ProtocolType.Ethereum,
-  ProtocolType.Starknet,
-];
+export const ADD_ASSET_SUPPORTED_PROTOCOLS: ProtocolType[] = [ProtocolType.Ethereum];


xeno097 (approved):

src/features/tokens/hooks.ts
```diff
@@ -98,3 +102,23 @@ export function tryFindTokenConnection(token: Token, chainName: string) {
...
+export function useAddToken() {
+  const multiProvider = useMultiProvider();
+  const activeChains = useActiveChains(multiProvider);
+  const watchAsset = useWatchAsset(multiProvider);
+  const { isPending, mutateAsync } = useMutation({
+    mutationFn: (token: IToken) => {
+      const { addAsset } = watchAsset[token.protocol];
+      const activeChain = activeChains.chains[token.protocol];
+
+      if (!ADD_ASSET_SUPPORTED_PROTOCOLS.includes(token.protocol) || !activeChain.chainName) {
```
antigremlin: Would it make sense to reject the promise or return an error in some other way?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -233,6 +242,17 @@ function ChainSelectSection({ isReview }: { isReview: boolean }) {
     setOriginChainName(origin);
   };
 
+  const onAddToken = useCallback(
+    async (token: IToken) => {
+      try {
+        await addToken(token);
+      } catch {
+        logger.debug('Canceled add asset');
```
antigremlin: Could/should we display an error message in the UI?

src/features/chains/ChainSelectField.tsx
```diff
@@ -28,14 +55,14 @@ export function ChainSelectField({ name, label, onChange, disabled, customListIt
     if (onChange) onChange(chainName, name);
   };
 
-  const [isModalOpen, setIsModalOpen] = useState(false);
-
   const onClick = () => {
     if (!disabled) setIsModalOpen(true);
   };
 
+  const showAddToken = token && isAccountReady && isSupportedProtocol;
```
antigremlin: Here's an idea: As we are introducing a hook that allows adding a token to the users wallet, why not encapsulate the logic of when this is allowed in the same hook?

What I mean is that the hook could return `[ addAsset, canAddAsset ]` and instead of having all the logic around `isAccountReady` and `isSupportedProtocol` we could just call the hook.

antigremlin (commented):
Here are some comments + an idea on encapsulating the logic better

src/features/transfer/TransferTokenForm.tsx
Xaroz: Now shows a toast for error outside of User rejecting the import

src/features/chains/ChainSelectField.tsx
Xaroz: I moved the logic inside `useAddToken` hook, one thing about this is that we would have two instance of the hook because moving it inside `ChainSelectField` is easier to handle.

antigremlin (approved):
Nice!

PR #695, opened 2025-08-01
fix: remove auto switch to Base network when connecting wallet

CJ42:
This PR addresses an issue where users are automatically prompted to switch to the Base Mainnet when connecting with their wallet on the Hyperlane Warp bridge UI.

## 📋 Context

When attempting to connect a wallet, users were forced to switch to the Base network regardless of the chain selected, or the previous chain selected in their wallet. 

This is due to the fact that the code of the Hyperlane UI fetches the list of available warp routes, and pass the first origin chain of the first warp route as initial chain to rainbow kit configuration in the wallet connector.

Although Metamask performs the chain switch in the background in the wallet, this still causes confusion for users expecting to use the origin chain from the selected dropdown. For instance when using the Universal Profile browser extension, the user gets automatically prompted to switch network, which leads to confusion.

image

https://github.com/user-attachments/assets/114d703d-1cd3-4316-bab8-b51027970a0a

## ✅ Fix

Removed the logic that enforced a network switch to Base on wallet connection. The code now does not provide an initial chain to switch to when connecting. The last network selected in the wallet is now used. Wallets now connect without triggering an automatic network change

## Summary by CodeRabbit

* **Refactor**
  * Removed logic related to deriving and passing an initial chain based on Ethereum protocol tokens in wallet context setup. No visible changes to the user experience.

### DIFF
--- a/src/features/wallet/context/EvmWalletContext.tsx
+++ b/src/features/wallet/context/EvmWalletContext.tsx
@@ -1,5 +1,4 @@
 import { MultiProtocolProvider } from '@hyperlane-xyz/sdk';
-import { ProtocolType } from '@hyperlane-xyz/utils';
 import { getWagmiChainConfigs } from '@hyperlane-xyz/widgets';
 import { RainbowKitProvider, connectorsForWallets, lightTheme } from '@rainbow-me/rainbowkit';
 import '@rainbow-me/rainbowkit/styles.css';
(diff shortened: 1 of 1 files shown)


Xaroz:
Thanks for the fix! Not exactly sure about this but I think if the current network the user is connected to is not in the `wagmiConfig` chain list, it will still enforce a network switch but otherwise this fixes the mentioned issue

Xaroz (approved):

PR #714, opened 2025-08-08
feat: add warnings for smart wallets if bridging to same address on destination chain that is not a smart contract

CJ42:
# Context

When users bridge tokens between chains (e.g. from Base to Ethereum) using a smart contract wallet, they might enter their own wallet address as the recipient on the destination chain.

But if that smart wallet is only deployed on the source chain (e.g. Base ✅) and not on the destination chain (e.g. Ethereum ❌), there's no contract on the destination side to control the funds.

This can result in the user loosing its bridged tokens, as they cannot be accessed on the destination chain in this scenario, **since the smart contract wallet is not deployed on the destination chain**.

# ☑️ How does this PR mitigates this issue?

This PR prevents accidental loss by checking the bytecode at the recipient address on the destination chain, if a smart contract wallet is connected to the UI and the same recipient address is used.

# 🔄 Flow

1. User connects a smart contract wallet to the dApp.
2. They select the source/destination chain, token, and enter an amount to bridge.
2. They enter their own wallet address as the recipient.

Then the app checks:

- ✅ Is the connected wallet a smart contract? _Yes_
- ✅ Is the recipient address the same? _Yes_
- ❌ Is there any bytecode at that address on the destination chain? Is it a smart contract at that address on the destination chain? _No_

> If so, the "Continue" button is disabled and a warning message is shown.
> The user can tick a checkbox to proceed only if they are sure they can control the recipient address on the destination chain.

image

# Demo

https://github.com/user-attachments/assets/562da9bb-64e9-4601-b48b-5784811fc70b

## Summary by CodeRabbit

* **New Features**
  * Recipient warning banner for Ethereum→Ethereum transfers with explanatory text and a confirmation checkbox that gates submission.
  * Submit button disabled until the user confirms control of the recipient address.
  * On-chain contract checks and friendly chain names with clear error feedback; wallet disconnect hides the banner and resets confirmation.

* **Bug Fixes**
  * Reduces accidental transfers by blocking submission until recipient ownership is confirmed.

### DIFF
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -63,6 +63,8 @@ export interface AppState {
     options?: { msgId?: string; originTxHash?: string },
   ) => void;
   failUnconfirmedTransfers: () => void;
+  recipientAddressConfirmed: boolean;
+  setRecipientAddressConfirmed: (confirmed: boolean) => void;
 
   // Shared component state
   transferLoading: boolean;
@@ -173,6 +175,10 @@ export const useStore = create<AppState>()(
       showEnvSelectModal: false,
       setShowEnvSelectModal: (showEnvSelectModal) => {
         set(() => ({ showEnvSelectModal }));
+      },
+      recipientAddressConfirmed: true,
+      setRecipientAddressConfirmed: (recipientAddressConfirmed) => {
+        set(() => ({ recipientAddressConfirmed }));
       },
       originChainName: '',
       setOriginChainName: (originChainName: ChainName) => {
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -63,8 +63,6 @@ export interface AppState {
     options?: { msgId?: string; originTxHash?: string },
   ) => void;
   failUnconfirmedTransfers: () => void;
-  recipientAddressConfirmed: boolean;
-  setRecipientAddressConfirmed: (confirmed: boolean) => void;
 
   // Shared component state
   transferLoading: boolean;
@@ -175,10 +173,6 @@ export const useStore = create<AppState>()(
       showEnvSelectModal: false,
       setShowEnvSelectModal: (showEnvSelectModal) => {
         set(() => ({ showEnvSelectModal }));
-      },
-      recipientAddressConfirmed: true,
-      setRecipientAddressConfirmed: (recipientAddressConfirmed) => {
-        set(() => ({ recipientAddressConfirmed }));
       },
       originChainName: '',
       setOriginChainName: (originChainName: ChainName) => {
--- a/src/features/transfer/TransferTokenForm.tsx
+++ b/src/features/transfer/TransferTokenForm.tsx
@@ -396,11 +396,12 @@ function ButtonSection({
         sourceProtocol !== ProtocolType.Ethereum ||
         destinationProtocol !== ProtocolType.Ethereum
       ) {
+        setRecipientInfos({ showWarning: false, addressConfirmed: true });
         return;
       }
 
       if (!isValidAddressEvm(recipient)) {
-        setRecipientInfos({ showWarning: false, addressConfirmed: false });
+        setRecipientInfos({ showWarning: false, addressConfirmed: true });
         return;
       }
 
--- a/src/features/transfer/TransferTokenForm.tsx
+++ b/src/features/transfer/TransferTokenForm.tsx
@@ -2,6 +2,7 @@ import { IToken, Token, TokenAmount, WarpCore } from '@hyperlane-xyz/sdk';
 import {
   ProtocolType,
   convertToScaledAmount,
+  eqAddress,
   errorToString,
   fromWei,
   isNullish,
@@ -67,8 +68,6 @@ import { useTokenTransfer } from './useTokenTransfer';
 import { isSmartContract } from './utils';
 
 export function TransferTokenForm() {
-  const store = useStore();
-
   const multiProvider = useMultiProvider();
   const warpCore = useWarpCore();
 
@@ -124,7 +123,7 @@ export function TransferTokenForm() {
 
   useEffect(() => {
     if (!originChainName) setOriginChainName(initialValues.origin);
-  }, [initialValues.origin, originChainName, setOriginChainName, store.recipientAddressConfirmed]);
+  }, [initialValues.origin, originChainName, setOriginChainName]);
 
   return (
     <Formik<TransferFormValues>
@@ -145,7 +144,6 @@ export function TransferTokenForm() {
           <RecipientSection isReview={isReview} />
           <ReviewDetails visible={isReview} routeOverrideToken={routeOverrideToken} />
           <ButtonSection
-            disabled={!store.recipientAddressConfirmed}
             isReview={isReview}
             isValidating={isValidating}
             setIsReview={setIsReview}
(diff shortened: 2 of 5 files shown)


src/features/store.ts
```diff
@@ -174,6 +176,10 @@ export const useStore = create<AppState>()(
       setShowEnvSelectModal: (showEnvSelectModal) => {
         set(() => ({ showEnvSelectModal }));
       },
+      recipientAddressConfirmed: true,
+      setRecipientAddressConfirmed: (recipientAddressConfirmed) => {
+        set(() => ({ recipientAddressConfirmed }));
+      },
```
Xaroz: I'd prefer if this is not a store and rather just a state, no need to add this to the app store

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -120,7 +124,7 @@ export function TransferTokenForm() {
 
   useEffect(() => {
     if (!originChainName) setOriginChainName(initialValues.origin);
-  }, [initialValues.origin, originChainName, setOriginChainName]);
+  }, [initialValues.origin, originChainName, setOriginChainName, store.recipientAddressConfirmed]);
```
Xaroz: can you explain why you need this dependency here?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -314,10 +319,67 @@ function AmountSection({ isNft, isReview }: { isNft: boolean; isReview: boolean
...
+      setDestinationChainName(destinationChainName || '');
+
+      // check first if the address on origin is a smart contract
+      const isSenderSmartContract = await isSmartContract(
+        multiProvider,
+        values.origin,
+        connectedWallet,
+      );
+
+      const isSelfRecipient = values.recipient?.toLowerCase() === connectedWallet.toLowerCase();
```
Xaroz: - Could you use our utils function called: `eqAddress` for this?
- Is this only for EVM contracts? If so preferred we do an early return that checks for protocol types

src/features/transfer/utils.ts
```diff
@@ -129,3 +132,30 @@ export function tryGetMsgIdFromTransferReceipt(
...
+    if (!code || code === '0x') {
+      return false;
+    }
+
+    return true;
+  } catch (error) {
+    const msg = `Error checking if address is a smart contract on ${getChainDisplayName(multiProvider, origin)}`;
+    logger.error(msg, error);
+    toast.error(msg);
+    return undefined;
```
Xaroz: your catch cover all the cases where it throws because you return undefined, this mean other VMs or other errors could trigger the banner

src/features/transfer/TransferTokenForm.tsx
CJ42: Refactored and removed it. Not needed anymore

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -360,8 +363,82 @@ function ButtonSection({
...
+      }
+
+      if (recipientCheckError) {
+        toast.error(recipientCheckError);
+      }
+
+      if (senderCheckError || recipientCheckError) {
+        setRecipientInfos({ addressConfirmed: true, showWarning: false });
+        return;
+      }
```
Xaroz:
```suggestion
      if (senderCheckError || recipientCheckError) {
        toast.error(senderCheckError || recipientCheckError);
        setRecipientInfos({ addressConfirmed: true, showWarning: false });
        return;
      }
```

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -360,8 +363,82 @@ function ButtonSection({
...
+
+      const { protocol: destinationProtocol } = multiProvider.getChainMetadata(values.destination);
+      const { protocol: sourceProtocol } = multiProvider.getChainMetadata(values.origin);
+
+      // Check if we are only dealing with bridging between two EVM chains
+      if (sourceProtocol != ProtocolType.Ethereum || destinationProtocol != ProtocolType.Ethereum) {
+        return;
+      }
```
Xaroz:
```suggestion
      if (sourceProtocol !== ProtocolType.Ethereum || destinationProtocol !== ProtocolType.Ethereum) {
        return;
      }
```

CJ42:
@Xaroz I have addressed all your review comments, including the suggestions from **coderabbitai**.
Let me know if it is good for you. I have also rebased on the latest `main` branch.

Xaroz:
@CJ42 hey just checked the preview deploy, whenever I connect a normal wallet, and try to change any chain or type in the recipient the following toast shows which is not something we want:
image

CJ42:
@Xaroz this should be fixed now

Xaroz:
@CJ42 the issue is still happening if you type anything that is not an address, I understand that people would usually use the self button or paste their addresses but we should not be defaulting to showing an error toast if somebody types anything

CJ42:
> @CJ42 the issue is still happening if you type anything that is not an address, I understand that people would usually use the self button or paste their addresses but we should not be defaulting to showing an error toast if somebody types anything

Can you try again? Hopefully this should be fixed now.

> 
If it is easier for you, you can always try adding the warning after on submit similar to the RecipientConfirmationModal

The problem is that I needed to put it inside the `<ButtonSection />` component as I need 1) the Formik form values, and 2) to disable the button if needed.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -360,8 +364,83 @@ function ButtonSection({
...
+        sourceProtocol !== ProtocolType.Ethereum ||
+        destinationProtocol !== ProtocolType.Ethereum
+      ) {
+        return;
+      }
+
+      if (!isAddress(recipient)) {
+        setRecipientInfos({ showWarning: false, addressConfirmed: false });
+        return;
+      }
```
Xaroz: could you change this to us `isValidAddressEvm` from our utils package?

b00ste: yes, on it

src/features/transfer/utils.ts
```diff
@@ -129,3 +131,33 @@ export function tryGetMsgIdFromTransferReceipt(
...
 }
+
+export async function isSmartContract(
+  multiProvider: MultiProtocolProvider,
+  chain: string,
+  address: string,
+): Promise<{ isContract: boolean; error?: string }> {
+  if (!isAddress(address)) {
+    return { isContract: false };
+  }
```
Xaroz: ditto

src/features/transfer/TransferTokenForm.tsx
b00ste: Fixed 🫡

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -360,8 +364,83 @@ function ButtonSection({
...
+      const { protocol: sourceProtocol } = multiProvider.getChainMetadata(values.origin);
+
+      // Check if we are only dealing with bridging between two EVM chains
+      if (
+        sourceProtocol !== ProtocolType.Ethereum ||
+        destinationProtocol !== ProtocolType.Ethereum
+      ) {
+        return;
+      }
+
```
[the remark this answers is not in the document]
Xaroz: @b00ste is this an issue?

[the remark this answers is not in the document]
b00ste: Just tested and it is indeed a issue when changing chain after initially inputting an address (edge case), pushed a change to reset state when protocol is not EVM

CJ42:
@Xaroz we have addressed all the review comments.
Can this be merged? 🙏🏻

PR #729, opened 2025-08-28
chore: update tip card for new domain

Xaroz:
Update TipCard with information about the upcoming domain change
image

### DIFF
--- a/src/components/tip/TipCard.tsx
+++ b/src/components/tip/TipCard.tsx
@@ -10,20 +10,28 @@ export function TipCard() {
   if (!show) return null;
   return (
     <Card className="w-100 p-2 sm:w-[31rem]">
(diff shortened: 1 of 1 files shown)


src/components/tip/TipCard.tsx
```diff
@@ -10,20 +10,28 @@ export function TipCard() {
...
+            href="http://nexus.hyperlane.xyz/"
+            target="_blank"
+            rel="noopener noreferrer"
+            className="text-primary-500"
+          >
+            nexus.hyperlane.xyz
+          </a>
+          . Nothing you need to do except access us at the new domain.
         </p>
         <a
```
paulbalaji, line 27: is this the copy from marketing? or should we ask them for what to say here

Xaroz: This is the copy from marketing indeed

paulbalaji (approved):

PR #754, opened 2025-09-18
feat: multi-collateral warp routes maxTransfer

Xaroz:
This PR allows the maxTransfer button to check for route with lowest fee before setting the amount

- Gets route with lowest fee for max transfer button
- Fixed a bug where `getLowestFeeTransferToken` would use plain `amount` instead of `amountWei`

### DIFF
--- a/src/features/transfer/fees.ts
+++ b/src/features/transfer/fees.ts
@@ -61,7 +61,7 @@ export async function getLowestFeeTransferToken(
   warpCore: WarpCore,
   originToken: Token,
   destinationToken: IToken,
-  amount: string,
+  amounWei: string,
   recipient: string,
   sender: string | undefined,
 ) {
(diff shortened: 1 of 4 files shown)


src/features/transfer/fees.ts
```diff
@@ -85,12 +85,12 @@ export async function getLowestFeeTransferToken(
...
     tokensWithSameCollateralAddresses.map(async ({ originToken, destinationToken }) => {
       try {
-        const originTokenAmount = new TokenAmount(amount, originToken);
+        const originTokenAmount = new TokenAmount(amounWei, originToken);
         const fees = await warpCore.getInterchainTransferFee({
           originTokenAmount,
           destination: destinationToken.chainName,
           recipient,
-          sender,
+          sender: sender,
```
xeno097: why `sender: sender` you can omit it no?

xeno097 (approved):

src/features/transfer/fees.ts
Xaroz: whoops, forgot this since I was testing with `warpCore.estimateTransferRemote` that requires the sender to be defined

PR #788, opened 2025-10-27
feat: vercel analytics 

Xaroz:
This PR implements custom event tracking using vercel analytics as a first approach to track user interactions in the website. 

- Include a flag `enableTrackingEvents` in the configurations which will enable/disable custom even trackings depending on the branch
- Add `EVENT_NAME` and `EventProperties` which includes the custom event name and the fields that will be added
- Include functions to track custom events
- Current events being tracked
  - Chain Selection
  - Token Selection
  - Transaction Submission
  - Transaction Submission Failed
  - Wallet Connection Initiated
  - Wallet Connected

## Summary by CodeRabbit

* **Chores**
  * Added analytics for wallet connection initiation and connected wallets (deduplicated).
  * Added analytics for chain selection, token selection, transaction submission, and transaction failures.
  * Transaction submission events include transaction hash, chain info, amount and wallet details.
  * New config flag enableTrackingEvents (default: false) to toggle analytics.

* **Bug Fixes**
  * Submit button click now consistently triggers connection analytics when a wallet connection is initiated.

### DIFF
--- a/src/features/analytics/utils.ts
+++ b/src/features/analytics/utils.ts
@@ -1,5 +1,5 @@
 import { IToken, MultiProtocolProvider, Token, WarpCore } from '@hyperlane-xyz/sdk';
-import { ProtocolType } from '@hyperlane-xyz/utils';
+import { objLength, ProtocolType } from '@hyperlane-xyz/utils';
 import { AccountInfo, getAccountAddressAndPubKey } from '@hyperlane-xyz/widgets';
 import { track } from '@vercel/analytics';
 import { config } from '../../consts/config';
@@ -45,7 +45,7 @@ export function trackTransactionFailedEvent(
   accounts: Record<ProtocolType, AccountInfo>,
   overrideToken: Token | null,
 ) {
-  if (!errors || Object.keys(errors).length < 1) return;
+  if (!errors || objLength(errors) < 1) return;
 
   const firstError = `${Object.values(errors)[0]}` || 'Unknown error';
 


src/features/analytics/types.ts
```diff
@@ -0,0 +1,61 @@
...
+import { ProtocolType } from '@hyperlane-xyz/utils';
+
+export enum EVENT_NAME {
+  CHAIN_SELECTION = 'Chain Selection',
+  TOKEN_SELECTION = 'Token Selection',
+  TRANSACTION_SUBMISSION = 'Transaction Submission',
+  TRANSACTION_SUBMISSION_FAILED = 'Transaction Submission Failed',
+  WALLET_CONNECTION_INITIATED = 'Wallet Connection Initiated',
+  WALLET_CONNECTED = 'Wallet Connected',
+}
```
xeno097, line 12: Based on the doc you shared aren't we missing some events here? like the msg dispatched and processed?

src/components/layout/AppLayout.tsx
```diff
@@ -18,6 +21,8 @@ export function AppLayout({ children }: PropsWithChildren) {
     }),
   );
 
+  useWalletConnectionTracking();
```
xeno097, line 24: curiosity question: what is the difference between doing this function call and writing the same code but instead of activating tracking as function call it is registered as provider like the wallet connectors are?

just to be clearer couldn't we do something like this but for the analytics?

Image

src/features/analytics/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+import { TransferFormValues } from '../transfer/types';
+import { EVENT_NAME, EventProperties } from './types';
+
+export function trackEvent<T extends EVENT_NAME>(eventName: T, properties: EventProperties[T]) {
+  if (!config.enableTrackingEvents) return;
+
+  // take into consideration vercel only allows up to 8 properties
+  track(eventName, {
+    ...properties,
+  });
```
xeno097, line 16: maybe it would be a good idea to check that the object has at most 8 properties and throw an error if it doesn't ?

src/features/analytics/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+    tokenSymbol: token.symbol,
+    origin,
+    destination,
+    originChainId,
+    destinationChainId,
+  });
+}
+
+// errors that happen because of form not being filled correctly
+const SKIPPED_ERRORS = ['Token is required', 'Invalid amount'];
```
xeno097, line 39: Is there a way we can avoid magic strings, so that if the error msgs change were they are thrown we don't end up out of sync here?

src/features/analytics/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export function trackTransactionFailedEvent(
+  errors: Record<string, string> | null,
+  warpCore: WarpCore,
+  values: TransferFormValues,
+  accounts: Record<ProtocolType, AccountInfo>,
+  overrideToken: Token | null,
+) {
+  if (!errors || Object.keys(errors).length < 1) return;
```
xeno097: why not use the `objLength` function from the utils package?

src/features/analytics/types.ts
Xaroz: Those won't be tracked on the UI

src/components/layout/AppLayout.tsx
Xaroz: It would be more boilerplate and if we add another context, there could be moments where a re-rendering could cause the entire tree to re-render, just trying to avoid adding more context in our app which already has so many

src/features/analytics/utils.ts
Xaroz: It could be done with `Token is required`, but `Invalid amount` comes from the SDK so we would need to export an error enum from there which seems unnecessary

src/features/analytics/utils.ts
Xaroz: Rather not throw an error for this, tested with more than 8 properties in the events and seem that it still goes through without issue, will keep monitoring on this

PR #793, opened 2025-10-30
feat: token price in USD

Xaroz:
fixes [ENG-2200](https://linear.app/hyperlane-xyz/issue/ENG-2200/warp-template-ui-should-show-fees-also-in-usd-terms)

- Fetches token prices in USD from the CoinGecko API
- Shows token price multiplied by the amount
- Attach `coinGeckoId` when warpCore is when built to each token that has no `coinGeckoId` (synthetic tokens mostly)

image

## Summary by CodeRabbit

* **New Features**
  * Token transfers now show approximate USD prices below the amount input (en-US, two decimals). Prices appear only when the converted total is at least $0.01.
  * Pricing updates automatically every 60 seconds and loads without refetching on window focus.

* **Improvements**
  * Price availability is more consistent across routes by filling missing price identifiers from related tokens when possible.

### DIFF
--- a/src/features/tokens/useTokenPrice.tsx
+++ b/src/features/tokens/useTokenPrice.tsx
@@ -29,13 +29,23 @@ type CoinGeckoResponse = Record<string, { usd: number }>;
 
 async function fetchTokenPrice(originToken?: Token): Promise<number | null> {
   if (!originToken || !originToken.coinGeckoId) return null;
-  logger.debug('Fetching token price');
 
-  const res = await fetch(
-    `https://api.coingecko.com/api/v3/simple/price?ids=${originToken.coinGeckoId}&vs_currencies=usd`,
-  );
-  const data: CoinGeckoResponse = await res.json();
+  try {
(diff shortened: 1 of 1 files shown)


src/features/tokens/useTokenPrice.tsx
```diff
@@ -0,0 +1,41 @@
...
+  logger.debug('Fetching token price');
+
+  const res = await fetch(
+    `https://api.coingecko.com/api/v3/simple/price?ids=${originToken.coinGeckoId}&vs_currencies=usd`,
+  );
+  const data: CoinGeckoResponse = await res.json();
+
+  const priceData = Object.values(data)[0];
+  return priceData?.usd ?? null;
+}
```
paulbalaji, line 51: do we not need an api key?

Xaroz: From what I understand, if we use an API key then the rate limit is applied to the key itself, it we don't the rate limit would be applied to the user, so I preferred this approach.

PR #837, opened 2025-12-05
chore: change token identification to use an identifier instead of tokenIndex

Xaroz:
This PR changes the tokens identification system to use an stable ID instead of `tokenIndex`, this ID is made using the token's `chain`, `symbol` and `addressOrDenom`. For example:
`ethereum-usdc-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48`

fixes [ENG-2764](https://linear.app/hyperlane-xyz/issue/ENG-2764/replace-tokenindex-with-stable-token-identifiers)

Coded using help of Claude

### DIFF
--- a/src/features/tokens/TokenSelectField.tsx
+++ b/src/features/tokens/TokenSelectField.tsx
@@ -10,7 +10,8 @@ import { trackTokenSelectionEvent } from '../analytics/utils';
 import { useMultiProvider } from '../chains/hooks';
 import { TransferFormValues } from '../transfer/types';
 import { TokenListModal } from './TokenListModal';
-import { getTokenByKey, getTokenKey, getTokenKeyFromChains, useWarpCore } from './hooks';
+import { getTokenByKey, getTokenKeyFromChains, useWarpCore } from './hooks';
+import { getTokenKey } from './utils';
 
 type Props = {
   name: string;
--- a/src/features/tokens/hooks.ts
+++ b/src/features/tokens/hooks.ts
@@ -1,30 +1,14 @@
 import { IToken, Token, WarpCore } from '@hyperlane-xyz/sdk';
-import { normalizeAddress } from '@hyperlane-xyz/utils';
 import { useAccountForChain, useActiveChains, useWatchAsset } from '@hyperlane-xyz/widgets';
 import { useMutation } from '@tanstack/react-query';
 import { ADD_ASSET_SUPPORTED_PROTOCOLS } from '../../consts/args';
 import { useMultiProvider } from '../chains/hooks';
 import { useStore } from '../store';
+import { getTokenKey } from './utils';
 
 export function useWarpCore() {
   return useStore((s) => s.warpCore);
 }
-
-// ============================================
-// Token Key System
-// ============================================
-// Format: "chainName-symbol-addressOrDenom" (stable identifier)
-// Example: "ethereum-usdc-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"
-
-/**
- * Generate a stable token key from a token object
- * Uses chainName + lowercase symbol + normalized address
- */
-export function getTokenKey(token: IToken): string {
-  const normalizedAddress = normalizeAddress(token.addressOrDenom, token.protocol);
-  return `${token.chainName.toLowerCase()}-${token.symbol.toLowerCase()}-${normalizedAddress}`;
-}
-
 /**
  * Find a token by its key
  */
@@ -39,14 +23,6 @@ export function getTokenByKey(warpCore: WarpCore, key: string | undefined): Toke
 export function useTokenByKey(key: string | undefined): Token | undefined {
   const warpCore = useWarpCore();
   return getTokenByKey(warpCore, key);
-}
-
-/**
- * Hook to get a token key for a token
- */
-export function useTokenKey(token: IToken | undefined): string | undefined {
-  if (!token) return undefined;
-  return getTokenKey(token);
 }
 
 /**
--- a/src/features/transfer/TransferTokenForm.tsx
+++ b/src/features/transfer/TransferTokenForm.tsx
@@ -52,13 +52,13 @@ import {
   useOriginBalance,
 } from '../tokens/balances';
 import {
-  getTokenByKey,
-  getTokenKey,
   getInitialTokenKey,
+  getTokenByKey,
   getTokenKeyFromChains,
   useWarpCore,
 } from '../tokens/hooks';
 import { useTokenPrice } from '../tokens/useTokenPrice';
+import { getTokenKey } from '../tokens/utils';
 import { WalletConnectionWarning } from '../wallet/WalletConnectionWarning';
 import { FeeSectionButton } from './FeeSectionButton';
 import { RecipientConfirmationModal } from './RecipientConfirmationModal';
--- a/src/features/tokens/utils.ts
+++ b/src/features/tokens/utils.ts
@@ -128,6 +128,17 @@ export function getTokensWithSameCollateralAddresses(
     });
 }
 
+/**
+ * Generate a stable token key from a token object
+ * Uses chainName + lowercase symbol + normalized address
+ * Format: "chainName-symbol-addressOrDenom" (stable identifier)
+ * Example: "ethereum-usdc-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"
+ */
+export function getTokenKey(token: IToken): string {
+  const normalizedAddress = normalizeAddress(token.addressOrDenom, token.protocol);
+  return `${token.chainName.toLowerCase()}-${token.symbol.toLowerCase()}-${normalizedAddress}`;
+}
+
 // De-duplicate collaterized tokens
 // Returns a map of token with same origin and dest collateral address
 // And an array of tokens with repeated collateral addresses grouped into one
@@ -178,10 +189,13 @@ export function dedupeTokensByCollateral(tokens: Token[]): Token[] {
       return true;
     }
 
-    // For HypNative tokens, use their symbol and standard as identifier
+    const chainName = token.chainName.toLowerCase();
+    const symbol = token.symbol.toLowerCase();
+    const protocol = token.protocol;
+    // For HypNative tokens, use their symbol and protocol as identifier
     const collateralKey = token.collateralAddressOrDenom
-      ? `${token.chainName}-${token.symbol}-${normalizeAddress(token.collateralAddressOrDenom, token.protocol)}`
-      : `${token.chainName}-${token.symbol}-hypnative-${token.standard}`;
+      ? `${chainName}-${symbol}-${normalizeAddress(token.collateralAddressOrDenom, protocol)}`
+      : `${chainName}-${symbol}-hypnative-${protocol}`;
 
     // If we haven't seen this collateral on this chain, include it
     if (!seenCollaterals.has(collateralKey)) {
@@ -205,7 +219,7 @@ export function buildOriginTokens(tokens: Token[]): Token[] {
   // Deduplicate by chain-address
   const tokenMap = new Map<string, Token>();
   originTokens.forEach((token) => {
-    const key = `${token.chainName}-${token.symbol}-${token.addressOrDenom}`;
+    const key = getTokenKey(token);
     if (!tokenMap.has(key)) {
       tokenMap.set(key, token);
     }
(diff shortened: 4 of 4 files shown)


src/features/tokens/hooks.ts
```diff
@@ -10,6 +10,143 @@ export function useWarpCore() {
...
+  return `${token.chainName}-${token.symbol.toLowerCase()}-${normalizedAddress}`;
+}
+
+/**
+ * Find a token by its key
+ */
+export function getTokenByKey(warpCore: WarpCore, key: string | undefined): Token | undefined {
+  if (!key) return undefined;
+  return warpCore.tokens.find((token) => getTokenKey(token) === key);
+}
```
xeno097, line 18: Why not create a mapping at app start where the key is the token key an the value is the token so you don't have to iterate each time to filter?

xeno097 (commented):
Only complaint i have is that each time we need to get something we need to iterate the full list of tokens in the worst case, having a mapping might help in mitigating this

src/features/tokens/hooks.ts
Xaroz: I have thought about it but the reason I didn't do that it's because what we are gonna use to integrate this new swap architecture will require us to fetch tokens dynamically so creating a static map might be a bit pointless

src/features/tokens/hooks.ts
```diff
@@ -10,6 +10,143 @@ export function useWarpCore() {
...
+// Format: "chainName-symbol-addressOrDenom" (stable identifier)
+// Example: "ethereum-usdc-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"
+
+/**
+ * Generate a stable token key from a token object
+ * Uses chainName + lowercase symbol + normalized address
+ */
+export function getTokenKey(token: IToken): string {
+  const normalizedAddress = normalizeAddress(token.addressOrDenom, token.protocol);
+  return `${token.chainName}-${token.symbol.toLowerCase()}-${normalizedAddress}`;
```
antigremlin: Why not normalize (lowercase) the chain name as well, to make it even more stable?

src/features/tokens/hooks.ts
```diff
@@ -10,6 +10,143 @@ export function useWarpCore() {
...
+// Token Key System
+// ============================================
+// Format: "chainName-symbol-addressOrDenom" (stable identifier)
+// Example: "ethereum-usdc-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"
+
+/**
+ * Generate a stable token key from a token object
+ * Uses chainName + lowercase symbol + normalized address
+ */
+export function getTokenKey(token: IToken): string {
```
antigremlin: Did you consider to make this a token method? It doesn't seem a bad idea to make this ID available to the rest of the code.

Can be changed in a later SDK PR once this work has stabilised.

src/features/tokens/hooks.ts
```diff
@@ -10,6 +10,143 @@ export function useWarpCore() {
...
+ */
+export function useTokenByKey(key: string | undefined): Token | undefined {
+  const warpCore = useWarpCore();
+  return getTokenByKey(warpCore, key);
+}
+
+/**
+ * Hook to get a token key for a token
+ */
+export function useTokenKey(token: IToken | undefined): string | undefined {
```
antigremlin: Seems to go unused for now. And it's not a hook as is doesn't depend on other hooks. Remove?

src/features/tokens/hooks.ts
```diff
@@ -10,6 +10,143 @@ export function useWarpCore() {
...
+// Example: "ethereum-usdc-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"
+
+/**
+ * Generate a stable token key from a token object
+ * Uses chainName + lowercase symbol + normalized address
+ */
+export function getTokenKey(token: IToken): string {
+  const normalizedAddress = normalizeAddress(token.addressOrDenom, token.protocol);
+  return `${token.chainName}-${token.symbol.toLowerCase()}-${normalizedAddress}`;
+}
```
antigremlin: What if it returns `chain-null-undefined`? Should we validate somewhere?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -218,17 +217,18 @@ function ChainSelectSection({ isReview }: { isReview: boolean }) {
...
     const destinationToken = originToken.getConnectionForChain(values.destination)?.token;
     return { originToken, destinationToken };
-  }, [values.tokenIndex, values.destination, warpCore]);
+  }, [values.tokenKey, values.destination, warpCore]);
 
   const setTokenOnChainChange = (origin: string, destination: string) => {
-    const tokenIndex = getTokenIndexFromChains(warpCore, null, origin, destination);
-    const token = getTokenByIndex(warpCore, tokenIndex);
+    // Returns token key if only one route exists, undefined if 0 or multiple routes
+    const tokenKey = getTokenKeyFromChains(warpCore, null, origin, destination);
```
antigremlin, line 228: Why don't we try to preserve the same token symbol (or even key when we change the destination)?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -218,17 +217,18 @@ function ChainSelectSection({ isReview }: { isReview: boolean }) {
...
+    const token = tokenKey ? getTokenByKey(warpCore, tokenKey) : undefined;
     updateQueryParam(WARP_QUERY_PARAMS.TOKEN, token?.symbol);
```
antigremlin, line 230: With the new token keys, is it still OK to have the symbol as part of the query params? If a user was to share this URL, would it open the app in the same state?

src/features/transfer/types.ts
```diff
@@ -1,7 +1,7 @@
 export interface TransferFormValues {
   origin: ChainName;
   destination: ChainName;
-  tokenIndex: number | undefined;
+  tokenKey: string | undefined;
```
antigremlin, line 4: Looking at the usage, this should be renamed to `originTokenKey`

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -841,8 +839,8 @@ async function validateForm(
   // returns a tuple, where first value is validation result
   // and second value is token override
   try {
-    const { origin, destination, tokenIndex, amount, recipient } = values;
-    const token = getTokenByIndex(warpCore, tokenIndex);
+    const { origin, destination, tokenKey, amount, recipient } = values;
+    const token = getTokenByKey(warpCore, tokenKey);
```
antigremlin, line 843: Rename to `originToken` for clarity?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -822,11 +820,11 @@ function useFormInitialValues(): TransferFormValues {
       destination: chainsValid
         ? destinationQuery
         : config.defaultDestinationChain || connectedToken?.token?.chainName || '',
-      tokenIndex: tokenIndex,
+      tokenKey: tokenKey,
       amount: '',
       recipient: '',
     };
-  }, [warpCore, destinationQuery, originQuery, tokenIndex, defaultOriginToken]);
+  }, [warpCore, destinationQuery, originQuery, tokenKey, defaultOriginToken]);
```
antigremlin, line 827: What's the point of useMemo() when it doesn't do anything expensive? Maybe transform into a regular function? Or if we want to avoid calling `tryGetValidChainName` repeatedly, transform into useMemo with dependencies `[warpCore, ...various query params]`.

Could be another PR ofc

src/features/tokens/TokenSelectField.tsx
```diff
@@ -20,7 +20,7 @@ type Props = {
 
 export function TokenSelectField({ name, disabled, setIsNft }: Props) {
   const { values, setValues } = useFormikContext<TransferFormValues>();
-  const [field, , helpers] = useField<number | undefined>(name);
+  const [field, , helpers] = useField<string | undefined>(name);
```
antigremlin: Renaming to `tokenField` would make this component easier to read.

src/features/tokens/SelectTokenIdField.tsx
```diff
@@ -4,7 +4,7 @@ import { useState } from 'react';
 
 type Props = {
   name: string;
-  tokenIndex?: number;
+  tokenKey?: string;
```
antigremlin: This prop seems unused?

src/features/tokens/hooks.ts
antigremlin: If this is for one warpCore instance (one warp route only) then it's fine to do linear search. How many tokens are usually in a single route? Less than a dozen?

src/features/tokens/SelectOrInputTokenIds.tsx
```diff
@@ -17,13 +17,13 @@ export function SelectOrInputTokenIds({ disabled }: { disabled: boolean }) {
   const isContractAllowToGetTokenIds = true;
 
   return isContractAllowToGetTokenIds ? (
-    <SelectTokenIdField name="amount" disabled={disabled} tokenIndex={tokenIndex} />
+    <SelectTokenIdField name="amount" disabled={disabled} tokenKey={tokenKey} />
   ) : (
-    <InputTokenId disabled={disabled} tokenIndex={tokenIndex} />
+    <InputTokenId disabled={disabled} tokenKey={tokenKey} />
```
antigremlin: `tokenKey` also seems unused, why?

src/features/tokens/hooks.ts
Xaroz: I'd say this is for all the tokens inside the WarpCore instance, not only one warp route which is usually around 200~300 which is still fine

src/features/tokens/hooks.ts
Xaroz: looking at the utils function normalizeAddress should never return undefined, if it fails to normalize it just returns the same address

src/features/tokens/hooks.ts
Xaroz: good call, although `chainName` field is usually already lower case

src/features/tokens/hooks.ts
Xaroz: Ill add this as a task in linear as a reminder

src/features/transfer/TransferTokenForm.tsx
Xaroz: yep! It should still be fine however I have further refined this on a later PR to make sure that we can handle this

src/features/transfer/types.ts
Xaroz: this is done on a further PR

src/features/transfer/TransferTokenForm.tsx
Xaroz: will also be done on a further PR that is in progress

src/features/transfer/TransferTokenForm.tsx
Xaroz: This could be something interesting to explore to see the tradeoffs between using useMemo vs not, especially since this is relatively stable

src/features/tokens/TokenSelectField.tsx
Xaroz: good call!

src/features/tokens/SelectTokenIdField.tsx
Xaroz: I am gonna remove these props as they are not used but seems that they were mostly for NFTs interface which I have never seen used before

src/features/tokens/SelectOrInputTokenIds.tsx
Xaroz: same as above

src/features/tokens/hooks.ts
antigremlin: Wait, I thought we have one WarpCore instance per warp route. Do we use one instance for many routes in the UI?

src/features/tokens/hooks.ts
antigremlin: Do we guarantee to never have empty chainName and symbol?

Xaroz: looking at the `WarpCore.FromConfig` implementation:
`const parsedConfig = WarpCoreConfigSchema.parse(config);` 
when generating the token array it would throw if `chainName` and `symbol` were empty and the initialization of the app would fail.

Only thing that can happen is that `addressOrDenom` ends up as an empty string in which case I think we can probably

src/features/transfer/TransferTokenForm.tsx
Xaroz: This is a good idea and I think in general we could implement this even for the old architecture

src/features/tokens/hooks.ts
Xaroz: Yeah, we have one `WarpCore` which contains tokens created from many routes

antigremlin: If it's long living, I'd vote for introducing a map instead of linear search. We could do this in a separate PR.

PR #838, opened 2025-12-08
feat: origin and destination tokens array

Xaroz:
This work is part of the upcoming PR to have origin and destination token selector

fixes [ENG-2766](https://linear.app/hyperlane-xyz/issue/ENG-2766/create-separate-arrays-for-origin-tokens-and-destination-tokens) and [ENG-2767](https://linear.app/hyperlane-xyz/issue/ENG-2767/merge-tokens-that-have-the-same-collateral)

- Creates two separate arrays for tokens that are only origin and token that are only destinations
- Deduplicate these if they share the same collateral address
- Create `TokenChainIcon` component (might change name later)
- Remove deprecated `tokenIndex` related code

### DIFF
--- a/src/features/tokens/utils.ts
+++ b/src/features/tokens/utils.ts
@@ -128,6 +128,17 @@ export function getTokensWithSameCollateralAddresses(
     });
 }
 
+/**
+ * Generate a stable token key from a token object
+ * Uses chainName + lowercase symbol + normalized address
+ * Format: "chainName-symbol-addressOrDenom" (stable identifier)
+ * Example: "ethereum-usdc-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"
+ */
+export function getTokenKey(token: IToken): string {
+  const normalizedAddress = normalizeAddress(token.addressOrDenom, token.protocol);
+  return `${token.chainName.toLowerCase()}-${token.symbol.toLowerCase()}-${normalizedAddress}`;
+}
+
 // De-duplicate collaterized tokens
 // Returns a map of token with same origin and dest collateral address
 // And an array of tokens with repeated collateral addresses grouped into one
@@ -178,10 +189,13 @@ export function dedupeTokensByCollateral(tokens: Token[]): Token[] {
       return true;
     }
 
-    // For HypNative tokens, use their symbol and standard as identifier
+    const chainName = token.chainName.toLowerCase();
+    const symbol = token.symbol.toLowerCase();
+    const protocol = token.protocol;
+    // For HypNative tokens, use their symbol and protocol as identifier
     const collateralKey = token.collateralAddressOrDenom
-      ? `${token.chainName}-${token.symbol}-${normalizeAddress(token.collateralAddressOrDenom, token.protocol)}`
-      : `${token.chainName}-${token.symbol}-hypnative-${token.standard}`;
+      ? `${chainName}-${symbol}-${normalizeAddress(token.collateralAddressOrDenom, protocol)}`
+      : `${chainName}-${symbol}-hypnative-${protocol}`;
 
     // If we haven't seen this collateral on this chain, include it
     if (!seenCollaterals.has(collateralKey)) {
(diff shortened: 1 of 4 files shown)


src/features/tokens/utils.ts
```diff
@@ -162,3 +163,80 @@ export function dedupeMultiCollateralTokens(tokens: Tokens, destination) {
...
+  return tokens.filter((token) => {
+    // If not a collateralized token, include it
+    if (!TOKEN_COLLATERALIZED_STANDARDS.includes(token.standard)) {
+      return true;
+    }
+
+    // For HypNative tokens, use their symbol and standard as identifier
+    const collateralKey = token.collateralAddressOrDenom
+      ? `${token.chainName}-${token.symbol}-${normalizeAddress(token.collateralAddressOrDenom, token.protocol)}`
+      : `${token.chainName}-${token.symbol}-hypnative-${token.standard}`;
```
xeno097: Might be worth putting this into its own function

xeno097 (approved):

src/features/tokens/utils.ts
```diff
@@ -162,3 +163,80 @@ export function dedupeMultiCollateralTokens(tokens: Tokens, destination) {
...
+
+  return tokens.filter((token) => {
+    // If not a collateralized token, include it
+    if (!TOKEN_COLLATERALIZED_STANDARDS.includes(token.standard)) {
+      return true;
+    }
+
+    // For HypNative tokens, use their symbol and standard as identifier
+    const collateralKey = token.collateralAddressOrDenom
+      ? `${token.chainName}-${token.symbol}-${normalizeAddress(token.collateralAddressOrDenom, token.protocol)}`
```
antigremlin: This is inconsistent with `getTokenKey()` from the other PR. We normalise with `toLowerCase` there but not here. Maybe move the token key functions from hooks to this file and resue?

src/features/tokens/utils.ts
```diff
@@ -162,3 +163,80 @@ export function dedupeMultiCollateralTokens(tokens: Tokens, destination) {
...
+ * Deduplicates by address and by collateral address per chain
+ */
+export function buildOriginTokens(tokens: Token[]): Token[] {
+  // Filter to only tokens with outgoing connections
+  const originTokens = tokens.filter((token) => token.connections && token.connections.length > 0);
+
+  // Deduplicate by chain-address
+  const tokenMap = new Map<string, Token>();
+  originTokens.forEach((token) => {
+    const key = `${token.chainName}-${token.symbol}-${token.addressOrDenom}`;
```
antigremlin: Ditto, on key inconsistency

src/features/tokens/utils.ts
```diff
@@ -162,3 +163,80 @@ export function dedupeMultiCollateralTokens(tokens: Tokens, destination) {
...
+ * Deduplicates by address and by collateral address per chain
+ */
+export function buildDestinationTokens(tokens: Token[]): Token[] {
+  const tokenMap = new Map<string, Token>();
+
+  // Traverse all connections to find destination tokens
+  tokens.forEach((token) => {
+    token.connections?.forEach((conn) => {
+      const destToken = conn.token as Token;
+      const key = `${destToken.chainName}-${destToken.symbol}-${destToken.addressOrDenom}`;
```
antigremlin: Ditto, on key inconsistency

antigremlin (approved):

src/features/tokens/utils.ts
Xaroz: Moved `getTokenKey` to this file, as for the other token key functions my reasoning for not moving them its that they use WarpCore as a param and although a silly reasoning, warpCore comes from a hook and it seemed to make more sense to keep those in the hooks folder.

Also updated the key here to use lowercase and be more consistent, didn't use getTokenKey for this specific one because this one uses `collateralAddressOrDenom` instead and I preferred to keep the original function signature the same.

src/features/tokens/utils.ts
Xaroz: done

PR #848, opened 2025-12-11
chore: sort by collateral balance in getLowestFeeTransferToken

Xaroz:
- Sorts the tokens by collateral balance from highest to lowest before returning the actual token in `getLowestFeeTransferToken`
- Separates the sorting function into `filterAndSortTokensByBalance` and `sortByTokensFee`
- Created unit tests for the new functions and the main flow for getLowestFeeTransferToken

### DIFF
--- a/src/utils/promises.ts
+++ b/src/utils/promises.ts
@@ -1,0 +1,9 @@
+/**
+ * Extracts fulfilled values from Promise.allSettled results, filtering out rejected and null values.
+ */
+export function getPromisesFulfilledValues<T>(results: PromiseSettledResult<T | null>[]): T[] {
+  return results
+    .filter((r) => r.status === 'fulfilled')
+    .map((r) => r.value)
+    .filter((v) => v != null);
+}


src/features/transfer/fees.ts
```diff
@@ -2,8 +2,56 @@ import { IToken, Token, TokenAmount, WarpCore } from '@hyperlane-xyz/sdk';
...
+      result.status === 'fulfilled' &&
+      result.value?.balance &&
+      result.value.balance >= minAmount
+    ) {
+      tokenBalances.push(result.value);
+    }
+  }
+
+  // sort by descending balance (highest to lowest)
+  tokenBalances.sort((a, b) => compareByBalanceDesc(a.balance, b.balance));
```
antigremlin: The name isn't true (the function compares numbers, we extract balances at call site). Inline or turn into compareByBalanceDesc(a: {balance: bigint}, b: {balance: bigint})?

Considering my other comments, this function will be reduced to:
```typescript
tokenBalances.filter(t => t.balance >= minAmount).sort(compareByBalanceDesc)
```

src/features/tokens/types.ts
```diff
@@ -8,3 +8,15 @@ export type TokenChainMap = {
 };
 
 export type Tokens = Array<{ token: Token; disabled: boolean }>;
+
+export interface TokensWithDestinationBalance {
+  originToken: Token;
+  destinationToken: Token;
+  balance: bigint;
```
antigremlin, line 15: Inconsistent field names. Why `TokenWithDestinationBalance.balance` vs. `TokenWithFee.tokenBalance`?

src/features/transfer/fees.ts
```diff
@@ -92,24 +140,13 @@ export async function getLowestFeeTransferToken(
...
-  for (const result of balanceResults) {
-    if (
-      result.status === 'fulfilled' &&
-      result.value?.balance &&
-      result.value.balance >= amountWeiBigInt
-    ) {
-      tokenBalances.push(result.value);
-    }
-  }
+  const tokenBalances = filterAndSortTokensByBalance(balanceResults, BigInt(amountWei));
```
antigremlin: It would be better to keep promise-settled filtering close to the `Promise.allSettled` call because we may later decide to handle errors somehow; then they can be handled differently depending on the call, etc. And then `filterAndSort...()` no longer needs to use the `PromiseSettledResult<>` type at all.

src/features/transfer/fees.ts
```diff
@@ -118,36 +155,25 @@ export async function getLowestFeeTransferToken(
...
     }),
   );
 
-  const tokenFees: Array<{ token: Token; tokenFee?: TokenAmount }> = [];
+  const tokenFees: TokenWithFee[] = [];
   for (const result of feeResults) {
     if (result.status === 'fulfilled' && result.value) {
-      tokenFees.push({ token: result.value.token, tokenFee: result.value.fees.tokenFeeQuote });
+      const { fees, token, tokenBalance } = result.value;
+      tokenFees.push({ token, tokenFee: fees.tokenFeeQuote, tokenBalance });
```
antigremlin: Or with nice inferred typing:
```typescript
const goodFeeResults = feeResult.filter(r => r.status === 'fulfilled').map(r => r.value).filter(v => v != null);
const tokenFees = goodFeeResults.map(({token, fees, tokenBalance }) => ({
  token,
  tokenFee: fees.tokenFeeQuote,
  tokenBalance
}));
```

src/features/transfer/fees.test.ts
```diff
@@ -253,14 +240,490 @@ describe('getTotalFee', () => {
...
+    // Should be sorted: token2 (500) > token3 (300) > token1 (100)
+    expect(result[0].originToken).toBe(token2);
+    expect(result[0].balance).toBe(BALANCE_SMALL);
+    expect(result[1].originToken).toBe(token3);
+    expect(result[1].balance).toBe(BigInt(300));
+    expect(result[2].originToken).toBe(token1);
+    expect(result[2].balance).toBe(BALANCE_TINY);
+  });
+
+  test('should handle rejected promises', () => {
```
antigremlin: A better name would be "should ignore rejected promises"

antigremlin (commented):
This can be simplified a bit

src/features/transfer/fees.test.ts
Xaroz: Updated this test as promise checking wouldn't be in this function anymore

PR #869, opened 2025-12-18
fix: update deps to build under pnpm link

antigremlin:
These are draft fixes to the build system to allow building under `pnpm link`. We will try to avoid `pnpm link` in general but it will take more work and most of the fixes are correct even without it (except that the starknet dependency is questionable).

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5,6 +5,8 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
+  '@chain-registry/types': 0.50.269
+  '@cosmjs/math': 0.32.4
   '@solana/web3.js': ^1.95.4
   axios: ^1.7.9
   bignumber: 9.1.2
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -63,10 +63,11 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-toastify": "^10.0.6",
+    "starknet": "7.6.4",
```
antigremlin, line 66: Starknet fails to resolve as a peer dependency under `pnpm link` because of a major version conflict.

package.json
```diff
@@ -63,10 +63,11 @@
...
     "starknetkit": "2.6.1",
-    "viem": "^2.21.41",
+    "viem": "^2.21.45",
```
antigremlin, line 68: This is to align versions with the monorepo

package.json
```diff
@@ -63,10 +63,11 @@
...
     "wagmi": "^2.12.26",
-    "zod": "3.21.4",
+    "zod": "3.25.76",
```
antigremlin, line 70: This is to align versions with the monorepo

PR #871, opened 2025-12-19
feat: Unified Chain Token Selector Modal

Xaroz:
This PR focuses on creating the modal to so that users can have one unified modal for selecting a token that also includes the chains.

- Creates `Modal` to show both token selector and chain selector at the same time
- Aside from showing the `Token` Logo, they will also include the chains it belongs to
- When opening the modal, it will show the tokens that have matching routes first from the opposite selected token
- Create logic to group same collateral tokens into one. The key (if a collateral token) will be a combination of the collateral address plus chain name, symbol and the values will be an array of all the tokens that matches this description.
- Since we are highlighting tokens that can be bridged from and to, removed the duplication of `originTokens` and `destinationTokens` as separate array and just unified into one, while filtering out tokens that are neither `origin` or `destination`.
- Change URL params logic to fit `originToken` and `destinationToken` selection

Fixes [ENG-2765](https://linear.app/hyperlane-xyz/issue/ENG-2765/create-unified-modal-for-chain-and-token-selection-at-the-same-time)

### DIFF
--- a/src/features/tokens/utils.ts
+++ b/src/features/tokens/utils.ts
@@ -8,7 +8,7 @@ import {
 } from '@hyperlane-xyz/sdk';
 import { eqAddress, isNullish, normalizeAddress } from '@hyperlane-xyz/utils';
 import { isChainDisabled } from '../chains/utils';
-import { MultiCollateralTokenMap, TokenChainMap, Tokens } from './types';
+import { TokenChainMap } from './types';
 
 // Map of token symbols and token chain map
 // Symbols are not duplicated to avoid the same symbol from being shown
@@ -141,42 +141,6 @@ export function getTokenKey(token: IToken): string {
   return `${token.chainName.toLowerCase()}-${token.symbol.toLowerCase()}-${normalizedAddress}`;
 }
 
-// De-duplicate collaterized tokens
-// Returns a map of token with same origin and dest collateral address
-// And an array of tokens with repeated collateral addresses grouped into one
-export function dedupeMultiCollateralTokens(tokens: Tokens, destination: ChainName) {
-  return tokens.reduce<{ tokens: Tokens; multiCollateralTokenMap: MultiCollateralTokenMap }>(
-    (acc, t) => {
-      const originToken = t.token;
-      const isMultiCollateralToken = isValidMultiCollateralToken(originToken, destination);
-      if (!isMultiCollateralToken) return { ...acc, tokens: [...acc.tokens, t] };
-
-      const destinationToken = originToken.getConnectionForChain(destination)!.token;
-
-      // For HypNative tokens, use their symbol and standard as identifier since they don't have collateralAddressOrDenom
-      const originAddress = originToken.collateralAddressOrDenom
-        ? normalizeAddress(originToken.collateralAddressOrDenom, originToken.protocol)
-        : `hypnative-${originToken.standard}-${originToken.symbol}`;
-
-      const destinationAddress = destinationToken.collateralAddressOrDenom
-        ? normalizeAddress(destinationToken.collateralAddressOrDenom, destinationToken.protocol)
-        : `hypnative-${destinationToken.standard}-${destinationToken.symbol}`;
-
-      // now origin and destination are both collaterals
-      // create map for tokens with same origin and destination collateral addresses
-      acc.multiCollateralTokenMap[originAddress] ||= {};
-      if (!acc.multiCollateralTokenMap[originAddress][destinationAddress]) {
-        acc.multiCollateralTokenMap[originAddress][destinationAddress] = [];
-        acc.tokens.push(t);
-      }
-
-      acc.multiCollateralTokenMap[originAddress][destinationAddress].push(originToken);
-      return acc;
-    },
-    { tokens: [], multiCollateralTokenMap: {} },
-  );
-}
-
 /**
  * De-duplicate tokens by collateral address on the same chain
  * Returns only one token per unique collateral address per chain
@@ -304,4 +268,42 @@ export function checkTokenHasRoute(
     if (!destConnection?.token) return false;
     return getCollateralKey(destConnection.token) === destCollateralKey;
   });
+}
+
+/**
+ * Find the actual warpCore token that has a route to the destination.
+ * The passed originToken may be from a deduplicated array and may not have
+ * the connection, but a token with the same collateral in warpCore does.
+ */
+export function findRouteToken(
+  warpCore: WarpCore,
+  originToken: Token,
+  destinationChain: string,
+): Token | undefined {
+  // First check if the passed token already has the connection
+  if (originToken.getConnectionForChain(destinationChain)) {
+    return originToken;
+  }
+
+  // Otherwise, find a token from warpCore that has the route and shares collateral
+  const routeTokens = warpCore.getTokensForRoute(originToken.chainName, destinationChain);
+  if (routeTokens.length === 0) return undefined;
+
+  const normalizedOriginCollateral = originToken.collateralAddressOrDenom
+    ? normalizeAddress(originToken.collateralAddressOrDenom, originToken.protocol)
+    : null;
+
+  // Find a route token that shares collateral with the origin token
+  const matchingToken = routeTokens.find((t) => {
+    const normalizedRouteCollateral = t.collateralAddressOrDenom
+      ? normalizeAddress(t.collateralAddressOrDenom, t.protocol)
+      : null;
+    // Match by collateral address if both have one, otherwise match by symbol
+    if (normalizedOriginCollateral && normalizedRouteCollateral) {
+      return normalizedOriginCollateral === normalizedRouteCollateral;
+    }
+    return t.symbol === originToken.symbol;
+  });
+
+  return matchingToken || routeTokens[0];
 }
--- a/src/features/transfer/TransferTokenForm.tsx
+++ b/src/features/transfer/TransferTokenForm.tsx
@@ -52,11 +52,11 @@ import {
   useWarpCore,
 } from '../tokens/hooks';
 import { useTokenPrice } from '../tokens/useTokenPrice';
-import { checkTokenHasRoute } from '../tokens/utils';
+import { checkTokenHasRoute, findRouteToken } from '../tokens/utils';
 import { WalletConnectionWarning } from '../wallet/WalletConnectionWarning';
 import { FeeSectionButton } from './FeeSectionButton';
 import { RecipientConfirmationModal } from './RecipientConfirmationModal';
-import { findRouteToken, getInterchainQuote, getLowestFeeTransferToken, getTotalFee } from './fees';
+import { getInterchainQuote, getLowestFeeTransferToken, getTotalFee } from './fees';
 import { useFetchMaxAmount } from './maxAmount';
 import { TransferFormValues } from './types';
 import { useRecipientBalanceWatcher } from './useBalanceWatcher';
--- a/src/features/transfer/fees.ts
+++ b/src/features/transfer/fees.ts
@@ -1,10 +1,14 @@
 import { IToken, Token, TokenAmount, WarpCore } from '@hyperlane-xyz/sdk';
-import { normalizeAddress, objKeys } from '@hyperlane-xyz/utils';
+import { objKeys } from '@hyperlane-xyz/utils';
 import { chainsRentEstimate } from '../../consts/chains';
 import { logger } from '../../utils/logger';
 import { getPromisesFulfilledValues } from '../../utils/promises';
 import { TokensWithDestinationBalance, TokenWithFee } from '../tokens/types';
-import { getTokensWithSameCollateralAddresses, isValidMultiCollateralToken } from '../tokens/utils';
+import {
+  findRouteToken,
+  getTokensWithSameCollateralAddresses,
+  isValidMultiCollateralToken,
+} from '../tokens/utils';
 
 // Compare two objects with balance field in descending order (highest first)
 export function compareByBalanceDesc(a: { balance: bigint }, b: { balance: bigint }) {
@@ -85,44 +89,6 @@ export function getInterchainQuote(
   return originToken && objKeys(chainsRentEstimate).includes(originToken.chainName)
     ? interchainQuote.plus(chainsRentEstimate[originToken.chainName])
     : interchainQuote;
-}
-
-/**
- * Find the actual warpCore token that has a route to the destination.
- * The passed originToken may be from a deduplicated array and may not have
- * the connection, but a token with the same collateral in warpCore does.
- */
-export function findRouteToken(
-  warpCore: WarpCore,
-  originToken: Token,
-  destinationChain: string,
-): Token | undefined {
-  // First check if the passed token already has the connection
-  if (originToken.getConnectionForChain(destinationChain)) {
-    return originToken;
-  }
-
-  // Otherwise, find a token from warpCore that has the route and shares collateral
-  const routeTokens = warpCore.getTokensForRoute(originToken.chainName, destinationChain);
-  if (routeTokens.length === 0) return undefined;
-
-  const normalizedOriginCollateral = originToken.collateralAddressOrDenom
-    ? normalizeAddress(originToken.collateralAddressOrDenom, originToken.protocol)
-    : null;
-
-  // Find a route token that shares collateral with the origin token
-  const matchingToken = routeTokens.find((t) => {
-    const normalizedRouteCollateral = t.collateralAddressOrDenom
-      ? normalizeAddress(t.collateralAddressOrDenom, t.protocol)
-      : null;
-    // Match by collateral address if both have one, otherwise match by symbol
-    if (normalizedOriginCollateral && normalizedRouteCollateral) {
-      return normalizedOriginCollateral === normalizedRouteCollateral;
-    }
-    return t.symbol === originToken.symbol;
-  });
-
-  return matchingToken || routeTokens[0];
 }
 
 // Checks if a token is a multi-collateral token and if so
--- a/src/features/transfer/maxAmount.ts
+++ b/src/features/transfer/maxAmount.ts
@@ -7,7 +7,8 @@ import { logger } from '../../utils/logger';
 import { useMultiProvider } from '../chains/hooks';
 import { isMultiCollateralLimitExceeded } from '../limits/utils';
 import { useWarpCore } from '../tokens/hooks';
-import { findRouteToken, getLowestFeeTransferToken } from './fees';
+import { findRouteToken } from '../tokens/utils';
+import { getLowestFeeTransferToken } from './fees';
 
 interface FetchMaxParams {
   accounts: Record<ProtocolType, AccountInfo>;


src/features/analytics/utils.ts
```diff
@@ -55,13 +55,19 @@ export function trackTransactionFailedEvent(
...
+  if (!token) return;
+
+  const origin = token.chainName;
   const { address } = getAccountAddressAndPubKey(warpCore.multiProvider, origin, accounts);
-  const token = overrideToken || getTokenByKey(warpCore, tokenKey);
 
-  if (!token) return;
+  // Find destination token to get destination chain
+  const destToken = warpCore.tokens.find((t) => getTokenKey(t) === destinationTokenKey);
+  const destination = destToken?.chainName || '';
```
xeno097: should we allow this to be an empty string? shouldn't we enforce that this is defined?

src/features/analytics/utils.ts
```diff
@@ -55,13 +55,19 @@ export function trackTransactionFailedEvent(
...
 
-  const originChainId = warpCore.multiProvider.getChainId(origin);
-  const destinationChainId = warpCore.multiProvider.getChainId(destination);
+  const originChainId = warpCore.multiProvider.tryGetChainId(origin);
+  const destinationChainId = destination ? warpCore.multiProvider.tryGetChainId(destination) : null;
   return trackEvent(EVENT_NAME.TRANSACTION_SUBMISSION_FAILED, {
     amount,
     chains: `${origin}|${originChainId}|${destination}|${destinationChainId}`,
```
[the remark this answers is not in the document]
Xaroz: this shouldn't really happen but I will address analytics in a further PR (plus Vercel does take null values)

src/features/tokens/TokenList.tsx
```diff
@@ -0,0 +1,209 @@
...
+        chainDisplayName.includes(q)
+      );
+    });
+
+    // Sort: tokens with routes first, then by symbol, then by chain name
+    const sorted = [...filtered].sort((a, b) => {
+      // If we have route info, sort tokens with routes first
+      if (tokenRouteMap) {
+        const aHasRoute = tokenRouteMap.get(getTokenKey(a)) ?? true;
+        const bHasRoute = tokenRouteMap.get(getTokenKey(b)) ?? true;
```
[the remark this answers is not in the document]
Xaroz: Will address this in a follow up commit

src/features/tokens/hooks.ts
```diff
@@ -1,138 +1,123 @@
...
-    const queryToken = tokensWithRoute.find(
-      (token) =>
-        token.addressOrDenom === addressOrDenom &&
-        token.symbol.toLowerCase() === symbol.toLowerCase(),
-    );
-    if (queryToken) return getTokenKey(queryToken);
+  tokens: Token[],
+): { originTokenKey: string | undefined; destinationTokenKey: string | undefined } {
+  // Early return if no tokens
+  if (tokens.length === 0) {
```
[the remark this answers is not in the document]
Xaroz: there is no issues here as it is meant to return undefined if no tokens are found

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -337,33 +238,49 @@ function AmountSection({ isNft, isReview }: { isNft: boolean; isReview: boolean
 
 function RecipientSection({ isReview }: { isReview: boolean }) {
   const { values } = useFormikContext<TransferFormValues>();
-  const { balance } = useDestinationBalance(values);
-  useRecipientBalanceWatcher(values.recipient, balance);
+  const tokens = useTokens();
+  const destinationToken = getTokenByKey(tokens, values.destinationTokenKey);
+
+  const { balance } = useDestinationBalance(values, destinationToken);
+  useRecipientBalanceWatcher(destinationToken?.chainName || '', balance);
```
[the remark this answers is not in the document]
Xaroz: There is a check inside the `useRecipientBalanceWatcher` that will prevent anything from happen if chain name is not provided

src/features/analytics/utils.ts
Xaroz: fixed

src/features/store.ts
```diff
@@ -206,8 +209,8 @@ export const useStore = create<AppState>()(
       },
       tokensBySymbolChainMap: {},
       routerAddressesByChainMap: {},
-      originTokens: [],
-      destinationTokens: [],
+      tokens: [],
+      collateralGroups: new Map(),
```
antigremlin, line 213: I can see we are carefully handling this `Map()` with zustand and exclude it from localStorage serialization. Should we add a comment explaining this? Why do we want a Map() instead of a plain object?

src/consts/args.ts
```diff
@@ -3,7 +3,8 @@ import { ProtocolType } from '@hyperlane-xyz/utils';
 export enum WARP_QUERY_PARAMS {
   ORIGIN = 'origin',
   DESTINATION = 'destination',
-  TOKEN = 'token',
```
antigremlin, line 6: Do we care for backward compatibility of the query params?

src/consts/config.ts
```diff
@@ -15,8 +15,8 @@ const rpcOverrides = process?.env?.NEXT_PUBLIC_RPC_OVERRIDES || '';
 interface Config {
   addressBlacklist: string[]; // A list of addresses that are blacklisted and cannot be used in the app
   chainWalletWhitelists: ChainMap<string[]>; // A map of chain names to a list of wallet names that work for it
-  defaultOriginChain: string | undefined; // The initial origin chain to show when app first loads
-  defaultDestinationChain: string | undefined; // The initial destination chain to show when app first loads
+  defaultOriginToken: string | undefined; // The initial origin token to show when app first loads (format: chainName-symbol, e.g. "ethereum-hyper")
```
antigremlin, line 18: The format is actually `chainName-symbol-address`

src/features/chains/ChainConnectionWarning.tsx
```diff
@@ -41,24 +39,11 @@ export function ChainConnectionWarning({
...
-
-  const onClickEdit = () => {
-    if (!unhealthyChain) return;
-    setIsModalOpen(true);
-  };
-
   return (
     <>
-      <FormWarningBanner isVisible={!!unhealthyChain} cta="Edit" onClick={onClickEdit}>
+      <FormWarningBanner isVisible={!!unhealthyChain}>
```
antigremlin, line 44: Why do we remove the edit button and the related logic?

src/features/tokens/hooks.ts
```diff
@@ -1,138 +1,123 @@
...
+import { tryGetValidChainName } from '../chains/utils';
 import { useStore } from '../store';
 import { getTokenKey } from './utils';
 
 export function useWarpCore() {
   return useStore((s) => s.warpCore);
 }
 /**
- * Find a token by its key
+ * Find a token by its key from a WarpCore or Token array
```
antigremlin, line 16: But we don't accept WarpCore any longer

src/features/tokens/TokenListPanel.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <div className="shrink-0 border-b border-gray-200 px-4 py-4">
+        <SearchInput
+          inputRef={inputRef}
+          value={searchQuery}
+          onChange={onSearchChange}
+          placeholder="Search for a token or paste address"
+          aria-label="Search tokens"
+        />
+        {/* Mobile chain quick select (hidden on desktop) */}
+        <div className="mt-3 md:hidden">
```
antigremlin, line 52: Should we `useIsMobile()` for conditional rendering instead of `md:hidden`?

src/features/chains/MobileChainQuickSelect.tsx
```diff
@@ -0,0 +1,116 @@
...
+  onSelectChain,
+  onMoreClick,
+  preferredChains,
+}: MobileChainQuickSelectProps) {
+  const { chainMetadata } = useStore((s) => ({
+    chainMetadata: s.chainMetadata,
+  }));
+
+  // Build and sort all chains - only rebuilds when chainMetadata changes
+  const allChains = useMemo(() => {
```
antigremlin, line 27: The `allChains` computation seems to be duplicated from `<ChainList>`

src/features/tokens/TokenSelectField.tsx
```diff
@@ -1,98 +1,125 @@
...
-  setIsNft: (value: boolean) => void;
+  setIsNft?: (value: boolean) => void;
 };
 
-export function TokenSelectField({ name, disabled, setIsNft }: Props) {
-  const { values, setValues } = useFormikContext<TransferFormValues>();
-  const [tokenField, , helpers] = useField<string | undefined>(name);
+export function TokenSelectField({ name, label, selectionMode, disabled, setIsNft }: Props) {
+  const { values, setFieldValue } = useFormikContext<TransferFormValues>();
+  const [field, , helpers] = useField<string | undefined>(name);
```
antigremlin, line 31: Could add clarity with `const [{value: tokenKey},, {setValue: setTokenKey}] =`

src/features/tokens/TokenSelectField.tsx
```diff
@@ -1,98 +1,125 @@
...
-        onClick={onClickField}
-        isAutomatic={isAutomaticSelection}
-      />
-      <TokenListModal
+      <div className="flex flex-col">
+        <label htmlFor={name} className="mb-1 pl-0.5 text-sm text-gray-600">
+          {label}
+        </label>
+        <TokenButton
+          token={token}
```
antigremlin, line 100: Could improve clarity with
```
token={selectedToken}
onClick={doOpenModal}
```

src/features/tokens/TokenSelectField.tsx
```diff
@@ -1,98 +1,125 @@
...
+          disabled={disabled}
+          onClick={onClickField}
+          multiProvider={multiProvider}
+        />
+        {canAddAsset && (
+          <button
+            type="button"
+            className={styles.addButton}
+            onClick={onAddToken}
```
antigremlin, line 109: Could improve clarity with
```
onClick={doAddToken}
disabled={isAddingToken}
```

src/features/tokens/TokenSelectField.tsx
```diff
@@ -1,98 +1,125 @@
...
+          >
+            <PlusIcon height={16} width={16} /> Import token to wallet
+          </button>
+        )}
+      </div>
+
+      <UnifiedTokenChainModal
         isOpen={isModalOpen}
         close={() => setIsModalOpen(false)}
         onSelect={onSelectToken}
```
antigremlin, line 120: Could improve clarity with `onSelect={doSelectToken}`

src/features/tokens/UnifiedTokenChainModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+  close,
+  onSelect,
+  selectionMode,
+  counterpartToken,
+}: Props) {
+  const [chainSearch, setChainSearch] = useState('');
+  const [tokenSearch, setTokenSearch] = useState('');
+  const [selectedChain, setSelectedChain] = useState<ChainName | null>(null);
+  // Mobile-only state: whether to show the full chain list
+  const [showMobileChainList, setShowMobileChainList] = useState(false);
```
antigremlin, line 28: Should we use conditional rendering rather than CSS hiding? The two variants seem mutually exclusive, right?

src/features/tokens/UnifiedTokenChainModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+      panelClassname="p-0 max-w-sm md:max-w-4xl overflow-hidden"
+    >
+      <div className="flex max-h-[80vh] min-h-[24rem]">
+        {/* Chain filter panel: always visible on desktop, conditionally visible on mobile */}
+        <div className={`${showMobileChainList ? 'flex flex-1' : 'hidden'} md:flex md:flex-none`}>
+          <ChainFilterPanel
+            searchQuery={chainSearch}
+            onSearchChange={setChainSearch}
+            selectedChain={selectedChain}
+            onSelectChain={onMobileSelectChain}
```
antigremlin, line 62: May be easier to read as `onSelectChain={mobileSelectChain}`.

src/features/tokens/UnifiedTokenChainModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </div>
+
+        {/* Token list panel: hidden on mobile when showing chain list */}
+        <div className={`min-w-0 flex-1 ${showMobileChainList ? 'hidden md:flex' : 'flex'}`}>
+          <TokenListPanel
+            selectionMode={selectionMode}
+            searchQuery={tokenSearch}
+            onSearchChange={setTokenSearch}
+            chainFilter={selectedChain}
+            onSelect={onSelectAndClose}
```
antigremlin, line 75: May be easier to read as `onSelect={selectAndClose}`.

src/features/tokens/utils.ts
```diff
@@ -58,12 +58,14 @@ export function isValidMultiCollateralToken(
...
-    (!destinationToken.collateralAddressOrDenom && !destinationToken.isHypNative()) ||
-    !TOKEN_COLLATERALIZED_STANDARDS.includes(destinationToken.standard)
-  )
-    return false;
+  // Check if destination is collateralized - has collateral address, is HypNative, or is in SDK's list
+  // Note: Some standards like EvmHypCollateralFiat may not be in TOKEN_COLLATERALIZED_STANDARDS
+  const isDestCollateralized =
+    !!destinationToken?.collateralAddressOrDenom ||
+    destinationToken?.isHypNative?.() ||
+    TOKEN_COLLATERALIZED_STANDARDS.includes(destinationToken?.standard as any);
```
antigremlin, line 66: We can avoid `as any` by checking for non-empty `.standard` first or by casting `TOKEN_COLLATERALIZED_STANDARDS as string[]`

src/features/tokens/utils.ts
```diff
@@ -189,13 +155,7 @@ export function dedupeTokensByCollateral(tokens: Token[]): Token[] {
...
     }
 
-    const chainName = token.chainName.toLowerCase();
-    const symbol = token.symbol.toLowerCase();
-    const protocol = token.protocol;
-    // For HypNative tokens, use their symbol and protocol as identifier
-    const collateralKey = token.collateralAddressOrDenom
-      ? `${chainName}-${symbol}-${normalizeAddress(token.collateralAddressOrDenom, protocol)}`
-      : `${chainName}-${symbol}-hypnative-${protocol}`;
+    const collateralKey = getCollateralKey(token);
```
antigremlin, line 158: Please add tests for `dedupeTokensByCollateral`

src/features/tokens/utils.ts
```diff
@@ -209,48 +169,141 @@ export function dedupeTokensByCollateral(tokens: Token[]): Token[] {
...
- * Deduplicates by address and by collateral address per chain
+ * Build a unified tokens array containing all tokens that can participate in transfers
+ * (either as origin or destination). Deduplicates by address and by collateral.
  */
-export function buildOriginTokens(tokens: Token[]): Token[] {
-  // Filter to only tokens with outgoing connections
-  const originTokens = tokens.filter((token) => token.connections && token.connections.length > 0);
-
-  // Deduplicate by chain-address
+export function buildTokensArray(warpCoreTokens: Token[]): Token[] {
```
antigremlin, line 175: Please add tests for `buildTokensArray`

src/features/tokens/utils.ts
```diff
@@ -209,48 +169,141 @@ export function dedupeTokensByCollateral(tokens: Token[]): Token[] {
...
+
+  // Deduplicate tokens that have same collateral address on the same chain
+  return dedupeTokensByCollateral(Array.from(tokenMap.values()));
+}
+
+/**
+ * Build collateral groups - groups tokens by their collateral key for O(1) lookup
+ * Used for fast route checking in the token selection modal
+ */
+export function buildCollateralGroups(tokens: Token[]): Map<string, Token[]> {
```
antigremlin, line 207: A clearer name could be `groupTokensByCollateral`

src/features/tokens/utils.ts
```diff
@@ -209,48 +169,141 @@ export function dedupeTokensByCollateral(tokens: Token[]): Token[] {
...
+  const groups = new Map<string, Token[]>();
+  for (const token of tokens) {
+    const key = getCollateralKey(token);
+    const existing = groups.get(key) || [];
+    existing.push(token);
+    groups.set(key, existing);
```
antigremlin, line 213: nit: The `.set()` is redundant when adding to existing group.

src/features/tokens/utils.ts
```diff
@@ -209,48 +169,141 @@ export function dedupeTokensByCollateral(tokens: Token[]): Token[] {
...
+/**
+ * Check if a route exists between origin and destination tokens
+ * Uses pre-computed collateral groups for fast O(1) lookups
+ *
+ * @param originToken - The origin token (what the user is sending)
+ * @param destToken - The destination token (what the user will receive)
+ * @param collateralGroups - Pre-computed map of collateral key → tokens
+ * @returns true if a valid route exists between the tokens
+ */
+export function checkTokenHasRoute(
```
antigremlin, line 256: Please add tests for `checkTokenHasRoute`

src/features/tokens/utils.ts
```diff
@@ -209,48 +169,141 @@ export function dedupeTokensByCollateral(tokens: Token[]): Token[] {
...
   });
+}
 
-  const dedupedTokens = Array.from(tokenMap.values());
+/**
+ * Find the actual warpCore token that has a route to the destination.
+ * The passed originToken may be from a deduplicated array and may not have
+ * the connection, but a token with the same collateral in warpCore does.
+ */
+export function findRouteToken(
```
antigremlin, line 278: Is it easy enough to mock warpCore to test this function as well?

antigremlin (commented):
Great work! Almost all files reviewed, just a few remaining. Please see my comments.

src/features/store.ts
Xaroz: I could add a comment but basically:
1. We don't wanna persist this in the localStorage because the `collateralGroup` is a derived state from the `warpCore.tokens` that is built through the registry, any changes in the registry (as in added or removed) would mean the `collateralGroup` is out of sync
2. There is really not big difference between map vs just a plain object here, but I wanted this to be a strict dictionary for collateral groups lookup and objects include the prototype chain so having a map felt more right for this use case

src/consts/args.ts
Xaroz: Nope, if any user has wrong params they would simply be ignored, we want to move forward with strictly the new way of handling query params, at least imo

src/consts/config.ts
Xaroz: Ah in this case it doesn't matter, for the initial load I set it to find the first chain-symbol combination without relying on the address as that can be removed with the dedupe logic

src/features/chains/ChainConnectionWarning.tsx
Xaroz: Old Chain Select allowed you the option to edit a chain metadata and update the RPC, that is on my todo list for after I complete the whole rebranding

src/features/tokens/hooks.ts
Xaroz: I think it would be more correct to say `WarpCore.tokens` here

src/features/tokens/TokenListPanel.tsx
Xaroz: We usually try to avoid using this hook unless we really need to for performances reasons

src/features/tokens/TokenSelectField.tsx
Xaroz: you mean for `TransferFormValues`? The values and setFieldValue includes the whole form values and not just specifically for token-key as we want to clear some fields when values are updated and some logic depends on other values aside from the current token

src/features/tokens/UnifiedTokenChainModal.tsx
Xaroz: yeah, same comment as above about using the hook, we want to use it as little as possible

src/features/tokens/UnifiedTokenChainModal.tsx
Xaroz: `mobileSelectChain` makes it sound more like a state rather than this function being called when the an user select a chain

src/features/tokens/TokenListPanel.tsx
antigremlin: Do you know why we avoid doing this?

src/features/tokens/TokenSelectField.tsx
antigremlin: No, I mean this particular field:
```typescript
const [{value: tokenKey},, {setValue: setTokenKey}] = useField<>(name);
```

This will make subsequent code easier to understand. We aren't using other helpers.

src/features/tokens/UnifiedTokenChainModal.tsx
antigremlin: This is just a nit. How about `selectChainMobile`?

antigremlin (commented):
We agreed to resolve the comments in the next PR.

src/features/tokens/TokenListPanel.tsx
Xaroz: There are a few reasons but the main ones are:
1. Every window resize would trigger an state update and causing re-renders (Even if there is no resize of the browser window, which is usually the case, there is still a listener just sitting there)
2. Because the logic depends on JS, if loading JS is slow or somehow fails then layout breaks

antigremlin:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

src/components/buttons/ConnectAwareSubmitButton.tsx
```diff
@@ -58,7 +58,7 @@ export function ConnectAwareSubmitButton<FormValues = any>({
 
   return (
     <SolidButton
-      disabled={disabled}
+      disabled={disabled && isAccountReady}
```
antigremlin, line 61: This _seems_ like inverted logic. Should this be `disabled || !isAccountReady`?

antigremlin (approved):

PR #882, opened 2026-01-07
feat: Warp UI v2

Xaroz:
## Summary

Major v2 overhaul of the transfer experience — replaces the separate chain + token selection flows with a unified picker, restructures token identification around string keys, and adds new chain filtering/sorting capabilities.

### New: Unified Token/Chain Picker
Replaces the old `ChainSelectModal` + `TokenListModal` with a single `UnifiedTokenChainModal` where chain selection acts as a filter for tokens. Users pick a token and chain together in one flow.
- `ChainFilterPanel` side panel with search, protocol/network-type filters, name/chainId/protocol sort
- `ChainList` with "All Chains" and per-chain buttons
- `MobileChainQuickSelect` horizontal pill selector for mobile
- `ChainEditModal` for editing chain RPCs/block explorers
- Disabled chains support (grayed out, unclickable, sorted to bottom)

### New: Token Identification Refactor (`tokenIndex` → `tokenKey`)
Tokens identified by composite string key (`chain-symbol-address`) instead of numeric array index. More stable — indices shift when warp routes change, keys don't.
- `TransferFormValues` changed from `{ origin, destination, tokenIndex }` to `{ originTokenKey, destinationTokenKey }`
- Origin/destination chains derived from token keys (no separate chain form fields)
- Pre-computed `tokenByKeyMap` in store for O(1) lookups
- `getTokenKey()` / `getCollateralKey()` with WeakMap caching

### New: Token Deduplication & Store Pre-computation
- `buildTokensArray()` creates unified, deduplicated token array at init
- `groupTokensByCollateral()` pre-computes collateral groups
- Store extended with `tokens[]`, `collateralGroups`, `tokenByKeyMap`

### New: UI Components
- `TransferSection` card with gradient header/shadow
- `TokenChainIcon` composite icon (token + chain badge overlay)
- `WalletDropdown` inline wallet connect per chain
- `RecipientAddressModal`, `SearchInput`, `ModalHeader`, `SwapIcon`, `ImportTokenButton`

### Adapted from main: Token Balances in Picker
Balance display + USD values already exist on main (#952, #956, #964). V2 adapts the hook signatures to be more flexible — `useTokenBalances(tokens, scope, addressOverride?)` instead of origin/destination-specific. Same underlying multicall3 (EVM) / batched RPC (Sealevel) / CoinGecko infrastructure.

### Adapted from main: Visual Rebrand
Base rebrand already merged on main (#934). V2 extends it with the new components above (gradient cards, composite icons, unified modal layout).

### Carried over unchanged: Fee USD Display
Fee quotes with USD values (#949) carried over as-is.

### Transfer Form Restructured
`TransferTokenForm` significantly restructured around tokenKey model — two `TokenSelectField`s (origin/dest) each open the unified modal. No separate `ChainSelectField`. Core transfer execution logic (fees, validation, `useTokenTransfer`) is similar.

### Removed
- `FloatingButtonStrip`, `ChainSelectField`, `ChainSelectModal`, `TokenListModal`
- Config flags: `showAddRouteButton`, `showAddChainButton`, `showDisabledTokens`
- Consolidated `tokens/balances.ts` → `balances/hooks.ts`

---
**61 files changed** — +3,796 / -1,461

## Summary by CodeRabbit

* **New Features**
  * Unified token & chain selection modal/panels, chain edit UI, mobile quick-select, wallet dropdown, recipient address modal, token import button, new icons/components, token list/panel and transfer section.

* **Bug Fixes**
  * Improved recipient validation/clearing, wallet-connect submit behavior, route-aware fee and max calculations, more reliable route detection and balance handling.

* **UI Updates**
  * Banner repositioning and rounded styling, TipCard/layout and modal header refinements, updated token/transfer layouts and compact details.

* **Chores**
  * Dependency bump, ESLint rule adjusted, .gitignore updated, removed legacy wallet adapters/components.

* **Tests**
  * Added chain filter/sort and token utilities tests; updated fee-related tests.

### DIFF
--- a/src/features/tokens/utils.ts
+++ b/src/features/tokens/utils.ts
@@ -133,17 +133,18 @@ export function getTokensWithSameCollateralAddresses(
         ? normalizeAddress(destinationToken.collateralAddressOrDenom, destinationToken.protocol)
         : null;
 
-      // For HypNative tokens if both addresses are null then it matches, otherwise check with eqAddress
+      // For HypNative tokens both collateral addresses are null — match by symbol
+      // to avoid treating different native deployments (e.g. ETH vs ETHSTAGE) as interchangeable
       const originMatches =
         isNullish(originCollateralAddress) && isNullish(currentOriginCollateralAddress)
-          ? true
+          ? originToken.symbol === origin.symbol
           : originCollateralAddress && currentOriginCollateralAddress
             ? eqAddress(originCollateralAddress, currentOriginCollateralAddress)
             : false;
 
       const destinationMatches =
         isNullish(destinationCollateralAddress) && isNullish(currentDestinationCollateralAddress)
-          ? true
+          ? destinationToken.symbol === destination.symbol
           : destinationCollateralAddress && currentDestinationCollateralAddress
             ? eqAddress(destinationCollateralAddress, currentDestinationCollateralAddress)
             : false;
--- a/README.md
+++ b/README.md
@@ -46,6 +46,9 @@ NEXT_PUBLIC_WALLET_CONNECT_ID=<projectId> pnpm run dev
 ### Test
 
 ```sh
+# Run tests
+pnpm test
+
 # Lint check code
 pnpm run lint
 


src/features/tokens/utils.ts
```diff
@@ -132,40 +136,202 @@ export function getTokensWithSameCollateralAddresses(
...
+    // Already seen this collateral, skip it
+    return false;
+  });
+}
+
+/**
+ * Build a unified tokens array containing all tokens that can participate in transfers
+ * (either as origin or destination). Deduplicates by address and by collateral.
+ */
+export function buildTokensArray(warpCoreTokens: Token[]): Token[] {
```
paulbalaji, line 206: ### 🔴 Critical: Disabled-chain tokens not filtered out

`buildTokensArray` builds from all `warpCore.tokens` with no disabled-chain filter. `useChainInfos()` marks chains as `disabled: true` (via `ChainStatus.Disabled`), but that flag only grays out chains in `ChainList` — it's never enforced on the token side.

With `config.shouldDisableChains = true`, users can still see, select, and bridge tokens on disabled chains because `TokenList` consumes `useTokens()` (the unfiltered store array).

**Suggested fix:** Filter out disabled-chain tokens here or in store initialization:
```ts
const tokens = buildTokensArray(
  warpCore.tokens.filter(t => !isChainDisabled(multiProvider, t.chainName))
);
```
Or filter in `TokenList` using the disabled chain set from `useChainInfos()`.

src/features/transfer/useFeeQuotes.ts
```diff
@@ -64,31 +74,30 @@ export function useFeeQuotes(
...
-        warpCore,
-        originToken,
-        destinationToken,
-        amountWei,
-        recipient,
-        sender,
-        defaultMultiCollateralRoutes,
-      );
-    }
+    transferToken = await getTransferToken(
```
paulbalaji, line 94: ### 🔴 Critical: Fee quote can use different multi-collateral route than execution

`getTransferToken()` is called here **without** `defaultMultiCollateralRoutes`, but the validation path (`TransferTokenForm.tsx:941`) and max-amount path (`maxAmount.ts:65`) **do** pass it. This means the fee preview could select a different collateral route than the one actually used for the transfer.

**Fix:** Pass `defaultMultiCollateralRoutes` here too:
```ts
import { defaultMultiCollateralRoutes } from '../../consts/defaultMultiCollateralRoutes';
// ...
transferToken = await getTransferToken(
  warpCore,
  originToken,
  destinationToken,
  amountWei,
  recipient,
  sender,
  defaultMultiCollateralRoutes, // missing
);
```

src/features/wallet/context/SolanaWalletContext.tsx
```diff
@@ -4,9 +4,7 @@ import { ConnectionProvider, WalletProvider } from '@solana/wallet-adapter-react
 import { WalletModalProvider } from '@solana/wallet-adapter-react-ui';
 import '@solana/wallet-adapter-react-ui/styles.css';
 import {
-  BackpackWalletAdapter,
   LedgerWalletAdapter,
```
paulbalaji, line 7: ### 🔴 Critical: Solana wallet regression — Phantom and Backpack removed

`PhantomWalletAdapter` and `BackpackWalletAdapter` were present on `main` but are removed here. Phantom is the most popular Solana wallet.

If intentional (e.g. Solana Wallet Standard auto-detection makes explicit adapters unnecessary), a code comment explaining the rationale would help. Otherwise, restore them.

src/features/wallet/RecipientAddressModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export function RecipientAddressModal({
+  isOpen,
+  close,
+  onSave,
+  initialValue = '',
+  protocol = ProtocolType.Ethereum,
+}: RecipientAddressModalProps) {
+  const [address, setAddress] = useState(initialValue);
```
paulbalaji, line 21: ### 🟡 Important: State not synced when modal re-opens with different `initialValue`

`useState(initialValue)` only sets state on first mount. If the modal is closed and re-opened with a different `initialValue` (e.g. user pastes address, closes, switches chain, re-opens), the input still shows the stale value.

**Fix:**
```ts
useEffect(() => {
  if (isOpen) {
    setAddress(initialValue);
    setError('');
  }
}, [isOpen, initialValue]);
```

src/features/tokens/utils.ts
```diff
@@ -132,40 +136,202 @@ export function getTokensWithSameCollateralAddresses(
...
+      ? normalizeAddress(t.collateralAddressOrDenom, t.protocol)
+      : null;
+    // Match by collateral address if both have one, otherwise match by symbol
+    if (normalizedOriginCollateral && normalizedRouteCollateral) {
+      return normalizedOriginCollateral === normalizedRouteCollateral;
+    }
+    return t.symbol === originToken.symbol;
+  });
+
+  return matchingToken || routeTokens[0];
```
paulbalaji: ### 🟡 Important: `findRouteToken` silently returns arbitrary token on no match

Falling back to `routeTokens[0]` when no collateral/symbol match is found could silently pick the wrong multi-collateral route. Consider returning `undefined` so the caller can handle the error explicitly — this matches the function's contract of "find the token that has a route."

```ts
return matchingToken; // let caller handle undefined
```

src/features/tokens/TokenList.tsx
```diff
@@ -0,0 +1,364 @@
+import { ChainName, Token } from '@hyperlane-xyz/sdk';
+import { Tooltip, useDebounce } from '@hyperlane-xyz/widgets';
+import React, { useEffect, useMemo, useRef, useState, useTransition } from 'react';
+import { config } from '../../consts/config';
+import { useTokenBalances } from '../balances/hooks';
+import { tokenKey } from '../balances/tokens';
```
paulbalaji: ### 🟡 Important: Two `tokenKey` functions with incompatible formats

This file imports both `tokenKey` (from `balances/tokens.ts`, format: `chainName-address-symbol`) and `getTokenKey` (from `tokens/utils.ts`, format: `chainName-symbol-address`, all lowercased). The field order and casing differ.

Currently works because each key is used for its own lookup map. But having two near-identical functions with different formats is a maintenance hazard — if anyone changes one without the other, balance lookups silently fail.

Consider unifying into a single canonical `tokenKey` function, or at minimum adding prominent JSDoc comments on both warning they must stay in sync.

src/features/chains/ChainFilterPanel.tsx
```diff
@@ -0,0 +1,313 @@
+import { ChainName } from '@hyperlane-xyz/sdk';
```
paulbalaji, line 1: ### 🟡 Important: Filter/sort dropdowns lack keyboard accessibility

The `FilterButton` and `SortButton` dropdowns use custom div-based menus without `role="menu"`/`role="menuitem"`, arrow key navigation, or Escape-to-close. The `useClickOutside` hook only listens for `pointerdown`, so Tab-away leaves dropdowns open.

Consider adding ARIA roles, Escape key handler, and arrow key navigation — or use an existing accessible dropdown component.

src/features/tokens/utils.ts
```diff
@@ -132,40 +136,202 @@ export function getTokensWithSameCollateralAddresses(
...
+  });
+}
+
+/**
+ * Find the actual warpCore token that has a route to the destination.
+ * The passed originToken may be from a deduplicated array and may not have
+ * the connection, but another token with the same collateral in the warpCore might
+ * have this token.
+ */
+export function findRouteToken(
```
paulbalaji, line 329: ### 🟡 Important: Missing test coverage for critical new functions

`checkTokenHasRoute` and `findRouteToken` are high-risk new functions with complex logic (collateral matching, fallback chain) but have zero direct test coverage. `getTokenKey`/`getCollateralKey` are also untested despite caching behavior.

The existing tests for `buildTokensArray` and `dedupeTokensByCollateral` are solid — extending that pattern to these functions would significantly reduce regression risk.

src/features/tokens/hooks.ts
```diff
@@ -1,39 +1,140 @@
...
 import { useStore } from '../store';
+import { getTokenKey } from './utils';
 
 export function useWarpCore() {
   return useStore((s) => s.warpCore);
 }
+/**
+ * Find a token by its key from a WarpCore or Token array
+ */
+export function getTokenByKey(tokens: Token[], key: string | undefined): Token | undefined {
```
paulbalaji, line 18: ### 🟢 Suggestion: O(n) scan despite O(1) map available

`getTokenByKey()` does a linear `.find()` but the store already has a pre-computed `tokenByKeyMap` for O(1) lookups (via `useTokenByKeyMap()` at line 124). The O(1) version (`getTokenByKeyFromMap` at line 132) exists but appears unused in practice. With 700+ tokens and 20+ call sites, consider switching to the map-based lookup or removing the unused map to reduce confusion.

src/consts/config.ts
```diff
@@ -67,4 +62,24 @@ export const config: Config = Object.freeze({
   shouldDisableChains: false,
   rpcOverrides,
   enableTrackingEvents: false,
+  featuredTokens: [
```
paulbalaji, line 65: ### 🟢 Suggestion: Hardcoded `featuredTokens` limits deployer customization

This list and `preferredChains` (hardcoded in `TokenListPanel.tsx`) should ideally be configurable per deployment — deployers may have entirely different chains/tokens. Consider making these env-var driven or derivable from warp route config.

src/features/tokens/TokenList.tsx
```diff
@@ -0,0 +1,364 @@
...
+  recipient,
+}: TokenListProps) {
+  const multiProvider = useMultiProvider();
+  const allTokens = useTokens();
+  const collateralGroups = useCollateralGroups();
+  const debouncedSearch = useDebounce(searchQuery, 300);
+  const scrollRef = useRef<HTMLDivElement>(null);
+
+  // Deferred state for route map - allows UI to render immediately
+  const [tokenRouteMap, setTokenRouteMap] = useState<Map<string, boolean> | null>(null);
```
paulbalaji, line 68: ### 🟢 Suggestion: Route map `null` treated as "all routes valid" during computation

During `useTransition` computation, `tokenRouteMap` is `null` and multiple places treat this as "token is routable" (`tokenRouteMap ? ... : true`). This means tokens briefly appear routable before the map is computed.

Not a correctness blocker — validation still catches unroutable selections on submit — but a shimmer/skeleton state would be cleaner UX.

eslint.config.mjs
```diff
@@ -69,6 +69,9 @@ export default [
       'no-multiple-empty-lines': ['error'],
       'jsx-a11y/alt-text': ['off'],
 
+      // React Compiler rule - downgrade to warning for valid useEffect patterns
```
paulbalaji, line 72: ### 🟢 Suggestion: Prefer per-line ESLint disables over project-wide downgrade

Downgrading `react-hooks/set-state-in-effect` to `warn` project-wide is reasonable for the legitimate async-setState-in-effect patterns here, but could mask genuinely problematic new code. Consider `// eslint-disable-next-line` on the specific lines instead.

src/features/tokens/utils.ts
Xaroz: Fixed in 3e00b05d — added `useDisabledChains()` hook and filtering in `TokenList` so disabled-chain tokens are excluded from search/display.

src/features/transfer/useFeeQuotes.ts
Xaroz: Fixed in 3e00b05d — added `defaultMultiCollateralRoutes` import and passed it to `getTransferToken()` in `fetchFeeQuotes`.

src/features/wallet/context/SolanaWalletContext.tsx
Xaroz: Intentional — Solana Wallet Standard auto-detection handles Phantom and Backpack now, explicit adapters are no longer needed.

src/features/wallet/RecipientAddressModal.tsx
Xaroz: `handleClose` already resets to `initialValue` on close. The stale state only occurs if `initialValue` changes between close and reopen without user interaction — which only happens on protocol change (clears recipient). Very narrow edge case, skipping for now.

src/features/tokens/utils.ts
Xaroz: The `routeTokens[0]` fallback only fires when `getTokensForRoute` found valid routes but none matched by collateral or symbol — a very unlikely state. All callers already handle `undefined` from the `routeTokens.length === 0` path. Keeping as-is since the fallback picks from a valid route set.

src/features/tokens/TokenList.tsx
Xaroz: By design — `tokenKey` (balances) and `getTokenKey` (token identity) serve different purposes and each is used for its own lookup map. Worth a comment, not a refactor now.

src/features/chains/ChainFilterPanel.tsx
Xaroz: Noted — not in scope for this PR.

src/features/tokens/utils.ts
Xaroz: Done — added 10 tests for `checkTokenHasRoute` and `findRouteToken` covering edge cases (no connection, mismatched collateral keys, empty groups, symbol matching, fallback behavior) in #983.

src/features/tokens/hooks.ts
Xaroz: Done — replaced all `getTokenByKey(tokens, ...)` calls with O(1) `getTokenByKeyFromMap(tokenMap, ...)` across `TokenSelectField`, `TransferTokenForm`, and `RecipientConfirmationModal` in #983.

src/consts/config.ts
Xaroz: Agree — should eventually be env-var driven or derivable from warp route config. Fine for now as a starting point.

src/features/tokens/TokenList.tsx
Xaroz: Intentional — treating as routable during the `useTransition` computation avoids a flash of "route unavailable" badges on first render. Validation still catches unroutable selections on submit.

eslint.config.mjs
Xaroz: Reasonable suggestion — will consider switching to per-line disables once the patterns stabilize.

src/features/chains/ChainEditModal.tsx
```diff
@@ -0,0 +1,50 @@
+import { ChainMetadata } from '@hyperlane-xyz/sdk';
+import { ChainDetailsMenu, Modal } from '@hyperlane-xyz/widgets';
+import { useCallback } from 'react';
+import { ModalHeader } from '../../components/layout/ModalHeader';
+import { useStore } from '../store';
+
+interface Props {
+  isOpen: boolean;
+  close: () => void;
+  chainName: ChainName;
```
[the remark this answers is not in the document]
Xaroz: `ChainName` is declared globally in `src/global.d.ts`, no import needed.

src/features/chains/chainFilterSort.ts
```diff
@@ -0,0 +1,102 @@
...
+      // Sort options
+      .sort((c1, c2) => {
+        // Disabled chains always at the bottom
+        if (c1.disabled && !c2.disabled) return 1;
+        if (!c1.disabled && c2.disabled) return -1;
+
+        if (sort.sortBy === ChainSortBy.ChainId) {
+          const result = c1.chainId.toString().localeCompare(c2.chainId.toString());
+          return sort.sortOrder === SortOrder.Asc ? result : -result;
+        }
```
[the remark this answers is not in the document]
Xaroz: Fixed in 3e00b05d — using `localeCompare` with `{ numeric: true }` so numeric IDs sort correctly while string IDs (e.g. Cosmos) still sort sensibly.

src/features/chains/MobileChainQuickSelect.tsx
```diff
@@ -0,0 +1,98 @@
...
+      // Get preferred chains that exist, maintaining preferred order
+      const preferred = preferredChains
+        .filter((name) => chainNameSet.has(name))
+        .map((name) => allChains.find((c) => c.name === name))
+        .filter((c): c is ChainInfo => !!c);
+
+      // Fill remaining slots with non-preferred, non-disabled chains
+      const remaining = allChains.filter((c) => !preferredSet.has(c.name) && !c.disabled);
+      const slotsToFill = DEFAULT_MAX_VISIBLE_CHAINS - preferred.length;
+      const fillers = remaining.slice(0, Math.max(0, slotsToFill));
```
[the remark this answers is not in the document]
Xaroz: Fixed in 3e00b05d — disabled chains now filtered from both the preferred and default paths.

src/features/tokens/TokenSelectField.tsx
```diff
@@ -94,32 +148,42 @@ function TokenButton({
...
   );
 }
 
 const styles = {
-  base: 'mt-1.5 w-full px-2.5 py-2.5 flex items-center justify-between text-sm rounded-lg border border-primary-300 outline-none transition-colors duration-500',
-  enabled: 'hover:bg-gray-100 active:scale-95 focus:border-primary-500',
-  disabled: 'bg-gray-100 cursor-default',
+  base: 'w-full py-2 flex items-center justify-between transition-all rounded-xl px-1.5 border duration-150 border-gray-400/25 shadow-sm group',
+  enabled: 'hover:bg-gray-50 cursor pointer',
+  disabled: 'cursor-not-allowed opacity-60',
```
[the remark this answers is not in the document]
Xaroz: Fixed in 3e00b05d.

src/features/tokens/TokenList.tsx
```diff
@@ -0,0 +1,364 @@
...
+  // No filter     → defaultTokens, capped at 50 with routable prioritized
+  const balanceTokens = useMemo(() => {
+    const q = debouncedSearch?.trim().toLowerCase();
+
+    if (q || chainFilter) {
+      return allTokens.filter((t) => {
+        const chainMatch = chainFilter && t.chainName === chainFilter;
+        const searchMatch = q && matchesSearch(t, q, multiProvider);
+        return chainMatch || searchMatch;
+      });
```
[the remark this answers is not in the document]
Xaroz: Intentional — this is the balance fetch scope, not the display filter. We want balances for anything matching either filter so they're ready when the display narrows.

src/features/chains/ChainFilterPanel.tsx
```diff
@@ -0,0 +1,313 @@
...
+          >
+            <ChevronIcon direction="w" width={14} height={14} />
+          </button>
+        )}
+        <SearchInput
+          value={searchQuery}
+          onChange={onSearchChange}
+          placeholder="Search Chains"
+          aria-label="Search chains"
+        />
```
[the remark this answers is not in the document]
Xaroz: Correction: `SearchInput` does not accept rest props — it only takes `inputRef`, `value`, `onChange`, `placeholder`. Fixed in 3739c35e by adding an optional `aria-label` prop to `SearchInput` and forwarding it to `TextInput`.

eslint.config.mjs
```diff
@@ -69,6 +69,9 @@ export default [
...
+      'react-hooks/set-state-in-effect': ['warn'],
```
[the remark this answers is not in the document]
Xaroz: Keeping as `warn` — this was intentionally downgraded because the rule is overly strict for valid `useEffect` patterns in the codebase.

src/components/input/TextField.tsx
```diff
@@ -30,4 +30,4 @@ export const TextInput = forwardRef(function _TextInput(
 });
 
 const defaultClassName =
-  'mt-1.5 px-2.5 py-2.5 text-sm rounded-lg border border-primary-300 focus:border-primary-500 disabled:bg-gray-150 outline-none transition-all duration-300';
+  'rounded-lg border border-primary-300 focus:border-primary-500 disabled:bg-gray-150 outline-none transition-all duration-300';
```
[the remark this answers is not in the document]
Xaroz: Already verified — spacing looks correct across all usages.

src/features/tokens/UnifiedTokenChainModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+    [onSelect, onClose],
+  );
+
+  const handleSelectChain = (chain: ChainInfo | null) => {
+    setSelectedChain((prev) => {
+      if (prev?.name === chain?.name) return prev;
+      trackChainSelectionEvent(selectionMode, chain, prev);
+      return chain;
+    });
+  };
```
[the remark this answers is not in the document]
Xaroz: Fixed in 3739c35e — moved analytics tracking outside the `setState` callback. Now reads `selectedChain` directly, early-returns if same chain, tracks, then sets state.

src/features/tokens/UnifiedTokenChainModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+  // Mobile: when selecting a chain from the full list, go back to tokens
+  const handleSelectChainMobile = (chain: ChainInfo | null) => {
+    handleSelectChain(chain);
+    setShowMobileChainList(false);
+  };
+
+  const handleEditChain = (chainName: string) => {
+    close();
+    onEditChain?.(chainName);
+  };
```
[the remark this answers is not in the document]
Xaroz: Intentional — when the user edits a chain, they return to the modal via `handleEditBack`. Preserving search/filter state gives continuity since they're coming right back.

src/features/transfer/fees.ts
```diff
@@ -107,16 +108,31 @@ export async function getTransferToken(
...
+  // Find the actual warpCore token that has the route
+  // Because we deduplicated tokens with the same collateral, the current token pair
+  // might not be correct, so it is necessary to get the correct token pair.
+  // Make sure to have used `checkTokenHasRoute` before calling getLowestFeeTransferToken
+  // as that will validate that the token pair actually refer to the same asset
+  const originRouteToken = findRouteToken(warpCore, originToken, destinationChain);
+  if (!originRouteToken) {
+    // No route exists, return original token (validation will catch this)
+    return originToken;
+  }
```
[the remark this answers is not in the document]
Xaroz: The fallback to `originToken` is reasonable here — validation catches unroutable selections on submit.

src/features/transfer/maxAmount.ts
```diff
@@ -31,21 +33,34 @@ export function useFetchMaxAmount() {
...
+      accounts,
+    );
+    const recipient = formRecipient || connectedDestAddress || address;
+
+    // Find the actual warpCore token that has the route (handles deduplicated tokens)
+    const originRouteToken = findRouteToken(warpCore, originToken, destination);
+    if (!originRouteToken) return undefined;
+
+    const destinationToken = originRouteToken.getConnectionForChain(destination)?.token;
     if (!destinationToken) return undefined;
```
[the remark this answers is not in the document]
Xaroz: Returning `undefined` is handled gracefully by the caller (max button no-ops). Throwing would trigger a toast with a misleading message about native balance being insufficient when the real issue is a missing route.

src/features/transfer/TransfersDetailsModal.tsx
```diff
@@ -117,102 +118,105 @@ export function TransfersDetailsModal({
...
-        </div>
-      )}
+            {showSignWarning && (
+              <div className="mt-3 text-center text-sm text-gray-600">
+                If your wallet does not show a transaction request or never confirms, please try the
+                transfer again.
+              </div>
+            )}
+          </div>
+        )}
```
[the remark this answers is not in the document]
Xaroz: Fixed in 3739c35e — removed the dead ternary, simplified to `text-gray-600`.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -102,19 +105,36 @@ async function executeTransfer({
...
+
+    // Get effective recipient (form value or fallback to connected wallet for destination)
+    const connectedDestAddress = getAccountAddressForChain(
+      multiProvider,
+      destinationToken.chainName,
+      activeAccounts.accounts,
+    );
+    const recipient = formRecipient || connectedDestAddress || '';
+    if (!recipient) throw new Error('No recipient address available');
+    // Find the actual connected token from the origin and destination chains
```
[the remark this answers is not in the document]
Xaroz: The SDK already normalizes addresses internally during transfer execution, so normalizing here would be redundant.

src/features/transfer/utils.ts
```diff
@@ -180,3 +180,14 @@ export async function isSmartContract(
...
+
+// Returns if the recipient should be cleared by checking if it is valid address from the current chain protocol
+export function shouldClearAddress(
+  multiProvider: MultiProtocolProvider,
+  recipient: string,
+  chainName: string,
+) {
+  const protocol = multiProvider.tryGetProtocol(chainName);
+  if (recipient && protocol && !isValidAddress(recipient, protocol)) return true;
+  return false;
```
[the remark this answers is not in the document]
Xaroz: `tryGetProtocol` is intentionally a "try" method — returning `false` (don't clear) is the safe default for unknown chains. Throwing would crash the form.

src/features/wallet/RecipientAddressModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+    if (!trimmedAddress) return;
+
+    if (!isValidAddress(trimmedAddress, protocol)) {
+      setError('Invalid address');
+      return;
+    }
+
+    setError('');
+    onSave(trimmedAddress);
+    close();
```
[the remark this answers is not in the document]
Xaroz: SDK normalizes addresses during transfer execution, so normalizing in form state is redundant.

src/features/wallet/RecipientAddressModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+    <Modal isOpen={isOpen} close={handleClose} panelClassname="max-w-sm p-0 overflow-hidden">
+      <div className="flex items-center justify-between px-4 py-3">
+        <h2 className="text-lg font-medium text-gray-900">Receive Address</h2>
+        <button
+          type="button"
+          onClick={handleClose}
+          className="text-gray-400 transition-colors hover:text-gray-600"
+        >
+          <XIcon width={12} height={12} />
+        </button>
```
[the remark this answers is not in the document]
Xaroz: Fixed in 3739c35e — added `aria-label="Close"` to the button.

src/styles/globals.css
```diff
@@ -262,3 +262,53 @@ Needs to be higher than Intercom chatbox (z-index: 2147483000)
...
+  border-color: #d1d5db;
+  border-radius: 4px;
+  font-size: 0.8125rem;
+}
+
+.chain-edit-container input[type='text']:focus,
+.chain-edit-container textarea:focus {
+  border-color: var(--color-primary-500, #9a0dff);
+  outline: none;
+}
```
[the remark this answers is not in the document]
Xaroz: The border-color change provides sufficient visual feedback for now. Will revisit in a broader accessibility pass.

src/features/tokens/hooks.ts
```diff
@@ -1,39 +1,151 @@
...
+}
 
-export function useTokens() {
+// Helper to find token by chainName-symbol format
+function findTokenByChainSymbol(tokens: Token[], chainSymbol: string): Token | undefined {
+  const [chainName, symbol] = chainSymbol.split('-');
+  if (!chainName || !symbol) return undefined;
+  return tokens.find(
+    (t) => t.chainName === chainName && t.symbol.toLowerCase() === symbol.toLowerCase(),
+  );
```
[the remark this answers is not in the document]
Xaroz: Checked the registry — no hyphenated chain names exist. The format is always `chainName-symbol`.

src/features/tokens/hooks.ts
```diff
@@ -1,39 +1,151 @@
...
+
+  // Try to find origin token from URL params (chain + symbol)
+  let originToken: Token | undefined;
+  if (originChainQuery && originTokenSymbol) {
+    originToken = tokens.find(
+      (t) =>
+        t.chainName === originChainQuery &&
+        t.symbol.toLowerCase() === originTokenSymbol.toLowerCase(),
+    );
+  }
```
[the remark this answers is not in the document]
Xaroz: URL params use `chain + symbol` for readability. If multiple tokens share a symbol on the same chain, downstream logic picks the correct route.

src/features/tokens/hooks.ts
```diff
@@ -1,39 +1,151 @@
...
+
+  // Last resort: first connection from origin token
+  if (!destinationToken && originToken) {
+    const firstConnection = originToken.connections?.[0];
+    const connectedChain = firstConnection?.token?.chainName;
+    const connectedSymbol = firstConnection?.token?.symbol;
+    destinationToken = connectedChain
+      ? tokens.find((t) => t.chainName === connectedChain && t.symbol === connectedSymbol)
+      : undefined;
```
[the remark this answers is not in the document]
Xaroz: Fixed in 3739c35e — now uses case-insensitive comparison.

src/features/chains/chainFilterSort.ts
```diff
@@ -0,0 +1,104 @@
...
+  return (
+    data
+      // Query search
+      .filter(
+        (chain) =>
+          !q ||
+          chain.name.toLowerCase().includes(q) ||
+          chain.displayName.toLowerCase().includes(q) ||
+          chain.chainId.toString().includes(q),
+      )
```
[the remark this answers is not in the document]
Xaroz: Fixed in 3739c35e — added `.toLowerCase()` to chainId query matching.

src/features/tokens/TokenList.tsx
```diff
@@ -0,0 +1,372 @@
...
+
+    // No filter: cap at 50, routable first
+    const maxDefault = 50;
+    if (defaultTokens.length <= maxDefault) return defaultTokens;
+
+    const routable: Token[] = [];
+    const rest: Token[] = [];
+    for (const t of defaultTokens) {
+      const hasRoute = tokenRouteMap ? (tokenRouteMap.get(getTokenKey(t)) ?? true) : true;
+      (hasRoute ? routable : rest).push(t);
```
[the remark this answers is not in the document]
Xaroz: Intentional — defaulting to `true` during `useTransition` computation avoids a flash of "route unavailable" badges on first render. Validation catches unroutable selections on submit.

src/features/tokens/TokenSelectField.tsx
```diff
@@ -1,91 +1,145 @@
...
-        onClick={onClickField}
-        isAutomatic={isAutomaticSelection}
-      />
-      <TokenListModal
+      <div className="flex flex-col">
+        {showLabel && label && (
+          <label htmlFor={name} className="mb-1 pl-0.5 text-sm text-gray-600">
+            {label}
+          </label>
+        )}
```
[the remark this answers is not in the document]
Xaroz: Fixed in 3739c35e — replaced `<label htmlFor>` with `<span>` since there's no matching control id.

src/features/transfer/useFeeQuotes.ts
```diff
@@ -28,6 +30,14 @@ export function useFeeQuotes(
...
   );
 
+  // Get effective recipient (form value or fallback to connected wallet for destination)
+  const { address: connectedDestAddress } = getAccountAddressAndPubKey(
+    multiProvider,
+    destinationToken?.chainName,
+    accounts,
+  );
+  const recipient = formRecipient || connectedDestAddress || '';
+
```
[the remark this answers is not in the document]
Xaroz: SDK normalizes addresses internally during transfer and fee estimation.

src/features/chains/ChainFilterPanel.tsx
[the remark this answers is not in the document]
Xaroz: fixed

src/features/tokens/utils.test.ts
```diff
@@ -332,3 +346,580 @@ describe('tryGetDefaultOriginToken', () => {
...
+    expect(result).toContain(token1);
+    expect(result).toContain(token2);
+  });
+
+  test('should keep collateralized tokens with different collateral addresses', () => {
+    const token1 = createMockToken({
+      standard: TokenStandard.EvmHypCollateral,
+      addressOrDenom: '0x1111111111111111111111111111111111111111',
+      collateralAddressOrDenom: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+    });
```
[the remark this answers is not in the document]
Xaroz: this is a test it doesn't matter

src/features/analytics/refiner.ts
```diff
@@ -40,5 +40,5 @@ export function refinerIdentifyAndShowTransferForm(params: {
     protocol: params.protocol,
     chain: params.chain,
   });
-  refiner('showForm', REFINER_RESEARCH_FORM_ID);
+  refiner('showForm', REFINER_TRANSFER_FORM_ID);
```
paulbalaji: ### 🔴 Critical: Successful transfer can be marked Failed if Refiner throws

`refinerIdentifyAndShowTransferForm` is called inside the same `try` block as the transfer confirmation in `useTokenTransfer.ts:251`. If either `refiner('identifyUser', ...)` or `refiner('showForm', ...)` throws (e.g., Refiner CDN failed to load, SDK error), the exception propagates to the `catch` block which calls `updateTransferStatus(transferIndex, TransferStatus.Failed)` — even though the on-chain transfer succeeded.

`initRefiner` (line 17) correctly wraps its calls in try/catch. This function should do the same:

```ts
export function refinerIdentifyAndShowTransferForm(params: { ... }): void {
  if (!config.enableTrackingEvents || !refiner || !REFINER_TRANSFER_FORM_ID) return;
  try {
    refiner('identifyUser', { ... });
    refiner('showForm', REFINER_TRANSFER_FORM_ID);
  } catch (error) {
    logger.warn('Failed to show Refiner form:', error);
  }
}
```

src/features/tokens/utils.ts
```diff
@@ -132,40 +152,200 @@ export function getTokensWithSameCollateralAddresses(
...
+  const routeTokens = warpCore.getTokensForRoute(originToken.chainName, destinationChain);
+  if (routeTokens.length === 0) return undefined;
+
+  const originCollateralKey = getCollateralKey(originToken);
+
+  // Find a route token that shares the same collateral key (resolved-underlying-aware)
+  const matchingToken = routeTokens.find((t) => getCollateralKey(t) === originCollateralKey);
+
+  // Fallback: match by symbol on same chain
+  return matchingToken || routeTokens.find((t) => t.symbol === originToken.symbol);
```
paulbalaji, line 349: ### 🟡 Important: Symbol fallback can select wrong multi-collateral route

When the collateral key match fails, this falls back to matching by `symbol` alone. If two tokens share the same symbol but have different collateral on the same chain pair (e.g., USDC.e vs native USDC bridged separately), this picks whichever comes first — potentially the wrong route.

The collateral key match above handles the common case well. But the symbol fallback should at minimum log a warning so it's observable in production:

```ts
const symbolFallback = routeTokens.find((t) => t.symbol === originToken.symbol);
if (!matchingToken && symbolFallback) {
  logger.warn('findRouteToken: collateral key miss, falling back to symbol match', {
    origin: originToken.symbol,
    chain: destinationChain,
  });
}
return matchingToken || symbolFallback;
```

Alternatively, return `undefined` when collateral key doesn't match and let callers handle it explicitly.

src/features/wallet/RecipientAddressModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export function RecipientAddressModal({
+  isOpen,
+  close,
+  onSave,
+  initialValue = '',
+  protocol = ProtocolType.Ethereum,
+}: RecipientAddressModalProps) {
+  const [address, setAddress] = useState(initialValue);
```
paulbalaji, line 21: ### 🟡 Important: State not synced when modal reopens with different `initialValue`

`useState(initialValue)` only sets state on first mount. If the modal stays mounted and `initialValue` changes between opens (e.g., user switches chain/protocol and the connected wallet address changes), the input shows the stale previous address.

`handleClose` resets to `initialValue` on close, but that doesn't help when `initialValue` itself changes between close and reopen.

Fix: add a sync effect, or key the component on `initialValue` to force remount:

```ts
useEffect(() => {
  setAddress(initialValue);
  setError('');
}, [initialValue]);
```

src/features/tokens/SelectOrInputTokenIds.tsx
```diff
@@ -17,13 +12,13 @@ export function SelectOrInputTokenIds({ disabled }: { disabled: boolean }) {
   const isContractAllowToGetTokenIds = true;
 
   return isContractAllowToGetTokenIds ? (
```
paulbalaji, line 14: ### 🟡 Important: Hardcoded bypass makes manual NFT ID input unreachable

`isContractAllowToGetTokenIds = true` with the contract check commented out (lines 5-6) means `InputTokenId` is dead code. For NFT contracts that don't support `tokenByOwner` enumeration, users will see an empty dropdown with no way to manually enter a token ID.

If this is intentional WIP, consider adding a `// TODO` with a tracking issue. If the contract check is planned for removal, delete the `InputTokenId` branch entirely.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -102,19 +105,36 @@ async function executeTransfer({
...
-  const { origin, destination, tokenIndex, amount, recipient } = values;
+  const { originTokenKey, destinationTokenKey, amount, recipient: formRecipient } = values;
   const multiProvider = warpCore.multiProvider;
 
   try {
-    const originToken = getTokenByIndex(warpCore, tokenIndex);
-    const connection = originToken?.getConnectionForChain(destination);
-    if (!originToken || !connection) throw new Error('No token route found between chains');
+    const originToken = routeOverrideToken || getTokenByKey(warpCore.tokens, originTokenKey);
+    const destinationToken = getTokenByKey(warpCore.tokens, destinationTokenKey);
```
paulbalaji, line 120: ### 🟢 Minor: Addresses not normalized before transfer execution

`recipient` (line 129) and `sender` (computed later from `getAccountAddressForChain`) pass through to `warpCore.getTransferRemoteTxs()` without normalization. The validation path in `TransferTokenForm.tsx` does normalize (via `normalizeAddress`), but the execution path doesn't.

If the SDK normalizes internally this is fine, but it's inconsistent with validation. Worth verifying or adding `normalizeAddress()` calls here for EVM addresses.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -801,85 +833,104 @@ function ReviewDetails({
...
-      recipient: '',
-    };
-  }, [warpCore, destinationQuery, originQuery, tokenIndex, defaultOriginToken]);
+  return useMemo(() => {
+    if (!originToken || !destinationToken) return true;
+    return checkTokenHasRoute(originToken, destinationToken, collateralGroups);
+  }, [originToken, destinationToken, collateralGroups]);
 }
 
 const insufficientFundsErrMsg = /insufficient.[funds|lamports]/i;
```
paulbalaji: ### 🟢 Minor: Regex uses character class instead of alternation

`[funds|lamports]` is a character class matching any single char from `f,u,n,d,s,|,l,a,m,p,o,r,t`. The intent is alternation. This works by accident (`.` matches the space, then `f` or `l` matches), but should be:

```ts
const insufficientFundsErrMsg = /insufficient.(funds|lamports)/i;
```

src/features/tokens/utils.ts
Xaroz: I don't think it would pick the wrong chain pair as it tries to match using chainName-symbol-collateralAddress, for your example: USDC vs USDC.e if they have the same symbol + collateral address in the same chain they are essentially the same token

src/features/analytics/refiner.ts
Xaroz: noted

src/features/wallet/RecipientAddressModal.tsx
Xaroz: I think we discussed this before, but when the Modal closes inside RecipientAddressModal it will unmount and will reset the stale state when `handleClose` happens

src/features/tokens/SelectOrInputTokenIds.tsx
Xaroz: valid gonna add a TODO here

src/features/transfer/useTokenTransfer.ts
Xaroz: the SDK already does the normalizing it would just add extra logic to the user for no reason

src/features/transfer/TransferTokenForm.tsx
Xaroz: interesting that it works by accident lmao, I changed it so that it is correct now

src/features/tokens/utils.ts
```diff
@@ -7,12 +8,28 @@ import {
...
+// Module-level caches for expensive key computations
+// WeakMap allows automatic garbage collection when token objects are no longer referenced
+const tokenKeyCache = new WeakMap<IToken, string>();
+let collateralKeyCache = new WeakMap<IToken, string>();
+
+// Resolved underlying addresses for lockbox/vault tokens.
+// Set once during initWarpContext via setResolvedUnderlyingMap().
+// getCollateralKey() uses this to group lockbox/vault tokens with their
+// non-wrapper counterparts (e.g., lockbox USDT grouped with regular USDT).
+let resolvedUnderlyingMap: Map<string, string> = new Map();
```
xeno097, line 22: **Design concern: module-level mutable state.** `resolvedUnderlyingMap` and `collateralKeyCache` are module-level `let` variables mutated via `setResolvedUnderlyingMap()`. This creates hidden initialization-order coupling — if `getCollateralKey()` is called before `setResolvedUnderlyingMap()`, it silently uses an empty map and caches wrong results.

This is safe today because `initWarpContext` calls `setResolvedUnderlyingMap` before `buildTokensArray`/`groupTokensByCollateral`, but it's fragile. Consider either:
1. Passing the resolved map as an explicit parameter to `getCollateralKey`/`groupTokensByCollateral`
2. Or adding a runtime guard that throws if `getCollateralKey` is called before init

src/features/tokens/utils.ts
```diff
@@ -61,13 +78,16 @@ export function isValidMultiCollateralToken(
...
-  )
-    return false;
+  // Check if destination is collateralized - has collateral address, is HypNative, or is in SDK's list
+  // Note: Some standards like EvmHypCollateralFiat may not be in TOKEN_COLLATERALIZED_STANDARDS
+  const isDestCollateralized =
+    !!destinationToken.collateralAddressOrDenom ||
+    destinationToken.isHypNative() ||
+    TOKEN_COLLATERALIZED_STANDARDS.includes(destinationToken.standard);
+
+  if (!isDestCollateralized) return false;
```
xeno097, line 90: **Semantic change: relaxed from AND to OR logic.** The old code required `(collateralAddressOrDenom && !isHypNative()) && TOKEN_COLLATERALIZED_STANDARDS.includes(standard)`. The new code uses OR: `collateralAddressOrDenom || isHypNative() || TOKEN_COLLATERALIZED_STANDARDS.includes(standard)`.

This is significantly more permissive — a token with *only* a `collateralAddressOrDenom` but a non-collateralized standard now passes. The comment mentions `EvmHypCollateralFiat` — is that the only case? If so, consider an explicit allowlist for that standard rather than opening the gate broadly. This could change multi-collateral routing behavior for tokens that previously weren't considered multi-collateral.

src/features/tokens/utils.ts
```diff
@@ -132,40 +152,200 @@ export function getTokensWithSameCollateralAddresses(
...
+  const normalizedAddress = normalizeAddress(token.addressOrDenom, token.protocol);
+  const key = `${token.chainName.toLowerCase()}-${token.symbol.toLowerCase()}-${normalizedAddress}`;
+
+  tokenKeyCache.set(token, key);
+  return key;
+}
 
-      const destinationToken = originToken.getConnectionForChain(destination)!.token;
+/**
+ * De-duplicate tokens by collateral address on the same chain
```
xeno097, line 176: **Naming collision with `tokenKey` in `balances/tokens.ts`.** These two functions both produce 'a key for a token' but with different formats:
- `getTokenKey` (here): `chainName-symbol-address` (all lowercase)
- `tokenKey` (`balances/tokens.ts`): `chainName-address-symbol` (preserves case)

In `TokenList.tsx`, **both** are imported and used on the same tokens — `tokenKey()` for balance map lookups and `getTokenKey()` for route map lookups. This is the #1 footgun in this PR. If anyone uses the wrong function, balances silently won't match tokens.

Consider either:
1. Unifying into a single canonical key function
2. Renaming to make the distinction obvious (e.g. `balanceLookupKey` vs `tokenIdentityKey`)

src/features/tokens/utils.ts
xeno097, line 349: **Bug risk: loose symbol fallback.** `routeTokens.find((t) => t.symbol === originToken.symbol)` can match the wrong token if multiple tokens on the same route share a symbol (e.g., different USDC warp route variants on the same chain).

This fallback should at minimum also compare `chainName`, or better yet, compare collateral keys:
```ts
return matchingToken || routeTokens.find(
  (t) => t.symbol === originToken.symbol && t.chainName === originToken.chainName
);
```
If no match exists, returning `undefined` is safer than returning the wrong token — `getTransferToken` already handles the `undefined` case.

src/features/transfer/fees.ts
```diff
@@ -107,16 +108,31 @@ export async function getTransferToken(
...
+
+  if (!isValidMultiCollateralToken(originRouteToken, destinationToken)) {
+    return originRouteToken;
```
xeno097, line 127: **Subtle semantic shift worth documenting.** `findRouteToken` may return a *different* token than `originToken` (one that shares the same collateral key but has the actual connection). Then `isValidMultiCollateralToken` checks against `originRouteToken` instead of the user-selected `originToken`.

This is probably intentional (checking the actual routable token), but the function's original contract was to evaluate the passed-in token. Consider adding a comment explaining this substitution, e.g.:
```ts
// originRouteToken may differ from originToken due to deduplication — 
// it's the actual warpCore token with the connection to destinationChain
```

src/features/tokens/TokenList.tsx
```diff
@@ -0,0 +1,370 @@
...
+    }
+    return [...routable, ...rest].slice(0, maxDefault);
```
xeno097, line 105: **Display/balance mismatch risk.** `balanceTokens` is capped at 50 here (prioritizing routable tokens), but the display `tokens` memo (line ~190) independently sorts and caps at 50 after *different* criteria (USD value, balance, alphabetical). A token could appear in the display list but not in `balanceTokens`, showing no balance.

Consider deriving display tokens from `balanceTokens` rather than independently capping both, or at minimum ensuring the display cap uses the same token set.

src/features/tokens/TokenList.tsx
```diff
@@ -0,0 +1,370 @@
...
+
+      const routeMap = new Map<string, boolean>();
+
+      for (const token of allTokens) {
+        const key = getTokenKey(token);
+        const originToken = selectionMode === 'origin' ? token : counterpartToken;
+        const destToken = selectionMode === 'origin' ? counterpartToken : token;
+        const hasRoute = checkTokenHasRoute(originToken, destToken, collateralGroups);
+        routeMap.set(key, hasRoute);
+      }
```
xeno097, line 220: `startTransition` inside `useEffect` doesn't provide the expected non-blocking benefit. `useTransition` is designed for state updates triggered by user events (e.g., input onChange) to mark them as interruptible. Inside `useEffect`, the route map computation still runs synchronously — only the resulting `setTokenRouteMap` setState is deferred.

For truly non-blocking computation of the O(n) route checking loop, consider `requestIdleCallback`, a Web Worker, or simply batching the computation across frames. Alternatively, if performance is acceptable in practice, just drop the `useTransition` and use a plain `useEffect` — the indirection isn't buying anything.

src/features/transfer/useFeeQuotes.ts
```diff
@@ -12,14 +12,16 @@ import { TransferFormValues } from './types';
...
-  { destination, amount, recipient, tokenIndex }: TransferFormValues,
+  { originTokenKey, destinationTokenKey, amount, recipient: formRecipient }: TransferFormValues,
   enabled: boolean,
   originToken: Token | undefined,
+  destinationToken: IToken | undefined,
   searchForLowestFee: boolean = false,
 ) {
   const multiProvider = useMultiProvider();
   const warpCore = useWarpCore();
   const debouncedAmount = useDebounce(amount, 500);
```
xeno097, line 23: **Performance concern.** `debouncedAmount` is in the query key (line ~53), and `searchForLowestFee` is `!isReview` — meaning every debounced amount change when not in review mode triggers the full multi-collateral fee search (`getTransferToken` with `Promise.allSettled` for collateral balance + fee fetching across all routes).

For multi-collateral tokens this could mean 4-8+ RPC calls per keystroke pause. Consider:
1. Only including `debouncedAmount` in the query key when `searchForLowestFee` is true
2. A longer debounce (1-2s) specifically for the multi-collateral path
3. Caching the winning route token so subsequent amount changes skip re-evaluation unless the route pair changes

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -801,85 +833,104 @@ function ReviewDetails({
...
-        : config.defaultDestinationChain || connectedToken?.token?.chainName || '',
-      tokenIndex: tokenIndex,
-      amount: '',
-      recipient: '',
-    };
-  }, [warpCore, destinationQuery, originQuery, tokenIndex, defaultOriginToken]);
+  return useMemo(() => {
+    if (!originToken || !destinationToken) return true;
+    return checkTokenHasRoute(originToken, destinationToken, collateralGroups);
+  }, [originToken, destinationToken, collateralGroups]);
```
xeno097, line 880: **`useIsRouteSupported` returns `true` when tokens are undefined.** Line 878: `if (!originToken || !destinationToken) return true`. This means if tokens haven't loaded yet or are cleared during a swap, the route appears 'supported' and the submit button is enabled.

Consider returning `false` when tokens are undefined to prevent premature submission:
```ts
if (!originToken || !destinationToken) return false;
```
The UX impact is minimal — the button would show 'Route is not supported' briefly during loading, which is less confusing than allowing submission of an incomplete form.

src/features/wallet/WalletDropdown.tsx
```diff
@@ -0,0 +1,216 @@
...
+    connectFn?.();
+  }, [connectFn]);
+
+  return (
+    <button
+      type="button"
+      onClick={onConnect}
+      className="flex items-center gap-1.5 text-sm text-primary-500 underline transition-colors hover:text-primary-600"
+    >
+      <XIcon width={8} height={8} color={Color.red[500]} />
```
xeno097, line 159: **Minor UX: red X next to "Connect Wallet".** The `<XIcon>` with `Color.red[500]` visually signals 'error' or 'close' rather than 'disconnected'. The `DropdownWalletButton` at line ~192 uses the same pattern but for a dropdown context where it's less confusing. Consider using a gray dot (disconnected) / green dot (connected) pattern consistently, matching the green dot already used in `DropdownWalletButton`.

Xaroz: This is intended as we want user to know that you need your wallet connected for transfers/ other benefits like tx history or balance checking

src/features/tokens/utils.ts
Xaroz: it already compares collateral keys above it
```ts
const originCollateralKey = getCollateralKey(originToken);

// Find a route token that shares the same collateral key (resolved-underlying-aware)
const matchingToken = routeTokens.find((t) => getCollateralKey(t) === originCollateralKey);
  ```

src/features/tokens/utils.ts
Xaroz: As mentioned some tokens do not have collateral address but are inside the `TOKEN_COLLATERALIZED_STANDARDS`, the old logic doesn't really change much from the new changes if you think about it

src/features/tokens/utils.ts
Xaroz: Acknowledged — init order is enforced by `initWarpContext` calling `setResolvedUnderlyingMap` before `buildTokensArray`/`groupTokensByCollateral`. Threading the map as a param would bloat every function signature for marginal safety. The module-level pattern is consistent with `collateralKeyCache` and `tokenKeyCache`.

src/features/tokens/utils.ts
Xaroz: Addressed in #989 — removed `tokenKey()` from `balances/tokens.ts` and unified on `getTokenKey()` from `tokens/utils.ts` everywhere.

src/features/transfer/fees.ts
Xaroz: Added clarifying comment in e6d6872.

src/features/tokens/TokenList.tsx
Xaroz: `balanceTokens` exists to break a circular dependency — `tokens` (display) needs balance data to sort, so balances must be fetched before `tokens` is computed. If `balanceTokens` depended on `tokens`, we'd have `tokens → balanceMap → useTokenBalances(tokens)`. The mismatch at the cap boundary is a minor cosmetic edge case — tokens without fetched balance sort to the bottom naturally.

src/features/tokens/TokenList.tsx
Xaroz: `startTransition` in `useEffect` is valid — it marks the `setTokenRouteMap` update as low-priority so React won't block user interactions (typing, scrolling) while re-rendering with the new route map. The computation is synchronous, but the re-render triggered by the setState is deferred. This is a supported React pattern.

src/features/transfer/useFeeQuotes.ts
Xaroz: Already debounced at 500ms + TanStack Query caches results by query key, so repeated calls with the same inputs are no-ops. The multi-collateral fee search legitimately depends on the amount (checks balances to find cheapest route). Not a perf concern in practice.

src/features/transfer/TransferTokenForm.tsx
Xaroz: Intentional — returning `false` would flash "Route not supported" every time the user opens the form before selecting both tokens, or during token swap animations. The submit button has other guards (amount validation, wallet connection) that prevent premature submission.

xeno097 (approved):

PR #888, opened 2026-01-13
feat: script to download fonts, setup AWS

Xaroz:
- Creates script to fetch font from AWS and write to the local `public/fonts` folder
- Update package.json to include prebuild step that executes the script
- Setup AWS and example in `.env.example` file

### DIFF
--- a/scripts/fetch-fonts.mjs
+++ b/scripts/fetch-fonts.mjs
@@ -1,4 +1,3 @@
-import { config } from 'dotenv';
 import { S3Client, GetObjectCommand } from '@aws-sdk/client-s3';
 import { createWriteStream, mkdirSync, existsSync } from 'fs';
 import { pipeline } from 'stream/promises';
@@ -7,9 +6,6 @@ import { fileURLToPath } from 'url';
 import { dirname, join } from 'path';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
-
-// Load .env.local for local development (Vercel sets env vars directly)
-config({ path: join(__dirname, '..', '.env.local') });
 const FONTS_DIR = join(__dirname, '..', 'public', 'fonts');
 
 // Font files to download from S3
--- a/package.json
+++ b/package.json
@@ -81,7 +81,6 @@
     "@typescript-eslint/eslint-plugin": "^8.13.0",
     "@typescript-eslint/parser": "^8.13.0",
     "autoprefixer": "^10.4.20",
-    "dotenv": "^17.2.3",
     "eslint": "^9.14.0",
     "eslint-config-next": "^15.0.2",
     "eslint-config-prettier": "^9.1.0",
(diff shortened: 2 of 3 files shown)


scripts/fetch-fonts.mjs
```diff
@@ -0,0 +1,76 @@
...
+
+async function fetchFonts() {
+  const { AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_S3_BUCKET, AWS_REGION } = process.env;
+
+  // Gracefully skip if environment variables are not configured
+  if (!AWS_ACCESS_KEY_ID || !AWS_SECRET_ACCESS_KEY || !AWS_S3_BUCKET) {
+    console.warn('AWS environment variables not configured - skipping font download');
+    console.warn('To enable font fetching, set: AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_S3_BUCKET');
+    return;
+  }
```
xeno097, line 22: what happens if the fonts cannot be fetched before building? do we have fallback fonts?

scripts/fetch-fonts.mjs
```diff
@@ -0,0 +1,76 @@
...
+import { pipeline } from 'stream/promises';
+import { Readable } from 'stream';
+import { fileURLToPath } from 'url';
+import { dirname, join } from 'path';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const FONTS_DIR = join(__dirname, '..', 'public', 'fonts');
+
+// Font files to download from S3
+const FONTS = ['PPValve-PlainVariable.woff2', 'PPFraktionMono-Variable.woff2'];
```
paulbalaji, line 12: We probs need a few more fonts for OG as well (unsure how that works with this)

But that's an explorer problem for later

scripts/fetch-fonts.mjs
paulbalaji: ya the css is setup to fallback to system fonts iirc if it cant find/load the custom ones

.env.example
```diff
@@ -1,2 +1,8 @@
 NEXT_PUBLIC_WALLET_CONNECT_ID=12345678901234567890123456789012
-NEXT_PUBLIC_RPC_OVERRIDES='{"chain1":{"http":"https://..."}}'
\ No newline at end of file
+NEXT_PUBLIC_RPC_OVERRIDES='{"chain1":{"http":"https://..."}}'
+
+# AWS S3 Font Storage (server-side only, used by prebuild script)
+AWS_ACCESS_KEY_ID=your_access_key_id
+AWS_SECRET_ACCESS_KEY=your_secret_access_key
+AWS_S3_BUCKET=your_bucket_name
+AWS_REGION=us-east-1
```
paulbalaji, line 8: what if we just had like a bearer auth token instead for curling fonts instead of needing an aws sdk integration?

wonder if that simplifies things a little bit and makes it easier to port to other branches

scripts/fetch-fonts.mjs
Xaroz: if the fonts cannot be fetched it will fallback to the default fonts set in the tailwind.config.js:
```
primary: ['PP Fraktion Mono', 'system-ui', 'sans-serif'],
secondary: ['PP Valve', 'system-ui', 'sans-serif'],
```

.env.example
Xaroz: There are indeed many choices of doing this:
- Use another service outside of S3 that would provide an endpoint
- We could also setup a lambda function that can fetch the fonts through the S3 buckets too

But yeah I basically went with the fastest method I knew

paulbalaji: im down to ship so lets get this mergedddddd

PR #899, opened 2026-01-16
feat: add script for local dependency linking

troykessler:
This PR introduces a script to locally link packages from the monorepo with pnpm pack. It allows local development again after pnpm link did not work anymore after the pnpm catalog has been introduced on the monorepo

## Summary by CodeRabbit

* **New Features**
  * Added link/unlink tooling and npm scripts to streamline using local monorepo packages during development.

* **Documentation**
  * Added detailed developer docs and a new quick-start subsection with usage, workflow, and troubleshooting for monorepo linking.

* **Chores**
  * Updated ignore patterns to exclude local monorepo tarball artifacts from version control.


scripts/unlink-monorepo.js
```diff
@@ -0,0 +1,113 @@
...
+    packOverrides.forEach((name) => {
+      if (packageJson.dependencies[name]) {
+        try {
+          // Fetch the latest version from npm registry
+          const versionOutput = execSync(`npm view ${name} version`, { encoding: 'utf8' }).trim();
+          packageJson.dependencies[name] = versionOutput;
+          console.log(`   ${name} -> ${versionOutput}`);
+        } catch (err) {
+          console.warn(`   ⚠️  Failed to fetch version for ${name}, keeping current value`);
+        }
```
paulbalaji, line 81: codex says
```
    If npm view fails (e.g., offline or unpublished packages like @hyperlane-xyz/tron-sdk), the dependency stays as a file:.monorepo-tarballs/... reference,
    but the script later deletes .monorepo-tarballs and runs pnpm install. That combination makes the unlink flow fail with “tarball not found.” Consider
    preserving the tarballs (or aborting) when a dependency couldn’t be restored, or require a user-supplied version before cleanup.
```

troykessler: resolved 👍

Xaroz (approved):

paulbalaji:
I think you merged regular instead of doing a squash merge. Please take care to use squash merge going ahead

PR #903, opened 2026-01-19
chore: optimize token lookups

Xaroz:
fixes [ENG-3110](https://linear.app/hyperlane-xyz/issue/ENG-3110/optimize-token-collateral-groups)
## Summary                                                                                                                                                                                               
                                                                                                                                                                                                           
  - Add caching for expensive token key computations (getTokenKey, getCollateralKey)                                                                                                                       
  - Add pre-computed tokenByKeyMap to store for O(1) token lookups                                                                                                                                         
  - Memoize TokenButton component to prevent unnecessary re-renders                                                                                                                                        
  - Downgrade react-hooks/set-state-in-effect ESLint rule from error to warning                                                                                                                            
                                                                                                                                                                                                           
### Performance Optimizations

**1. Cache getTokenKey and getCollateralKey (utils.ts)**                                                                                                                                                     

                                                                                                                                                                                                           
  These functions call normalizeAddress() which does expensive string parsing/validation. With ~700 tokens and multiple calls per token during route checking (~2,800 calls), caching provides significant 
  improvement.                                                                                                                                                                                             
                                                                                                                                                                                                           
  - Added module-level Map caches using token object references as keys                                                                                                                                    
  - Cache invalidation is automatic: initWarpContext() creates new token objects via WarpCore.FromConfig(), so old cache entries are never accessed                                                        
                                                                                                                                                                                                           
**2. Add tokenByKeyMap to Store (store.ts, hooks.ts)**                                                                                                                                                       
                                                                                                                                                                                                           
  getTokenByKey() previously used .find() which is O(n). With 700 tokens and 20+ call sites in TransferTokenForm, this adds up.                                                                            
                                                                                                                                                                                                           
  - Pre-compute Map<string, Token> at initialization time                                                                                                                                                  
  - Added useTokenByKeyMap() hook and getTokenByKeyFromMap() helper for O(1) lookups                                                                                                                       
                                                                                                                                                                                                           
**3. Memoize TokenButton (TokenList.tsx)**                                                                                                                                                                   
                                                                                                                                                                                                           
  700+ TokenButton instances were recreated on every parent re-render (search query changes, route map updates).                                                                                           
                                                                                                                                                                                                           
  - Wrapped TokenButton with React.memo to skip re-renders when props haven't changed                                                                                                                      
                                                                                                                                                                                                         
### ESLint Configuration (eslint.config.mjs)                                                                                                                                                                 
                                                                                                                                                                                                           
  Downgraded react-hooks/set-state-in-effect from error to warning. This React Compiler rule is overly aggressive and flags valid patterns like:                                                           
  - setState in effects after async operations                                                                                                                                                                                                                                                                                                       

### Test plan                                                                                                                                                                                                
                                                                                                                                                                                                           
  - TypeScript type checking passes                                                                                                                                                                        
  - All 77 tests pass                                                                                                                                                                                      
  - Manual testing: Open token modal, verify responsive filtering/sorting with 700+ tokens                                                                                                                 
                                                                                                                                                                                                           
  ---

### DIFF
--- a/src/features/tokens/utils.ts
+++ b/src/features/tokens/utils.ts
@@ -11,11 +11,9 @@ import { isChainDisabled } from '../chains/utils';
 import { DefaultMultiCollateralRoutes, TokenChainMap } from './types';
 
 // Module-level caches for expensive key computations
-// These use Map with token object references as keys for O(1) lookups
-// Cache invalidation is automatic: when initWarpContext() runs, it creates NEW token objects
-// via WarpCore.FromConfig(), so old cache entries are never accessed
-const tokenKeyCache = new Map<IToken, string>();
-const collateralKeyCache = new Map<IToken, string>();
+// WeakMap allows automatic garbage collection when token objects are no longer referenced
+const tokenKeyCache = new WeakMap<IToken, string>();
+const collateralKeyCache = new WeakMap<IToken, string>();
 
 // Map of token symbols and token chain map
 // Symbols are not duplicated to avoid the same symbol from being shown
@@ -148,7 +146,7 @@ export function getTokensWithSameCollateralAddresses(
  */
 export function getTokenKey(token: IToken): string {
   const cached = tokenKeyCache.get(token);
-  if (cached !== undefined) return cached;
+  if (!isNullish(cached)) return cached;
 
   const normalizedAddress = normalizeAddress(token.addressOrDenom, token.protocol);
   const key = `${token.chainName.toLowerCase()}-${token.symbol.toLowerCase()}-${normalizedAddress}`;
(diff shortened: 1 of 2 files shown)


src/features/tokens/utils.ts
```diff
@@ -136,10 +143,18 @@ export function getTokensWithSameCollateralAddresses(
  * Uses chainName + lowercase symbol + normalized address
  * Format: "chainName-symbol-addressOrDenom" (stable identifier)
  * Example: "ethereum-usdc-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"
+ *
+ * Results are cached by token object reference for O(1) subsequent lookups.
  */
 export function getTokenKey(token: IToken): string {
+  const cached = tokenKeyCache.get(token);
+  if (cached !== undefined) return cached;
```
xeno097: isNullish no?

src/features/tokens/utils.ts
```diff
@@ -219,23 +234,34 @@ export function groupTokensByCollateral(tokens: Token[]): Map<string, Token[]> {
 /**
  * Get a unique collateral identifier for a token
  * Used to determine if two tokens share the same underlying collateral
+ *
+ * Results are cached by token object reference for O(1) subsequent lookups.
  */
 export function getCollateralKey(token: IToken): string {
+  const cached = collateralKeyCache.get(token);
+  if (cached !== undefined) return cached;
```
xeno097: ditto

PR #919, opened 2026-01-29
feat: prevent sending tokens to warp route and token contract addresses

nambrot-agent:
## Summary

Prevents users from accidentally sending tokens to the **warp route contract address** on the destination chain, which is a common mistake that results in lost funds. Also blocks other contract addresses and well-known invalid recipients.

## Problem

When bridging tokens (e.g., from Ethereum to Base), users sometimes mistakenly enter the warp route contract address or token contract address as the recipient instead of their wallet address. This results in tokens being permanently lost.

## Solution

Validates the recipient address against a blocklist that includes:

1. **Warp route contract addresses** - The HypCollateral/HypSynthetic contract on the destination chain
2. **Collateral token contract addresses** - The underlying token contract (e.g., USDC) on chains where it's used as collateral
3. **Well-known blocked addresses** - Null address (0x0), precompile addresses (0x1-0x9), burn addresses, etc.

## User Experience

When a user enters a blocked address as the recipient, they see a descriptive inline error message:

| Blocked Address Type | Error Message |
|---------------------|---------------|
| Warp Route contract | "This is a Warp Route contract address, not a wallet address" |
| Collateral token | "This is the USDC token contract address, not a wallet address" |
| Null address | "Null address (0x0)" |
| Burn address | "Burn address" |

## Changes

- Added `BLOCKED_RECIPIENT_ADDRESSES` constant for well-known invalid addresses
- Added `blockedAddressesByChainMap` to store state, built from warp route config
- Updated transfer form validation to check recipients against blocked addresses
- Only normalizes EVM hex addresses to lowercase (preserves case for non-EVM like Solana)
- Added comprehensive unit tests

## Testing

- 9 unit tests covering all edge cases
- Manually tested on local dev server

### DIFF
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -314,10 +314,12 @@ export function getBlockedAddressesByChain(
   // Collect all unique chain names first
   const allChainNames = new Set<ChainName>(tokens.map((t) => t.chainName));
 
-  // Helper to add address to the map (case-insensitive for EVM compatibility)
+  // Helper to add address to the map
+  // Only lowercase EVM hex addresses (0x...) for case-insensitive matching
+  // Non-EVM addresses (e.g., base58 for Solana) are case-sensitive and left unchanged
   const addBlockedAddress = (chain: ChainName, address: string, reason: string) => {
     result[chain] ||= new Map<string, string>();
-    const normalizedAddress = address.toLowerCase();
+    const normalizedAddress = /^0x[0-9a-fA-F]+$/i.test(address) ? address.toLowerCase() : address;
     // Don't overwrite existing entries (first reason wins)
     if (!result[chain].has(normalizedAddress)) {
       result[chain].set(normalizedAddress, reason);
--- a/src/features/transfer/TransferTokenForm.tsx
+++ b/src/features/transfer/TransferTokenForm.tsx
@@ -853,7 +853,11 @@ async function validateForm(
     if (!destinationToken) return [{ token: 'Token is required' }, null];
 
     // Check if recipient is a blocked address (warp route, token contract, or well-known address)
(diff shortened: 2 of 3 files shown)


src/features/store.ts
```diff
@@ -280,3 +303,58 @@ function getRouterAddressesByChain(
...
+  // Helper to add address to the map
+  // Only lowercase EVM hex addresses (0x...) for case-insensitive matching
+  // Non-EVM addresses (e.g., base58 for Solana) are case-sensitive and left unchanged
+  const addBlockedAddress = (chain: ChainName, address: string, reason: string) => {
+    result[chain] ||= new Map<string, string>();
+    const normalizedAddress = /^0x[0-9a-fA-F]+$/i.test(address) ? address.toLowerCase() : address;
+    // Don't overwrite existing entries (first reason wins)
+    if (!result[chain].has(normalizedAddress)) {
+      result[chain].set(normalizedAddress, reason);
+    }
```
Xaroz, line 326: why not just normalizeAddress from our utils?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -853,11 +852,14 @@ async function validateForm(
...
-    if (
-      objKeys(routerAddressesByChainMap).includes(destination) &&
-      routerAddressesByChainMap[destination].has(recipient)
-    ) {
-      return [{ recipient: 'Warp Route address is not valid as recipient' }, null];
+    // Check if recipient is a blocked address (warp route, token contract, or well-known address)
+    // Only lowercase EVM hex addresses for case-insensitive matching
+    const normalizedRecipient = /^0x[0-9a-fA-F]+$/i.test(recipient)
+      ? recipient.toLowerCase()
+      : recipient;
```
Xaroz, line 859: same here

PR #930, opened 2026-02-05
feat: fetch transaction history from Explorer API

### LINKED ISSUE #401: Transaction History improvements
## Problem
Tracking transactions are so difficult

If most dapp can help us track transactions other than going to Explorer and get lost in it. Example I tried to track transactions I made on forma bridging Tia from celestia to forma but I was only given the recent transactions.

Also on usenexus I can't seem to find transaction history seems as though it also displays only recent

Hyperlane currently displays only recent transactions in its interface. This proposal aims to enhance the user experience by implementing an all-time transaction history feature. 

## Solution

I would like if a code is written that will display all time history to make it easy for users to track transactions for reference purposes.
To modify the Hyperlane interface to allow users to view their complete transaction history, rather than being limited to recent transactions only. If possible can be built all dapp transaction in one and separate them by segment. Like a live dashboard where activities can be monitored

### 1. Data Retrieval
- Implement a pagination system to fetch historical transaction data from the blockchain in manageable chunks.
- Optimize database queries to efficiently retrieve older transactions.

### 2. User Interface Updates
- Add a new "All Transactions" tab or section in the user dashboard.
- Implement infinite scrolling or pagination controls for navigating through the extended transaction list.

### 3. Performance Considerations
- Implement lazy loading to ensure the interface remains responsive when displaying large amounts of data.
- Consider caching frequently accessed historical data to reduce blockchain queries.

### 4. Filtering and Sorting
- Add options for users to filter transactions by date range, transaction type, or other relevant criteria.
- Implement sorting functionality to allow users to order transactions by date, amount, or other parameters.

## Technical Considerations
- Ensure compatibility with Hyperlane's existing codebase and architecture.
- Address any potential scalability issues that may arise from handling larger datasets.
- Implement appropriate error handling for cases where historical data might be unavailable.

## Testing Plan
- Develop unit tests for new components and functions.
- Perform integration testing to ensure the new feature works seamlessly with existing functionality.
- Conduct user acceptance testing to gather feedback on the usability of the new feature.

## Timeline
[Outline estimated time for each phase of development]

## Conclusion
Implementing an all-time transaction history display will significantly enhance Hyperlane's functionality, providing users with a more comprehensive view of their blockchain activities. This feature aligns with the project's goal of improving transparency and user control in blockchain interactions.

Xaroz:
Closes #401

## Summary
- Replace localStorage-based transaction history with real history from Hyperlane Explorer GraphQL API
- Filter transactions by connected wallet addresses AND configured warp route addresses
- Merge API results with localStorage transfers for non-indexed routes (e.g., M0Portal)
- Add infinite scroll pagination and 60-second auto-refresh
- Display proper sender (tx sender), recipient (from message body), and formatted amount

## Test plan
- [ ] Connect wallet, open sidebar - verify transactions load from API
- [ ] Verify only configured warp route transactions appear
- [ ] Submit a transfer, verify it shows immediately (local state)
- [ ] Wait for indexing, verify local is replaced by API result
- [ ] Refresh page, verify transfer still shows (from API)
- [ ] Scroll down - verify pagination loads more
- [ ] Click refresh button - verify transactions reload
- [ ] Wait 60s - verify auto-refresh updates list
- [ ] Test with multiple wallet types (EVM, Solana, Cosmos)
- [ ] Test clicking on transaction to view details modal

## Summary by CodeRabbit

* **New Features**
  * Message history is now integrated with local transfers in a unified history view
  * Infinite scrolling support for browsing transfer history
  * Transfer history refresh capability in the sidebar
  * Improved transfer details display with better status icon handling

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,11 +1,9 @@
 /** @type {import('next').NextConfig} */
 
 const { version } = require('./package.json');
-const { withSentryConfig } = require('@sentry/nextjs');
 const withBundleAnalyzer = require('@next/bundle-analyzer')({
   enabled: process.env.ANALYZE === 'true',
 });
-
 const isDev = process.env.NODE_ENV !== 'production';
 
 // Sometimes useful to disable this during development
@@ -126,28 +124,41 @@ const nextConfig = {
 
   env: {
     NEXT_PUBLIC_VERSION: version,
+    NEXT_PUBLIC_SENTRY_DSN: process.env.NEXT_PUBLIC_SENTRY_DSN || '',
   },
 
   reactStrictMode: true,
+
+  serverExternalPackages: ['@sentry/nextjs'],
+
+  // Exclude heavy client-only chain SDKs from serverless function file tracing.
+  // These packages are only used client-side and not needed in serverless functions.
+  // Note: @sentry and @opentelemetry are kept for server-side instrumentation (see instrumentation.ts).
+  outputFileTracingExcludes: {
+    '*': [
+      './node_modules/@provablehq/**',
+      './node_modules/@radixdlt/**',
+      './node_modules/@solana/**',
+      './node_modules/@cosmjs/**',
+      './node_modules/@starknet-io/**',
+      './node_modules/ethers/**',
+    ],
+  },
+
+  experimental: {
+    webpackBuildWorker: true,
+    parallelServerBuildTraces: true,
+    optimizePackageImports: [
+      '@hyperlane-xyz/registry',
+      '@hyperlane-xyz/sdk',
+      '@hyperlane-xyz/utils',
+      '@hyperlane-xyz/widgets',
+    ],
+  },
 
   // Skip linting and type checking during builds — CI runs these separately
   eslint: { ignoreDuringBuilds: true },
   typescript: { ignoreBuildErrors: true },
 };
 
-const sentryOptions = {
-  org: 'hyperlane',
-  project: 'warp-ui',
-  authToken: process.env.SENTRY_AUTH_TOKEN,
-  hideSourceMaps: true,
-  tunnelRoute: '/monitoring-tunnel',
-  sourcemaps: { disable: true },
-  telemetry: false,
-  bundleSizeOptimizations: {
-    excludeDebugStatements: true,
-    excludeReplayIframe: true,
-    excludeReplayShadowDom: true,
-  },
-};
-
-module.exports = withBundleAnalyzer(withSentryConfig(nextConfig, sentryOptions));
+module.exports = withBundleAnalyzer(nextConfig);
--- a/sentry.client.config.js
+++ b/sentry.client.config.js
@@ -1,7 +1,7 @@
 import { sentryDefaultConfig } from './sentry.default.config';
 import * as Sentry from '@sentry/nextjs';
 
-if (process.env.NEXT_PUBLIC_SENTRY_DSN) {
+if (typeof window !== 'undefined' && process.env.NEXT_PUBLIC_SENTRY_DSN) {
   Sentry.init({
     ...sentryDefaultConfig,
     integrations: [
--- a/src/instrumentation.ts
+++ b/src/instrumentation.ts
@@ -1,13 +1,13 @@
-import * as Sentry from '@sentry/nextjs';
 import { sentryDefaultConfig } from '../sentry.default.config';
 
-export function register() {
-  if (process.env.NEXT_PUBLIC_SENTRY_DSN) {
-    if (process.env.NEXT_RUNTIME === 'nodejs') {
+export async function register() {
+  if (!process.env.NEXT_PUBLIC_SENTRY_DSN) return;
+  try {
+    const Sentry = await import('@sentry/nextjs');
+    if (process.env.NEXT_RUNTIME === 'nodejs' || process.env.NEXT_RUNTIME === 'edge') {
       Sentry.init({ ...sentryDefaultConfig, defaultIntegrations: false });
     }
-    if (process.env.NEXT_RUNTIME === 'edge') {
-      Sentry.init({ ...sentryDefaultConfig, defaultIntegrations: false });
-    }
+  } catch (error) {
+    console.error('Failed to load Sentry instrumentation:', error);
   }
 }
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -5,6 +5,7 @@ import { Analytics } from '@vercel/analytics/react';
 import type { AppProps } from 'next/app';
 import { ToastContainer, Zoom } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.css';
+import '../../sentry.client.config';
 import { ErrorBoundary } from '../components/errors/ErrorBoundary';
 import { AppLayout } from '../components/layout/AppLayout';
 import { MAIN_FONT } from '../consts/app';


package.json
```diff
@@ -51,6 +51,7 @@
     "@starknet-react/core": "^3.7.2",
     "@tanstack/query-core": "^5.90.12",
     "@tanstack/react-query": "^5.59.20",
+    "@urql/core": "^6.0.1",
```
paulbalaji: **[P1] Unused dependency.** `@urql/core` is added here but never imported anywhere in the codebase. `graphqlClient.ts` uses raw `fetch`. Remove this and the `urql` dep on line 71 to avoid unnecessary bundle bloat (~50KB). Keep `graphql` only if needed as a peer dep.

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,239 @@
...
+  // Auto-refresh every 60s
+  useEffect(() => {
+    if (!walletKey || !warpRouteKey) return;
+
+    const interval = setInterval(() => {
+      if (document.visibilityState === 'visible') {
+        fetchMessages(0)
+          .then((latest) => {
+            setMessages((prev) => {
+              const existingIds = new Set(prev.map((m) => m.msgId));
```
paulbalaji: **[P2] Auto-refresh doesn't update existing messages — stale Pending status.**

This merge is append-only by `msgId`. A message that transitions `Pending → Delivered` (same `msgId`, updated `is_delivered`) will never reflect in the UI until the user manually clicks Refresh.

The 60s refresh fires, fetches the updated row, but `existingIds.has(m.msgId)` filters it out. User sees a transfer stuck as "pending" indefinitely.

**Fix:** Replace matching entries instead of only prepending:
```ts
setMessages((prev) => {
  const latestMap = new Map(latest.map((m) => [m.msgId, m]));
  const updated = prev.map((m) => latestMap.get(m.msgId) ?? m);
  const existingIds = new Set(prev.map((m) => m.msgId));
  const newOnes = latest.filter((m) => !existingIds.has(m.msgId));
  return newOnes.length ? [...newOnes, ...updated] : updated;
});
```

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,239 @@
...
+    }, REFRESH_INTERVAL_MS);
+
+    return () => clearInterval(interval);
+  }, [walletKey, warpRouteKey, fetchMessages]);
+
+  const loadMore = useCallback(async () => {
+    if (isLoading || !hasMore || !walletKey || !warpRouteKey) return;
+
+    setIsLoading(true);
+    try {
```
paulbalaji: **[P2] Offset pagination duplicates/skips rows in a live feed.**

Two issues here:

1. **Offset drift:** If a new message arrives between page loads, `offset = messages.length` shifts the window — the next page repeats the last item from the previous page (or skips one).

2. **No dedup on append:** `setMessages((prev) => [...prev, ...result])` blindly appends, so the duplicate from (1) actually shows up twice in the UI.

**Minimal fix — dedup on append:**
```ts
setMessages((prev) => {
  const existingIds = new Set(prev.map((m) => m.msgId));
  const newMsgs = result.filter((m) => !existingIds.has(m.msgId));
  return [...prev, ...newMsgs];
});
```

**Proper fix:** Use cursor-based pagination (e.g. `send_occurred_at` of last loaded message) instead of offset. Or use TanStack `useInfiniteQuery` which handles this pattern.

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,239 @@
...
+      origin: {
+        timestamp: parseTimestamp(entry.send_occurred_at),
+        hash: postgresByteaToTxHash(entry.origin_tx_hash, originMetadata),
+        from: postgresByteaToAddress(entry.origin_tx_sender, originMetadata),
+        to: postgresByteaToAddress(entry.origin_tx_recipient, originMetadata),
+      },
+      destination: entry.is_delivered
+        ? {
+            timestamp: parseTimestamp(entry.delivery_occurred_at!),
+            hash: postgresByteaToTxHash(entry.destination_tx_hash!, destinationMetadata),
```
paulbalaji: **[P2] Non-null assertions on nullable fields.**

`is_delivered: true` doesn't guarantee the DB has fully populated all destination fields (indexer race). A single null `destination_tx_hash` throws here, the outer `catch` silently drops the entire message, and the user never sees it.

```ts
// These all use ! on nullable types:
hash: postgresByteaToTxHash(entry.destination_tx_hash!, ...)
from: postgresByteaToAddress(entry.destination_tx_sender!, ...)
```

Use null checks and only build the `destination` object if all fields are present.

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,239 @@
...
+      if (document.visibilityState === 'visible') {
+        fetchMessages(0)
+          .then((latest) => {
+            setMessages((prev) => {
+              const existingIds = new Set(prev.map((m) => m.msgId));
+              const newOnes = latest.filter((m) => !existingIds.has(m.msgId));
+              if (newOnes.length === 0) return prev;
+              return [...newOnes, ...prev];
+            });
+          })
```
paulbalaji: **[P3] Silent error swallowing.** Per project guidelines: *"For expected issues (external systems, user input): use explicit error handling."* At minimum log it:
```ts
.catch((err) => logger.warn('Auto-refresh failed', err));
```

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,239 @@
...
+      destination: entry.is_delivered
+        ? {
+            timestamp: parseTimestamp(entry.delivery_occurred_at!),
+            hash: postgresByteaToTxHash(entry.destination_tx_hash!, destinationMetadata),
+            from: postgresByteaToAddress(entry.destination_tx_sender!, destinationMetadata),
+            to: postgresByteaToAddress(entry.destination_tx_recipient!, destinationMetadata),
+          }
+        : undefined,
+      warpTransfer,
+    };
```
paulbalaji: **[P3] Silent `catch` in `parseMessageEntry` hides data issues.** Per project guidelines: *"For unexpected issues (invalid state, broken invariants): fail loudly."* A broken bytea encoding or missing field should be visible in logs, not silently dropped.
```ts
catch (err) {
  logger.error('Failed to parse message entry', entry.id, err);
  return null;
}
```

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,239 @@
...
+  hasMore: boolean;
+  loadMore: () => void;
+  refresh: () => void;
+}
+
+export function useMessageHistory(
+  walletAddresses: string[],
+  warpRouteAddresses: string[],
+  multiProvider: MultiProtocolProvider,
+): UseMessageHistoryResult {
```
paulbalaji, line 43: **[P3] Consider `useInfiniteQuery` from TanStack Query.**

The codebase consistently uses TanStack Query for data fetching (`useFeeQuotes`, `useTokenBalances`, `useTokenPrice`, etc.). This hook manually reimplements loading state, error state, caching, refetch intervals, and pagination — all things `useInfiniteQuery` handles with built-in dedup, stale-while-revalidate, and window-focus refetch.

Not a blocker, but the auto-refresh stale data bug and offset pagination bug would both be solved by switching to `useInfiniteQuery`. Worth considering for a follow-up.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -24,19 +30,73 @@ export function SideBarMenu({
...
+      if (accountInfo.addresses) {
+        accountInfo.addresses.forEach((addrInfo) => {
+          if (addrInfo.address) addresses.push(addrInfo.address);
+        });
+      }
+    });
+    return addresses;
+  }, [accounts]);
+
+  // Get all warp route addresses from configured routes
```
paulbalaji: **[P3] Normalize addresses before querying.**

Raw addresses flow into the API filter. Mixed-case EVM checksums can miss matches since the Explorer stores lowercase. Use the existing `normalizeAddress` util from `src/utils/` — it lowercases EVM while preserving Solana/Cosmos case sensitivity.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -85,28 +197,40 @@ export function SideBarMenu({
...
-                    }}
-                  />
-                ))}
+              {mergedTransfers.length === 0 && !isLoading && (
+                <div className="py-6 text-center text-sm text-gray-500">No transfers yet</div>
+              )}
+              {mergedTransfers.map((item, i) => (
+                <TransferSummary key={i} item={item} onClick={() => handleItemClick(item)} />
+              ))}
             </div>
```
paulbalaji: **[P3] Array index as key.** Since this list is sorted and can change order on refresh (new items prepended, local items replaced by API items), `key={i}` causes incorrect reconciliation. Use:
```tsx
key={item.type === 'local' ? `local-${item.data.timestamp}` : `api-${item.data.msgId}`}
```

src/features/wallet/SideBarMenu.tsx
```diff
@@ -125,7 +249,73 @@ export function SideBarMenu({
   );
 }
 
-function TransferSummary({
+function TransferSummary({ item, onClick }: { item: TransferItem; onClick: () => void }) {
+  const multiProvider = useMultiProvider();
+  const warpCore = useWarpCore();
+
+  if (item.type === 'local') {
```
paulbalaji: **[Nit] Duplicate hook calls.** `TransferSummary` calls `useMultiProvider()` and `useWarpCore()`, then for `local` items immediately delegates to `LocalTransferSummary` which calls them again. For a list of 15+ items this is wasteful. Pass as props instead.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -50,15 +110,63 @@ export function SideBarMenu({
...
+        sender: actualSender,
+        recipient: actualRecipient,
+        originTxHash: msg.origin.hash,
+        msgId: msg.msgId,
+        timestamp: msg.origin.timestamp,
+        // Store the warp route address to look up the token
+        originTokenAddressOrDenom: msg.sender,
+      });
+    }
+    setIsModalOpen(true);
```
paulbalaji, line 126: **[Nit] Duplicate mapping logic.** The API message → `TransferContext` conversion here is nearly identical to the logic in `TransferSummary` (L262-277): same chain resolution, token lookup, amount formatting. Extract a shared helper:
```ts
function apiMessageToTransferContext(msg: MessageStub, multiProvider, warpCore): TransferContext
```

src/features/messages/graphqlClient.ts
```diff
@@ -0,0 +1,30 @@
+const EXPLORER_API_URL = 'https://explorer4.hasura.app/v1/graphql';
```
paulbalaji: **[Nit] Hardcoded API URL.** Other external URLs in the codebase are configurable via env vars or `config.ts`. Consider making this configurable so deployments can point to different explorer instances:
```ts
const EXPLORER_API_URL = process.env.NEXT_PUBLIC_EXPLORER_API_URL || 'https://explorer4.hasura.app/v1/graphql';
```

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,239 @@
...
+      const currentLength = messages.length;
+      const result = await fetchMessages(currentLength);
+      setMessages((prev) => [...prev, ...result]);
+      setHasMore(result.length === PAGE_LIMIT);
+    } catch (err) {
+      setError(err instanceof Error ? err : new Error('Failed to load more messages'));
+    } finally {
+      setIsLoading(false);
+    }
+  }, [isLoading, hasMore, walletKey, warpRouteKey, fetchMessages, messages.length]);
```
paulbalaji: **[P2] `loadMore` race condition — multiple calls can fire with the same offset.**

The guard `if (isLoading || ...)` uses React state, which doesn't update synchronously. If the scroll handler triggers `loadMore` twice in the same tick (before re-render), both calls pass the guard, both read the same `messages.length` as offset, and both append the same page of results — producing duplicates.

This is separate from the offset-drift issue already noted. Two independent bugs on the same function.

**Fix — use a ref as a synchronous lock:**
```ts
const loadingRef = useRef(false);

const loadMore = useCallback(async () => {
  if (loadingRef.current || !hasMore || !walletKey || !warpRouteKey) return;
  loadingRef.current = true;
  setIsLoading(true);
  try {
    const currentLength = messages.length;
    const result = await fetchMessages(currentLength);
    // ... (with dedup from prior review)
  } finally {
    loadingRef.current = false;
    setIsLoading(false);
  }
}, [hasMore, walletKey, warpRouteKey, fetchMessages, messages.length]);
```

A ref updates synchronously, so the second call in the same tick sees `loadingRef.current = true` and bails.

package.json
Xaroz: Fixed in 5d819f77 - removed both `@urql/core` and `urql` dependencies. We're using plain fetch.

src/features/messages/useMessageHistory.ts
Xaroz: Fixed in 5d819f77 - auto-refresh now updates existing messages by msgId (not just prepends new ones). Messages that transition Pending → Delivered will now reflect in the UI.

src/features/messages/useMessageHistory.ts
Xaroz: Fixed in 5d819f77 - added dedup on append in loadMore to handle offset drift. Existing msgIds are filtered out before appending.

src/features/messages/useMessageHistory.ts
Xaroz: Fixed in 5d819f77 - added null checks for all destination fields. Only builds destination object if all fields are present (indexer race protection).

src/features/messages/useMessageHistory.ts
Xaroz: Fixed in 5d819f77 - added `loadingRef` as synchronous lock to prevent multiple calls in same tick.

src/features/messages/useMessageHistory.ts
Xaroz: Fixed in 5d819f77 - now logs with `logger.warn('Auto-refresh failed', err)`

src/features/messages/useMessageHistory.ts
Xaroz: Fixed in 5d819f77 - now logs with `logger.error('Failed to parse message entry', entry.id, err)`

src/features/messages/useMessageHistory.ts
Xaroz: Noted for follow-up. The current implementation addresses the immediate bugs (stale data, offset drift, race conditions). Migration to useInfiniteQuery can be considered in a future iteration.

src/features/wallet/SideBarMenu.tsx
Xaroz: Fixed in 5d819f77 - now using `normalizeAddress()` from @hyperlane-xyz/utils for both wallet and warp route addresses.

src/features/wallet/SideBarMenu.tsx
Xaroz: Fixed in 5d819f77 - now using stable keys: `local-${timestamp}` for local items and `api-${msgId}` for API items.

src/features/wallet/SideBarMenu.tsx
Xaroz: Noted - will consider for follow-up. The hook calls are cached by React so the actual overhead is minimal, but passing as props would be cleaner.

src/features/wallet/SideBarMenu.tsx
Xaroz: Noted - will consider extracting a shared helper in a follow-up. Keeping it inline for now to maintain clarity of the two code paths.

src/features/messages/graphqlClient.ts
Xaroz: Noted - will consider making this configurable in a follow-up if there's a use case for multiple explorer instances.

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,239 @@
...
+            setMessages((prev) => {
+              const existingIds = new Set(prev.map((m) => m.msgId));
+              const newOnes = latest.filter((m) => !existingIds.has(m.msgId));
+              if (newOnes.length === 0) return prev;
+              return [...newOnes, ...prev];
+            });
+          })
+          .catch(() => {
+            // Silently ignore refresh errors
+          });
```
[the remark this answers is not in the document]
Xaroz: Fixed in 5d819f77 - now logs with `logger.warn('Auto-refresh failed', err)`

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,239 @@
...
+            hash: postgresByteaToTxHash(entry.destination_tx_hash!, destinationMetadata),
+            from: postgresByteaToAddress(entry.destination_tx_sender!, destinationMetadata),
+            to: postgresByteaToAddress(entry.destination_tx_recipient!, destinationMetadata),
+          }
+        : undefined,
+      warpTransfer,
+    };
+  } catch {
+    return null;
+  }
```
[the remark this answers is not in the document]
Xaroz: Fixed in 5d819f77 - now logs with `logger.error('Failed to parse message entry', entry.id, err)`

src/features/wallet/SideBarMenu.tsx
```diff
@@ -24,19 +30,73 @@ export function SideBarMenu({
...
+  }, [accounts]);
+
+  // Get all warp route addresses from configured routes
+  const warpRouteAddresses = useMemo(() => {
+    const addresses: string[] = [];
+    Object.values(routerAddressesByChainMap).forEach((addressSet) => {
+      addressSet.forEach((addr) => addresses.push(addr));
+    });
+    return addresses;
+  }, [routerAddressesByChainMap]);
```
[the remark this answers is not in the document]
Xaroz: Fixed in 5d819f77 - now using `normalizeAddress()` from @hyperlane-xyz/utils for both wallet and warp route addresses.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -24,19 +30,73 @@ export function SideBarMenu({
...
+    });
+    return addresses;
+  }, [accounts]);
+
+  // Get all warp route addresses from configured routes
+  const warpRouteAddresses = useMemo(() => {
+    const addresses: string[] = [];
+    Object.values(routerAddressesByChainMap).forEach((addressSet) => {
+      addressSet.forEach((addr) => addresses.push(addr));
+    });
```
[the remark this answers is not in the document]
Xaroz: Fixed in 5d819f77 - replaced forEach with for-of loops to avoid implicit returns.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -50,15 +110,63 @@ export function SideBarMenu({
...
+
+  const handleItemClick = (item: TransferItem) => {
+    if (item.type === 'local') {
+      setSelectedTransfer(item.data);
+    } else {
+      // Convert API message to TransferContext for modal
+      const msg = item.data;
+      const originChain = multiProvider.tryGetChainName(msg.originDomainId) || '';
+      const destChain = multiProvider.tryGetChainName(msg.destinationDomainId) || '';
+
```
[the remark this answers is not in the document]
Xaroz: Kept as-is - empty string fallback is intentional here. The UI gracefully handles missing chain names by showing empty. Throwing would break the entire list for one bad entry.

src/features/messages/graphqlClient.ts
Xaroz: Fixed in b754b6d8 - now configurable via `NEXT_PUBLIC_EXPLORER_API_URL` env var with default fallback.

src/features/wallet/SideBarMenu.tsx
Xaroz: Fixed in b754b6d8 - now passing `multiProvider` and `warpCore` as props instead of calling hooks per list item.

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,267 @@
...
+        return [...prev, ...newMsgs];
+      });
+      setHasMore(result.length === PAGE_LIMIT);
+    } catch (err) {
+      setError(err instanceof Error ? err : new Error('Failed to load more messages'));
+    } finally {
+      loadingRef.current = false;
+      setIsLoading(false);
+    }
+  }, [hasMore, walletKey, warpRouteKey, fetchMessages, messages.length]);
```
paulbalaji: **[Remaining nit] `loadMore` still captures `messages.length` in both the closure body (L162) and the dependency array (L177).**

This means `loadMore` is recreated on every message change, which triggers `handleScroll` recreation (which depends on `loadMore`). Not a correctness bug anymore (the `loadingRef` prevents races), but it's wasteful. Consider using `setMessages` functional update + ref to derive offset:

```ts
const loadMore = useCallback(async () => {
  if (loadingRef.current || !hasMore) return;
  loadingRef.current = true;
  setIsLoading(true);
  try {
    // Derive offset from current state via ref or functional pattern
    const offset = messagesRef.current.length;
    const result = await fetchMessages(offset);
    setMessages((prev) => {
      const existingIds = new Set(prev.map((m) => m.msgId));
      const newMsgs = result.filter((m) => !existingIds.has(m.msgId));
      return [...prev, ...newMsgs];
    });
    setHasMore(result.length === PAGE_LIMIT);
  } catch (err) {
    setError(err instanceof Error ? err : new Error('Failed to load more'));
  } finally {
    loadingRef.current = false;
    setIsLoading(false);
  }
}, [hasMore, fetchMessages]);
```

This removes `messages.length` from the dep array, making `loadMore` stable.

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,267 @@
...
+          setIsLoading(false);
+        }
+      }
+    };
+
+    load();
+    return () => {
+      cancelled = true;
+    };
+  }, [walletKey, warpRouteKey, fetchMessages]);
```
paulbalaji: **[Nit] Initial load effect uses `walletKey`/`warpRouteKey` (JSON strings) as deps but guards on `sortedWallets.length`/`sortedWarpRoutes.length`.**

The effect fires when the JSON-serialized keys change, but the guard checks the sorted array length. These should be consistent — either both use keys or both use arrays. Currently works because they're derived from the same source, but it's a subtle inconsistency that could confuse future maintainers.

Same pattern appears in the auto-refresh effect (L128-153) and `refresh` callback (L179-193).

paulbalaji (commented):
## Re-review (post `5d819f7` → `1b79913`)

Good progress — most of the prior P1/P2 findings are now addressed. Here's the updated scorecard:

### Resolved

| Finding | Status |
|---|---|
| **P1** Unused `urql`/`@urql/core` deps | **Fixed** — removed from package.json |
| **P2** Auto-refresh append-only (stale Pending→Delivered) | **Fixed** — now uses `latestMap` to update existing entries (L137-143) |
| **P2** Offset pagination no dedup on append | **Fixed** — `loadMore` now deduplicates by `msgId` (L165-168) |
| **P2** Non-null assertions on nullable destination fields | **Fixed** — null checks added before building `destination` (L223-236) |
| **P2** `loadMore` race condition (same-tick double call) | **Fixed** — `loadingRef` synchronous lock (L38, L157-159, L174) |
| **P3** Silent error swallowing in auto-refresh | **Fixed** — now uses `logger.warn` (L147) |
| **P3** Silent catch in `parseMessageEntry` | **Fixed** — now uses `logger.error` with entry id (L258-259) |
| **P3** Array index as key | **Fixed** — composite key using type+timestamp/msgId (L183-187) |
| **P3** Address normalization missing | **Fixed** — `normalizeAddress` applied to both wallet and warp route addresses (L59, L72) |
| **P3** Fragile comma-join for memoization key | **Fixed** — uses `JSON.stringify` (L48-49) |
| **Nit** Duplicate hooks in TransferSummary | **Fixed** — `multiProvider` and `warpCore` passed as props (L190-191, L223-224) |
| **Nit** Duplicate mapping logic | **Fixed** — extracted `messageToTransferContext` helper in `useMergedTransferHistory.ts` (L16-50) |
| **Nit** Hardcoded API URL | **Fixed** — configurable via `NEXT_PUBLIC_EXPLORER_API_URL` env var in `config.ts` (L14-15) |
| **Nit** SideBarMenu too large | **Fixed** — merge logic extracted to `useMergedTransferHistory` hook |

### Still open

- **`delivered-icon.svg` still exists on disk** but is unreferenced anywhere in the codebase. Should be deleted.
- **`resetTransfers` store action unused** — still defined in `store.ts` (L59, L141) but no caller. Can be cleaned up.
- **No tests for `messages/` module** — `encoding.ts`, `build.ts`, and `parseMessageEntry` are pure functions. Per project guidelines, new utility functions need unit tests.
- **`error` from `useMessageHistory` still never consumed** — returned by the hook (L195) but `SideBarMenu` doesn't destructure it (L79). API failures are silent to the user.
- **`graphql` dep may be unnecessary** — no source files import from `graphql` directly. If it's only needed as a transitive peer dep of something that was removed (urql), it can likely be removed too. Verify before removing.

These are all minor/cleanup items. The core logic issues are resolved. Nice work addressing the feedback.

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,267 @@
...
+    try {
+      const currentLength = messages.length;
+      const result = await fetchMessages(currentLength);
+      // Dedup to handle offset drift from new messages
+      setMessages((prev) => {
+        const existingIds = new Set(prev.map((m) => m.msgId));
+        const newMsgs = result.filter((m) => !existingIds.has(m.msgId));
+        return [...prev, ...newMsgs];
+      });
+      setHasMore(result.length === PAGE_LIMIT);
```
paulbalaji: **[P2] `loadMore` can loop forever when dedup filters an entire page.**

Scenario:
1. 15 messages loaded (`messages.length = 15`)
2. 15+ new messages arrive at top (via auto-refresh or real-time)
3. User scrolls → `loadMore(offset=15)`
4. API returns 15 results, but all are already in the list (offset shifted)
5. Dedup filters all 15 → `messages.length` stays the same
6. `setHasMore(result.length === PAGE_LIMIT)` → `hasMore = true` (API returned 15)
7. Next scroll → `loadMore(offset=15)` again → same result → **infinite loop**

The dedup correctly prevents duplicate rendering, but `hasMore` is determined by raw API result count, not by how many actually got added.

**Fix — base `hasMore` on deduped count:**
```ts
setMessages((prev) => {
  const existingIds = new Set(prev.map((m) => m.msgId));
  const newMsgs = result.filter((m) => !existingIds.has(m.msgId));
  // If we fetched a full page but nothing was new, offset hasn't advanced
  if (newMsgs.length === 0 && result.length === PAGE_LIMIT) {
    // Advance past duplicates on next call
    // Option A: bump offset by result.length regardless
    // Option B: set hasMore = false (simpler, user can manually refresh)
  }
  return [...prev, ...newMsgs];
});
// Use deduped count for hasMore determination
setHasMore(result.length === PAGE_LIMIT && newMsgs.length > 0);
```

Since `setMessages` is a functional update, `newMsgs` isn't accessible outside it. Simplest correct fix:
```ts
let addedCount = 0;
setMessages((prev) => {
  const existingIds = new Set(prev.map((m) => m.msgId));
  const newMsgs = result.filter((m) => !existingIds.has(m.msgId));
  addedCount = newMsgs.length;
  return [...prev, ...newMsgs];
});
setHasMore(result.length === PAGE_LIMIT && addedCount > 0);
```

Or switch to cursor-based pagination to avoid offset issues entirely.

src/features/messages/useMergedTransferHistory.ts
```diff
@@ -0,0 +1,81 @@
...
+  const destChain = multiProvider.tryGetChainName(msg.destinationDomainId) || '';
+
+  // Use actual sender (tx sender) and recipient (from warp message body)
+  const actualSender = msg.origin.from;
+  const actualRecipient = msg.warpTransfer?.recipient || msg.recipient;
+
+  // Format amount if available
+  let formattedAmount = '';
+  const token = tryFindToken(warpCore, originChain, msg.sender);
+  if (msg.warpTransfer?.amount && token?.decimals) {
```
paulbalaji: **[P3] `token?.decimals` truthy check skips 0-decimal tokens.**

`if (msg.warpTransfer?.amount && token?.decimals)` — when `decimals === 0`, this is falsy, so `formattedAmount` stays as `''` instead of showing the raw amount. Same issue at `SideBarMenu.tsx:255`.

0-decimal tokens are uncommon but valid (some NFT-adjacent tokens, certain protocol tokens). `fromWei(amount, 0)` returns the amount unchanged, which is the correct display.

**Fix:**
```ts
if (msg.warpTransfer?.amount && token?.decimals != null) {
```

Apply in both locations.

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,189 @@
...
+
+        return entries
+          .map((entry) => parseMessageEntry(entry, multiProvider))
+          .filter((m): m is MessageStub => m !== null);
+      },
+      initialPageParam: 0,
+      getNextPageParam: (lastPage, allPages) => {
+        if (lastPage.length < PAGE_LIMIT) return undefined;
+        return allPages.flat().length;
+      },
```
paulbalaji, line 92: **[Nit] `getNextPageParam` offset can drift when `parseMessageEntry` filters entries.**

`allPages.flat().length` counts successfully parsed messages, not raw DB rows. If `parseMessageEntry` returns `null` for some entries (e.g., unknown chain metadata), the offset will be less than actual rows consumed, causing the next page to overlap with already-fetched rows.

The dedup in the `messages` memo (L83-96) prevents duplicate display, so this isn't a bug — just wasted API calls. If it matters for performance, track the raw entry count separately.

Not a blocker.

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,189 @@
...
+        );
+
+        if (result.error) throw result.error;
+
+        const entries = result.data?.message_view;
+        if (!entries?.length) return [];
+
+        return entries
+          .map((entry) => parseMessageEntry(entry, multiProvider))
+          .filter((m): m is MessageStub => m !== null);
```
paulbalaji: **[Nit] `multiProvider` captured in `queryFn` but not in `queryKey`.**

When `multiProvider` changes (e.g., chain overrides applied post-mount), cached pages won't be re-parsed with the new provider until the next `refetchInterval` fires. The `queryKey` only includes wallet/warpRoute keys.

In practice this is fine — `multiProvider` changes are rare and the 60s refetch covers it. But if you want instant invalidation on provider change, add a stable provider identity key to `queryKey`.

src/features/messages/graphqlClient.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    if (!response.ok) {
+      return { error: new Error(`HTTP error: ${response.status}`) };
+    }
+
+    const result = await response.json();
+
+    if (result.errors?.length) {
+      return { error: new Error(result.errors[0].message) };
+    }
```
xeno097, line 27: nit: 

```typescript
const [err] = result.errors ?? []
if (err) {
      return { error: new Error(err.message) };
    }
```

src/features/messages/graphqlClient.ts
```diff
@@ -0,0 +1,30 @@
+import { config } from '../../consts/config';
+
+export async function executeGraphQLQuery<T = unknown>(
+  query: string,
+  variables: Record<string, unknown>,
+): Promise<{ data?: T; error?: Error }> {
```
xeno097: I'd change this to

```typescript
Promise<{ type: "success"; data: T } | {type: "error"; error: Error }>
```

That way is easier to narrow if the response is an error or not

src/features/messages/useMergedTransferHistory.ts
```diff
@@ -0,0 +1,86 @@
...
+import { fromWei, normalizeAddress } from '@hyperlane-xyz/utils';
+import { useMemo } from 'react';
+import { RouterAddressInfo } from '../store';
+import { tryFindToken } from '../tokens/hooks';
+import { TransferContext, TransferStatus } from '../transfer/types';
+import { MessageStatus, MessageStub } from './types';
+
+// Union type for transfer items from both local state and API
+export type TransferItem =
+  | { type: 'local'; data: TransferContext }
```
xeno097: Can we have an enum like object for the possible values of the type filed so we can avoid magic strings when having to narrow the type?

package.json
```diff
@@ -62,6 +62,7 @@
     "cosmjs-types": "^0.9.0",
     "formik": "^2.4.6",
     "framer-motion": "^10.16.4",
+    "graphql": "^16.12.0",
```
xeno097: Are we using this package anywhere? i don't see it being imported anywhere

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -856,7 +857,7 @@ async function validateForm(
 
     if (
       objKeys(routerAddressesByChainMap).includes(destination) &&
-      routerAddressesByChainMap[destination].has(recipient)
+      normalizeAddress(recipient) in routerAddressesByChainMap[destination]
```
xeno097: why not `routerAddressesByChainMap[destination]?.[normalizeAddress(recipient)]`?

src/features/messages/useMessageHistory.ts
```diff
@@ -0,0 +1,197 @@
...
+    [walletKey, warpRouteKey],
+  );
+  const [isRefreshing, setIsRefreshing] = useState(false);
+
+  const { data, isLoading, isFetchingNextPage, error, hasNextPage, fetchNextPage } =
+    useInfiniteQuery({
+      queryKey,
+      queryFn: async ({ pageParam }): Promise<PageResult> => {
+        const wallets = JSON.parse(walletKey) as string[];
+        const warpRoutes = JSON.parse(warpRouteKey) as string[];
```
xeno097: Can't we assert that they are proper arrays instead `assert(Array.isArray(....)`?

src/features/messages/graphqlClient.ts
Xaroz: Done in 59b14bf — destructured the errors array:
```typescript
const [err] = result.errors ?? [];
if (err) {
  return { type: 'error', error: new Error(err.message) };
}
```

src/features/messages/graphqlClient.ts
Xaroz: Done in 59b14bf — added discriminated union type:
```typescript
export type GraphQLResult<T> =
  | { type: 'success'; data: T }
  | { type: 'error'; error: Error };
```

src/features/messages/useMergedTransferHistory.ts
Xaroz: Done in 59b14bf — added `TransferItemType` const object:
```typescript
export const TransferItemType = {
  Local: 'local',
  Api: 'api',
} as const;
```
Used throughout SideBarMenu and useMergedTransferHistory.

package.json
Xaroz: Nope, removed in 59b14bf. Was a leftover from earlier urql exploration.

src/features/transfer/TransferTokenForm.tsx
Xaroz: Done in 59b14bf — simplified to optional chaining:
```typescript
if (routerAddressesByChainMap[destination]?.[normalizeAddress(recipient)])
```

src/features/messages/useMessageHistory.ts
Xaroz: Done in 59b14bf — using `assert` from `@hyperlane-xyz/utils`:
```typescript
assert(Array.isArray(wallets), 'wallets must be an array');
assert(Array.isArray(warpRoutes), 'warpRoutes must be an array');
```

src/features/messages/useMergedTransferHistory.ts
```diff
@@ -0,0 +1,86 @@
...
+    destination: destChain,
+    amount: formattedAmount,
+    sender: actualSender,
+    recipient: actualRecipient,
+    originTxHash: msg.origin.hash,
+    msgId: msg.msgId,
+    timestamp: msg.origin.timestamp,
+    originTokenAddressOrDenom: msg.sender,
+  };
+}
```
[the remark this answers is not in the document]
Xaroz: Fixed in 1e0d9c8 — wrapped `fromWei` in try/catch with `logger.error` in `messageToTransferContext`. This also addresses the `handleItemClick` concern since the root cause is now guarded.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -24,19 +34,99 @@ export function SideBarMenu({
...
+  // Infinite scroll handler
+  const handleScroll = useCallback(() => {
+    const container = scrollContainerRef.current;
+    if (!container || isLoading || !hasMore) return;
+
+    const { scrollTop, scrollHeight, clientHeight } = container;
+    if (scrollHeight - scrollTop - clientHeight < 100) {
+      loadMore();
+    }
+  }, [isLoading, hasMore, loadMore]);
```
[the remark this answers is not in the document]
Xaroz: `loadMore` is just `() => fetchNextPage()` from useInfiniteQuery — the reference is already stable enough. The scroll handler recreating on address change is harmless (no perf issue on a div onScroll). Skipping this one.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -24,19 +34,99 @@ export function SideBarMenu({
...
+  const handleItemClick = (item: TransferItem) => {
+    if (item.type === TransferItemType.Local) {
+      setSelectedTransfer(item.data);
+    } else {
+      setSelectedTransfer(
+        messageToTransferContext(item.data, multiProvider, warpCore, routerAddressesByChainMap),
+      );
+    }
+    setIsModalOpen(true);
+  };
```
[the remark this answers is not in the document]
Xaroz: Addressed via 1e0d9c8 — the root cause (`fromWei` throwing) is now guarded with try/catch inside `messageToTransferContext` itself, so `handleItemClick` is safe without its own wrapper.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -85,28 +170,59 @@ export function SideBarMenu({
...
+                          ? `local-${item.data.timestamp}`
+                          : `api-${item.data.msgId}`
+                      }
+                      item={item}
+                      onClick={() => handleItemClick(item)}
+                      multiProvider={multiProvider}
+                      warpCore={warpCore}
+                      routerAddressesByChainMap={routerAddressesByChainMap}
+                    />
+                  ))}
```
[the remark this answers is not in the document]
Xaroz: Fixed in 1e0d9c8 — local transfer keys now include `originTxHash`/`msgId`:
```
`local-${item.data.timestamp}-${item.data.originTxHash || item.data.msgId || ''}`
```

PR #934, opened 2026-02-06
feat: v2 visual rebrand

Xaroz:
## Summary

Cherry-picks the visual/branding changes from the `xaroz/warp-ui-v2` branch — **only styling, no behavioral changes**.

- New purple/pink color palette, gradient buttons, and accent glows
- Custom fonts (PP Valve + PP Fraktion Mono) with `font-primary`/`font-secondary`
- Redesigned header (centered logo on desktop, hamburger menu on mobile/tablet)
- Simplified footer with gradient logo and nav links (hidden below `lg` to avoid Intercom overlap)
- Updated logos, background SVG, and icon colors to match v2
- TipCard with gradient background and watermark
- ConnectWalletButton and SideBar styled with accent gradients
- Fee section always visible with animated loading dots and dash for no fees
- Discord nav link replaced with Support (help.hyperlane.xyz) + question mark icon
- `font-secondary` applied to form labels, buttons, token select, and transfer detail modal
- Softer background gradient to reduce white glare in top-right corner
- Font fetching script (`scripts/fetch-fonts.mjs`) for S3-hosted fonts

## Summary by CodeRabbit

* **New Features**
  * Mobile navigation menu with hamburger control and additional nav items
  * Multiple new themed icons and gradient/transparent logos

* **Improvements**
  * Simplified, responsive header and footer layouts; footer uses new gradient logo
  * Consolidated fee UI with clearer loading states and updated modal
  * Server-side automated font fetching (downloads ignored from source)

* **Style**
  * Expanded color palette, new gradients/glow shadows, updated global fonts and scrollbar styles
  * Button and tip card visual refinements

### DIFF
--- a/src/features/wallet/SideBarMenu.tsx
+++ b/src/features/wallet/SideBarMenu.tsx
@@ -160,7 +160,7 @@ export function SideBarMenu({
           onScroll={handleScroll}
           className="flex h-full w-full flex-col overflow-y-auto"
         >
-          <div className="bg-accent-gradient shadow-accent-glow w-full px-3.5 py-2 text-base font-normal tracking-wider text-white">
+          <div className="w-full bg-accent-gradient px-3.5 py-2 text-base font-normal tracking-wider text-white shadow-accent-glow">
             Connected Wallets
           </div>
           <AccountList
@@ -170,7 +170,7 @@ export function SideBarMenu({
             className="px-3 py-3"
             chainName={originChainName}
           />
-          <div className="bg-accent-gradient shadow-accent-glow flex w-full items-center justify-between px-3.5 py-2">
+          <div className="flex w-full items-center justify-between bg-accent-gradient px-3.5 py-2 shadow-accent-glow">
             <span className="text-base font-normal tracking-wider text-white">
               Transfer History
             </span>


src/components/nav/Header.tsx
```diff
@@ -1,22 +1,46 @@
...
+      </div>
+
+      {/* Desktop: Centered Logo */}
+      <Link href="/" className="hidden flex-col py-2 lg:flex">
+        <div className="flex items-end">
+          <Image src={Logo} width={46} alt="" className="h-auto" />
+          <Image src={Name} width={150} alt="" className="ml-1.5" />
         </div>
+        <Image src={Title} width={43} alt="" className="self-end" />
+      </Link>
```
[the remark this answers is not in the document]
Xaroz: Fixed in 47be5b6b

src/components/nav/Header.tsx
```diff
@@ -1,22 +1,46 @@
...
+
+      <div className="lg:absolute lg:right-12">
+        <ConnectWalletButton />
       </div>
```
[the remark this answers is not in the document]
Xaroz: Fixed in 47be5b6b — added `relative` to the `<header>` element so the absolute positioning of ConnectWalletButton is anchored correctly.

src/consts/links.ts
```diff
@@ -15,5 +15,6 @@ export const links = {
     'https://github.com/search?q=org:hyperlane-xyz+label:bounty+is:open+is:issue&type=issues&s=&o=desc',
   imgPath: 'https://cdn.jsdelivr.net/gh/hyperlane-xyz/hyperlane-registry@main',
   transferFees: 'https://docs.hyperlane.xyz/docs/protocol/core/fees#fee-estimation',
+  stake: 'https://app.symbiotic.fi/vault/0xE1F23869776c82f691d9Cb34597Ab1830Fb0De58',
```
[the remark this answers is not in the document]
Xaroz: The checksum-cased address is intentional — app.symbiotic.fi uses the checksummed form in its URL routing. Lowercasing would break the link.

src/pages/_app.tsx
```diff
@@ -37,7 +36,7 @@ export default function App({ Component, pageProps }: AppProps) {
   // Note, the font definition is required both here and in _document.tsx
   // Otherwise Next.js will not load the font
   return (
-    <div className={`${MAIN_FONT.variable} font-sans text-black`}>
+    <div className="font-primary text-black">
```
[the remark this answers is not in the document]
Xaroz: Fixed in 47be5b6b — removed the stale comment about the old `next/font` approach.

tailwind.config.js
```diff
@@ -96,6 +114,23 @@ module.exports = {
...
+        'card-gradient': `linear-gradient(180deg, ${theme('colors.white')} 0%, ${theme('colors.cream.200')} 100%)`,
+        'tip-card-gradient': `radial-gradient(74.42% 40.45% at 50% 100%, ${theme('colors.primary.50')} 0%, ${theme('colors.cream.300')} 100%)`,
+      }),
+      boxShadow: ({ theme }) => ({
+        'accent-glow': `inset 2px 2px 13px 2px ${theme('colors.accent.100')}`,
+        'error-glow': `inset 2px 2px 13px 2px ${theme('colors.accent.100')}`,
+        card: `0px 4px 6px ${theme('colors.gray.950')}1A`,
+        button: `0 4px 6px ${theme('colors.gray.950')}1A`,
+        input: `0 0 4px ${theme('colors.gray.400')}4D`,
+      }),
```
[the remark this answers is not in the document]
Xaroz: Intentional — matches the v2 branch exactly. Both `accent-glow` and `error-glow` use `accent.100` in v2. The error state is visually differentiated via `error-gradient` (the background color), not the inset glow shadow.

src/features/transfer/FeeSectionButton.tsx
```diff
@@ -1,8 +1,24 @@
...
+      setDotCount((prev) => (prev % 3) + 1);
+    }, intervalMs);
+
+    return () => clearInterval(interval);
+  }, [isLoading, intervalMs]);
+
+  return 'Loading' + '.'.repeat(dotCount);
+}
+
 export function FeeSectionButton({
```
paulbalaji, line 24: `dotCount` doesn't reset when loading restarts. When `isLoading` flips `false → true`, the animation picks up from its last value.

```diff
 useEffect(() => {
-    if (!isLoading) return;
+    if (!isLoading) {
+      setDotCount(1);
+      return;
+    }
```

The lint fix in `8df9626` was correct approach but the reset line was removed in the same commit.

src/components/nav/Header.tsx
```diff
@@ -1,22 +1,46 @@
...
-      <div className="flex items-start justify-between">
-        <Link href="/" className="flex items-center py-2">
-          <Image src={Logo} width={24} alt="" />
-          <Image src={Name} width={130} alt="" className="ml-2 mt-0.5 hidden sm:block" />
-          <Image src={Title} width={210} alt="" className="ml-2 mt-0.5 pb-px" />
+    <header className="relative flex w-full items-center justify-between bg-primary-25 px-4 py-3 shadow-[0px_4px_7px_rgba(0,0,0,0.05)] lg:justify-center lg:bg-transparent lg:px-6 lg:pb-2 lg:pt-3 lg:shadow-none">
+      {/* Mobile/Tablet: Logo + Hamburger Menu */}
+      <div className="flex items-center gap-3 lg:hidden">
+        <Link href="/">
+          <Image src={Logo} width={36} alt="" className="h-auto" />
```
paulbalaji, line 17: Mobile logo `<Link>` is missing `aria-label="Homepage"` — the desktop version at line 36 has it. Screen readers will just announce "link" with no context.

```diff
-        <Link href="/">
+        <Link href="/" aria-label="Homepage">
```

src/features/wallet/ConnectWalletButton.tsx
```diff
@@ -18,8 +18,8 @@ export function ConnectWalletButton() {
       multiProvider={multiProvider}
       onClickWhenUnconnected={() => setShowEnvSelectModal(true)}
       onClickWhenConnected={() => setIsSideBarOpen(true)}
-      className="rounded-lg bg-white"
-      countClassName="bg-accent-500"
+      className="rounded-lg bg-accent-gradient shadow-accent-glow [&_*]:text-white [&_path]:fill-white"
+      countClassName="bg-white/20"
```
paulbalaji, line 22: `[&_*]:text-white` is a broad descendant override — every child element gets white text. If `ConnectWalletButtonInner` ever renders colored status badges or indicators, this will override their styles silently. Not a blocker, just worth noting the sledgehammer approach.

tailwind.config.js
paulbalaji, line 131: `accent-glow` and `error-glow` are identical — both resolve to `accent.100`. I see Xaroz mentioned this matches v2 intentionally (error differentiated via `error-gradient` background), but this means a dead config entry. Consider either:
- Removing `error-glow` and using `accent-glow` everywhere
- Adding a `// intentionally same as accent-glow` comment

Otherwise someone will "fix" this later thinking it's a bug.

src/components/icons/HyperlaneGradientLogo.tsx
```diff
@@ -0,0 +1,30 @@
...
+          <stop offset="0.3" stopColor="#9A0DFF" />
+          <stop offset="0.6" stopColor="#9A0DFF" />
+          <stop offset="1" stopColor="#FF4FE9" />
+        </linearGradient>
+      </defs>
+    </svg>
+  );
+}
+
+export const HyperlaneFooterLogo = memo(_HyperlaneGradientLogo);
```
paulbalaji: Export name `HyperlaneFooterLogo` doesn't match file name `HyperlaneGradientLogo.tsx`. Anyone searching for where this component is defined will be confused. Either rename the file to `HyperlaneFooterLogo.tsx` or the export to `HyperlaneGradientLogo`.

src/styles/globals.css
```diff
@@ -2,6 +2,30 @@
...
+  font-family: 'PP Valve';
+  src: url('/fonts/PPValve-PlainVariable.woff2') format('woff2');
+  font-weight: 100 900;
+  font-style: normal;
+  font-display: swap;
+}
+
+/* PP Fraktion Mono - Primary font */
+@font-face {
+  font-family: 'PP Fraktion Mono';
```
paulbalaji, line 23: If the custom fonts aren't available (no S3 creds, local dev), `font-display: swap` will use the system-ui fallback from Tailwind config — which is correct. But there's no visible indication to devs that fonts are missing vs. intentionally unstyled. Consider adding a `/* Fonts fetched via: pnpm run fetch-fonts */` hint here.

src/consts/app.ts
```diff
@@ -1,15 +1,8 @@
...
-});
 export const APP_NAME = 'Hyperlane Warp UI Template';
 export const APP_DESCRIPTION = 'A DApp for Hyperlane Warp Route transfers';
 export const APP_URL = 'hyperlane-warp-template.vercel.app';
 export const BRAND_COLOR = Color.primary['500'];
-export const BACKGROUND_COLOR = Color.primary['500'];
-export const BACKGROUND_IMAGE = 'url(/backgrounds/main.svg)';
+export const BACKGROUND_COLOR = '#F8F8FF';
+export const BACKGROUND_IMAGE =
+  'url(/backgrounds/main.svg), radial-gradient(120% 80% at 50% 100%, #E8CAFF 0%, #F2E4FF 60%, #F8F2FF 100%)';
```
paulbalaji: `BACKGROUND_COLOR = '#F8F8FF'` maps to `cream.300` in the Tailwind theme, and `#E8CAFF` maps to `primary.50`. These could reference the Color import for consistency with `BRAND_COLOR` on line 5. The other two hex values (`#F2E4FF`, `#F8F2FF`) aren't in the theme, so they'd need to stay hardcoded unless added to the config.

src/components/nav/Header.tsx
```diff
@@ -1,22 +1,46 @@
...
         </Link>
```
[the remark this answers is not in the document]
Xaroz: Fixed in 7def2961 — added `aria-label="Homepage"` to mobile logo link.

src/components/nav/Header.tsx
Xaroz: Fixed in 7def2961 — added `aria-label="Homepage"` to mobile logo link.

src/features/transfer/FeeSectionButton.tsx
Xaroz: Fixed in 7def2961 — `useLoadingDots` now uses a local counter inside the `setInterval` callback that resets to 0 when the effect re-runs (on `isLoading` change). This avoids both the `set-state-in-effect` and `refs` lint rules while properly resetting the animation on each loading cycle.

src/features/wallet/ConnectWalletButton.tsx
Xaroz: Acknowledged — the broad descendant override is intentional since we don't control `ConnectWalletButtonInner`'s internal DOM. If the widget's internals change upstream, we'd need to revisit. Leaving as-is for now.

tailwind.config.js
Xaroz: Added a comment in 7def2961 explaining they're intentionally identical — error is differentiated via `error-gradient` background, not the glow. Keeping both entries for semantic clarity in usage (`shadow-error-glow` vs `shadow-accent-glow`).

src/components/icons/HyperlaneGradientLogo.tsx
Xaroz: Fixed in 7def2961 — renamed export from `HyperlaneFooterLogo` to `HyperlaneGradientLogo` to match the filename. Updated Footer import accordingly.

src/styles/globals.css
Xaroz: Fixed in 7def2961 — added a comment above the `@font-face` declarations: `/* Custom fonts — fetched via pnpm run fetch-fonts (requires S3 creds). Falls back to system-ui if missing. */`

src/consts/app.ts
Xaroz: Fixed in 7def2961 — `BACKGROUND_COLOR` now uses `Color.cream['300']` and the gradient uses `Color.primary['50']` for the first stop. Added `cream` to `Color.ts` exports. The two non-theme hex values (`#F2E4FF`, `#F8F2FF`) remain hardcoded since they aren't in the palette.

.claude/skills/commit/SKILL.md
```diff
@@ -0,0 +1,59 @@
...
+git diff                      # verify changes are correct
+git add <specific-files>      # only related files
+git diff --staged             # double-check staged changes
+git commit -m "$(cat <<'EOF'
+feat: description of change
+
+- Detail 1
+- Detail 2
+
+Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
```
paulbalaji: can remove the co-authored by

package.json
```diff
@@ -119,7 +120,10 @@
   },
   "scripts": {
     "clean": "rm -rf dist cache .next",
+    "predev": "node scripts/fetch-fonts.mjs",
     "dev": "next dev",
+    "prebuild": "node scripts/fetch-fonts.mjs",
+    "fetch-fonts": "node --env-file-if-exists=.env.local scripts/fetch-fonts.mjs",
```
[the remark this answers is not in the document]
Xaroz: This is fine, we don't need those for `predev` and `prebuild`, `fetch-fonts` should be the one used locally

package.json
```diff
@@ -119,7 +120,10 @@
   },
   "scripts": {
     "clean": "rm -rf dist cache .next",
+    "predev": "node scripts/fetch-fonts.mjs",
     "dev": "next dev",
+    "prebuild": "node scripts/fetch-fonts.mjs",
```
paulbalaji, line 125: **[P2]** `predev` and `prebuild` invoke `node scripts/fetch-fonts.mjs` directly without loading `.env.local`, while the dedicated `fetch-fonts` script uses `--env-file-if-exists=.env.local`. In the standard local workflow (AWS creds in `.env.local`), `pnpm dev` / `pnpm build` will always skip font download.

Fix:
```json
"predev": "node --env-file-if-exists=.env.local scripts/fetch-fonts.mjs",
"prebuild": "node --env-file-if-exists=.env.local scripts/fetch-fonts.mjs",
```

Non-blocking (fonts fall back to `system-ui`), but means custom fonts never appear locally without manual env export.

src/features/transfer/FeeSectionButton.tsx
```diff
@@ -1,8 +1,26 @@
...
+  useEffect(() => {
+    if (!isLoading) return;
+
+    let count = 0;
+    const interval = setInterval(() => {
+      count = (count % 3) + 1;
+      setDotCount(count);
+    }, intervalMs);
+
+    return () => clearInterval(interval);
```
paulbalaji, line 18: **[P3]** When `isLoading` transitions `true → false → true`, `dotCount` retains its stale value for up to 1s before the new interval ticks. The local `count` restarts at 0 but the React state isn't reset.

Fix — add a reset at the top of the effect:
```ts
useEffect(() => {
  if (!isLoading) return;
  setDotCount(1); // reset on new loading cycle
  let count = 0;
  // ...
```

src/components/icons/HyperlaneTransparentLogo.tsx
```diff
@@ -0,0 +1,20 @@
+import { memo } from 'react';
+
+function _HyperlaneTransparentLogo() {
```
paulbalaji, line 3: **[P3]** Every other icon in this PR accepts `DefaultIconProps` and spreads onto the SVG (`BookIcon`, `XIcon`, `HamburgerIcon`, etc). This one accepts zero props — can't pass `className`, `width`, `height`, or `color`.

Consider:
```tsx
function _HyperlaneTransparentLogo({ color, ...props }: DefaultIconProps) {
  return (
    <svg viewBox="0 0 146 127" fill="none" xmlns="..." {...props}>
      <path ... stroke={color || 'white'} />
    </svg>
  );
}
```

src/consts/app.ts
```diff
@@ -1,15 +1,7 @@
...
-  fallback: ['sans-serif'],
-});
 export const APP_NAME = 'Hyperlane Warp UI Template';
 export const APP_DESCRIPTION = 'A DApp for Hyperlane Warp Route transfers';
 export const APP_URL = 'hyperlane-warp-template.vercel.app';
 export const BRAND_COLOR = Color.primary['500'];
-export const BACKGROUND_COLOR = Color.primary['500'];
-export const BACKGROUND_IMAGE = 'url(/backgrounds/main.svg)';
+export const BACKGROUND_COLOR = Color.cream['300'];
+export const BACKGROUND_IMAGE = `url(/backgrounds/main.svg), radial-gradient(120% 80% at 50% 100%, ${Color.primary['50']} 0%, #F2E4FF 60%, #F8F2FF 100%)`;
```
paulbalaji, line 7: **[Nit]** `#F2E4FF` and `#F8F2FF` aren't theme tokens. If the palette shifts, these won't follow. Consider adding them to the tailwind theme or referencing existing tokens.

src/components/icons/StakeIcon.tsx
```diff
@@ -0,0 +1,17 @@
+import { DefaultIconProps } from '@hyperlane-xyz/widgets';
+import { memo } from 'react';
+
+function _StakeIcon({ ...props }: DefaultIconProps) {
+  return (
+    <svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg" {...props}>
+      <circle cx="10" cy="10" r="10" fill="#9A0DFF" />
```
paulbalaji, line 7: **[Nit]** `#9A0DFF` (primary-500) and `#EBD3FF` are hardcoded in the SVG paths. The `DefaultIconProps` color prop is accepted but not applied to these fills. If the palette changes, this icon won't update.

package.json
Xaroz: This is fine, should use fetch-fonts for local usecases

src/features/transfer/FeeSectionButton.tsx
Xaroz: already discussed with claude and we are ok with this

src/components/icons/HyperlaneTransparentLogo.tsx
Xaroz: We are fine with not accepting props for this one

src/consts/app.ts
Xaroz: They don't really fit into the palette and are more like a one time thing

src/components/icons/StakeIcon.tsx
Xaroz: This is a bit of a unique icon too changing it to accept prop would make it look worse, can keep for now

PR #941, opened 2026-02-11
perf(build): conditional Sentry loading, parallel builds, and optimizePackageImports

paulbalaji:
## Summary

Build performance optimizations for Vercel deployments. Sentry error reporting is fully preserved — only the build-time wrapper is removed.

### Vercel build times

| Build | Duration | Change |
|---|---|---|
| Before #929 | ~9m 51s | — |
| After #929 (skip lint/typecheck, disable source maps) | ~4m 25s | -55% |
| **This PR** | **1m 58s** | **-55% from current, -80% from original** |

#### Sentry: build wrapper → runtime-only

- **Remove `withSentryConfig` build wrapper** from `next.config.js`. This was adding build overhead (source map upload, webpack plugin) without meaningful benefit since source maps and tunnel route were already disabled.
- **Client-side init**: `sentry.client.config.js` is now explicitly imported in `_app.tsx` (previously auto-injected by the build wrapper). DSN guard is unchanged — Sentry only initializes when `NEXT_PUBLIC_SENTRY_DSN` is set.
- **Server-side init**: `instrumentation.ts` uses dynamic `await import('@sentry/nextjs')` so the module is skipped entirely when DSN is unset. Wrapped in try/catch per project error handling guidelines. Merged duplicate `nodejs`/`edge` init blocks.
- **`serverExternalPackages: ['@sentry/nextjs']`**: Prevents webpack from bundling Sentry server-side, eliminating `@opentelemetry/instrumentation` and `require-in-the-middle` "Critical dependency" warnings.
- **Trade-off**: Loses the `/monitoring-tunnel` proxy route (was already configured with `sourcemaps: { disable: true }` so minimal impact).

#### Build performance

- **`outputFileTracingExcludes`**: Exclude client-only chain SDKs (`@provablehq`, `@radixdlt`, `@solana`, `@cosmjs`, `@starknet-io`, `ethers`) from serverless function file tracing. Reduces "Collecting build traces" time.
- **Parallel build flags**: `webpackBuildWorker`, `parallelServerCompiles`, `parallelServerBuildTraces`.
- **`optimizePackageImports`**: Tree-shake barrel exports from `@hyperlane-xyz/{registry,sdk,utils,widgets}`.
- **Skip lint/typecheck during build**: CI runs these separately — no need to duplicate in the build step.

### What's NOT changing

| Capability | Status |
|---|---|
| Client-side `captureException` (logger.ts) | ✅ Works — same `@sentry/nextjs` browser export |
| Server-side Sentry init (instrumentation.ts) | ✅ Works — dynamic import when DSN is set |
| DSN-gated loading | ✅ Preserved — no Sentry code runs without DSN |
| `sentry.client.config.js` / `sentry.default.config.js` | ✅ Unchanged |

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes error-reporting initialization and Next build/tracing configuration; misconfiguration could silently disable Sentry or omit needed serverless dependencies if any excluded packages are actually required at runtime.
> 
> **Overview**
> Removes the `withSentryConfig` wrapper from `next.config.js` and shifts Sentry to **runtime-only initialization**: client init is explicitly imported in `_app.tsx` and guarded to run only in the browser, and server init in `instrumentation.ts` now uses a DSN-gated dynamic import with error handling.
> 
> Improves build performance by excluding heavy client-only chain SDKs from `outputFileTracing`, externalizing `@sentry/nextjs` from server bundles, and enabling experimental build optimizations like `webpackBuildWorker`, `parallelServerBuildTraces`, and `optimizePackageImports` for `@hyperlane-xyz/*` packages. ასევე adds `NEXT_PUBLIC_SENTRY_DSN` to Next env config.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit aafe188648e8f05c53f0c0e6930b3b24858df694. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **Chores**
  * Optimized build configuration with webpack build worker and improved package import handling.
  * Enhanced initialization flow for error tracking with conditional runtime support.
  * Streamlined build configuration wrapper for improved maintainability.

### DIFF
--- a/sentry.client.config.js
+++ b/sentry.client.config.js
@@ -1,7 +1,7 @@
 import { sentryDefaultConfig } from './sentry.default.config';
 import * as Sentry from '@sentry/nextjs';
 
-if (process.env.NEXT_PUBLIC_SENTRY_DSN) {
+if (typeof window !== 'undefined' && process.env.NEXT_PUBLIC_SENTRY_DSN) {
   Sentry.init({
     ...sentryDefaultConfig,
     integrations: [


next.config.js
```diff
@@ -126,28 +124,42 @@ const nextConfig = {
...
+      './node_modules/@cosmjs/**',
+      './node_modules/@starknet-io/**',
+      './node_modules/ethers/**',
+    ],
+  },
+
+  experimental: {
+    webpackBuildWorker: true,
+    parallelServerCompiles: true,
+    parallelServerBuildTraces: true,
```
Xaroz: have we considered the tradeoffs of using more memory vs build time, would cost be the same/more?

As for `parallelServerCompiles` this one is only useful if we use any server side routes right? in the warp-ui we don't really use any of that.

Asking claude:

parallelServerCompiles is effectively a no-op for this project. It parallelizes compilation of server-side code (Server Components, API routes, getServerSideProps, etc.) during the build. This project has:      
                            
  - No /api routes                                                                                                                                                                                                   
  - No app/ directory (no Server Components)                                                                                                                                                                         
  - No getServerSideProps / getStaticProps
  - No "use server" directives
  - Pages Router only with _app, _document, index, blocked — all client-side

  To answer your questions:
  - Runtime API/RPC calls from the browser don't count — this is build-time only.
  - Next.js still compiles a minimal server bundle for _document and _app, but it's trivial — parallelizing it adds overhead (worker spawning) for negligible gain.

  Same applies to parallelServerBuildTraces — it parallelizes file tracing for serverless deployment of server entries, which barely exist here.

paulbalaji: on vercel cost is basically the same since it just uses whatever the available memory on a runner is

yeah idk how many of those options are still useful now that we have the sentry filtering tbh, it was mainly just me playing around with it and it didn't impact the project so i thought it's useful to keep in anyway

src/pages/_app.tsx
```diff
@@ -5,6 +5,7 @@ import { Analytics } from '@vercel/analytics/react';
 import type { AppProps } from 'next/app';
 import { ToastContainer, Zoom } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.css';
+import '../../sentry.client.config';
```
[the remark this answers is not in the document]
paulbalaji: Good catch. `sentry.client.config.js` was imported by `_app` and evaluated in both environments. I pushed `aafe1886` to gate client init with `typeof window !== "undefined"`, so `Sentry.init` there only runs in browser. Server init remains in `src/instrumentation.ts` via `register()` for `NEXT_RUNTIME` = `nodejs|edge`.

PR #949, opened 2026-02-16
feat: display USD values for fee quotes

nambrot:
## Summary
- Add `useFeePrices` hook that batches CoinGecko price fetches for fee tokens
- Add `feeUsdDisplay` helpers (`getUsdDisplayForFee`, `getTotalFeesUsd`) to format `≈$X.XX`
- Show USD values inline next to fees in the compact summary (FeeSectionButton), review section (TransferTokenForm), and fee detail modal (TransferFeeModal)
- Gracefully handles missing `coinGeckoId`, API failures, and values < $0.01
- Updated hyperlane packages

## Test plan
- [ ] `pnpm typecheck` — no new type errors
- [ ] `pnpm lint` — passes clean
- [ ] `pnpm test` — all 60 tests pass
- [ ] Start dev server, initiate transfer flow, verify USD appears next to fees in compact summary, review section, and fee modal
- [ ] Test with a token that has no `coinGeckoId` — fees display without USD (no errors)

## Summary by CodeRabbit

* **New Features**
  * USD equivalents shown for all fees in the transfer UI, plus fee % of the transfer amount.
  * Live price lookups to power USD displays and percentage calculations.

* **Components / UX**
  * USD labels added alongside fee amounts and in the fee details modal.

* **Tests**
  * Added comprehensive tests for USD formatting, totals, and percentage calculations.

* **Chores**
  * Made the price-fetch utility available for reuse.

### DIFF
--- a/src/features/balances/feeUsdDisplay.ts
+++ b/src/features/balances/feeUsdDisplay.ts
@@ -1,4 +1,5 @@
 import { TokenAmount } from '@hyperlane-xyz/sdk';
+import { isNullish } from '@hyperlane-xyz/utils';
 import type { FeePrices } from './useFeePrices';
 import { formatUsd } from './utils';
 
@@ -8,7 +9,7 @@ export function getUsdDisplayForFee(
 ): string | null {
   if (!tokenAmount || tokenAmount.amount === 0n) return null;
   const price = feePrices[tokenAmount.token.symbol];
-  if (price == null) return null;
+  if (isNullish(price)) return null;
   const value = tokenAmount.getDecimalFormattedAmount() * price;
   if (value <= 0) return null;
   return formatUsd(value, true);
--- a/src/features/balances/useFeePrices.ts
+++ b/src/features/balances/useFeePrices.ts
@@ -1,4 +1,5 @@
 import { IToken, Token, WarpCoreFeeEstimate } from '@hyperlane-xyz/sdk';
+import { isNullish } from '@hyperlane-xyz/utils';
 import { useQuery } from '@tanstack/react-query';
 import { useMemo } from 'react';
 import { fetchPrices } from '../tokens/useTokenPrice';
@@ -69,7 +70,7 @@ export function useFeePrices(
     for (const [symbol, id] of Object.entries(symbolToId)) {
       // Prefer batch price, fall back to separately fetched price
       const price = batchPrices[id] ?? data?.[id];
-      if (price != null) result[symbol] = price;
+      if (!isNullish(price)) result[symbol] = price;
     }
     return result;
   }, [symbolToId, batchPrices, data]);


paulbalaji:
@claude review

src/features/tokens/useFeePrices.ts
```diff
@@ -0,0 +1,74 @@
...
+  const result: FeePrices = {};
+  for (const [symbol, id] of Object.entries(symbolToId)) {
+    if (data[id]) result[symbol] = data[id];
+  }
+  return result;
+}
+
+type CoinGeckoResponse = Record<string, { usd: number }>;
+
+async function fetchFeePrices(ids: string[]): Promise<FeePrices> {
```
Xaroz: theres a very similar function in `useTokenPrice.tsx`, we should abstract it so that at least the portion that is the fetching part can be re-used

src/features/transfer/TransferFeeModal.tsx
```diff
@@ -55,9 +63,13 @@ export function TransferFeeModal({
...
             ) : (
-              <span>{`${fees.interchainQuote.getDecimalFormattedAmount().toFixed(8) || '0'} ${
-                fees.interchainQuote.token.symbol || ''
-              }`}</span>
+              <span>
+                {`${fees.interchainQuote.getDecimalFormattedAmount().toFixed(8) || '0'} ${fees.interchainQuote.token.symbol || ''}`}
+                {(() => {
+                  const usd = getUsdDisplayForFee(fees.interchainQuote, feePrices);
+                  return usd ? <span className="ml-1 text-gray-500">{usd}</span> : null;
+                })()}
```
Xaroz: generally doing a `IIFE` is considered bad practice in React, one of the reason is that it creates a new function on every render, I think you might wanna do something like this

```tsx
const interchainFeeUsdValue = getUsdDisplayForFee(fees.interchainQuote, feePrices);

return (
  <span>
    {`${fees.interchainQuote.getDecimalFormattedAmount().toFixed(8) || '0'} ${fees.interchainQuote.token.symbol || ''}`}
    {interchainFeeUsdValue && <span className="ml-1 text-gray-500">{interchainFeeUsdValue}</span>}
  </span>
);
```

src/features/transfer/FeeSectionButton.tsx
```diff
@@ -29,11 +35,28 @@ export function FeeSectionButton({
...
+              const usd = getTotalFeesUsd(fees, feePrices);
+              const pct = getFeePercentage(getTotalFeesUsdRaw(fees, feePrices), transferUsd);
+              if (!usd) return null;
+              return (
+                <span className="ml-1 text-gray-500">
+                  {usd}
+                  {pct ? ` (${pct})` : ''}
+                </span>
+              );
+            })()}
```
Xaroz: same here

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -751,25 +765,37 @@ function ReviewDetails({
...
                       <span className="min-w-[7.5rem]">Local Gas (est.)</span>
-                      <span>{`${fees.localQuote.getDecimalFormattedAmount().toFixed(8) || '0'} ${
-                        fees.localQuote.token.symbol || ''
-                      }`}</span>
+                      <span>
+                        {`${fees.localQuote.getDecimalFormattedAmount().toFixed(8) || '0'} ${fees.localQuote.token.symbol || ''}`}
+                        {(() => {
+                          const usd = getUsdDisplayForFee(fees.localQuote, feePrices);
+                          return usd ? <span className="ml-1 text-gray-500">{usd}</span> : null;
+                        })()}
```
Xaroz: I think since this is also repeated you can pass it as a prop to `FeeSectionButton` but no big deal here

src/features/tokens/useFeePrices.ts
```diff
@@ -0,0 +1,74 @@
...
+export function useFeePrices(fees: WarpCoreFeeEstimate | null, knownTokens: Token[]): FeePrices {
+  // Build symbol→coinGeckoId map for fee tokens
+  const symbolToId: Record<string, string> = {};
+  if (fees) {
+    for (const q of [fees.localQuote, fees.interchainQuote, fees.tokenFeeQuote]) {
+      if (!q || q.amount === 0n) continue;
+      const id = resolveCoinGeckoId(q.token, knownTokens);
+      if (id) symbolToId[q.token.symbol] = id;
+    }
+  }
```
Xaroz: an optimization here is that you could check the current fee with `symbolToId` and that way you can skip having to loop through the knownTokens

As a nit, it is a bit easier to read if we do actual words into of just `q` or `t`

src/features/tokens/feeUsdDisplay.ts
```diff
@@ -0,0 +1,32 @@
...
+  feePrices: FeePrices,
+): string | null {
+  let total = 0;
+  for (const quote of [fees.localQuote, fees.interchainQuote, fees.tokenFeeQuote]) {
+    if (!quote || quote.amount === 0n) continue;
+    const id = quote.token.coinGeckoId;
+    if (!id || !feePrices[id]) continue;
+    total += quote.getDecimalFormattedAmount() * feePrices[id];
+  }
+  return formatUsd(total);
```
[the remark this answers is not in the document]
nambrot: Intentionally keeping partial totals — showing what we can price is more useful than hiding the total entirely. Individual fee lines already show USD only when priced, so the user can see which ones are missing.

src/features/tokens/useFeePrices.ts
nambrot: Done — extracted `fetchCoinGeckoPrices` into its own module (`src/features/tokens/fetchCoinGeckoPrices.ts`) and reuse it in both `useTokenPrice` and `useFeePrices`.

src/features/transfer/TransferFeeModal.tsx
nambrot: Good call — replaced all IIFEs with a reusable `UsdLabel` component (`src/features/tokens/UsdLabel.tsx`).

src/features/transfer/FeeSectionButton.tsx
nambrot: Fixed — hoisted the computed values before JSX and also removed the redundant `visible` prop.

src/features/transfer/TransferTokenForm.tsx
nambrot: Switched to the `UsdLabel` component here too — no more repeated inline logic.

src/features/tokens/useFeePrices.ts
nambrot: Both addressed — `resolveCoinGeckoId` now checks the `resolvedSymbols` cache before looping through `knownTokens`, and renamed `q`/`t` to `quote`/`token`.

src/features/tokens/fetchCoinGeckoPrices.ts
```diff
@@ -0,0 +1,25 @@
...
+    );
+    if (!res.ok) {
+      logger.warn(`CoinGecko API error: ${res.status} ${res.statusText}`);
+      return {};
+    }
+    const data: CoinGeckoResponse = await res.json();
+    const prices: Record<string, number> = {};
+    for (const [id, val] of Object.entries(data)) {
+      if (val?.usd) prices[id] = val.usd;
+    }
```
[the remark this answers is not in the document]
nambrot: A token priced at exactly $0 USD is effectively valueless — dropping it is the intended behavior. We don't want to show `≈$0.00` for fees.

src/features/balances/useFeePrices.ts
```diff
@@ -0,0 +1,69 @@
...
+  fees: WarpCoreFeeEstimate | null,
+  knownTokens: Token[],
+  batchPrices: Record<string, number>,
+): FeePrices {
+  const symbolToId: Record<string, string> = {};
+  if (fees) {
+    for (const quote of [fees.localQuote, fees.interchainQuote, fees.tokenFeeQuote]) {
+      if (!quote || quote.amount === 0n) continue;
+      const id = resolveCoinGeckoId(quote.token, knownTokens, symbolToId);
+      if (id) symbolToId[quote.token.symbol] = id;
```
paulbalaji: **Medium — symbol collision risk**: `symbolToId` is keyed by bare `symbol`. If two fee tokens share a symbol across chains/assets (e.g. bridged USDC vs native USDC with different CoinGecko IDs), the last-resolved ID wins and all fee rows display that price.

Same issue on the read side in `feeUsdDisplay.ts:10` (`feePrices[tokenAmount.token.symbol]`).

Consider keying by chain-aware identity (e.g. `` `${chainName}:${symbol}` `` or the existing `tokenKey()` helper).

src/features/balances/feeUsdDisplay.ts
```diff
@@ -0,0 +1,46 @@
...
+import type { FeePrices } from './useFeePrices';
+import { formatUsd } from './utils';
+
+export function getUsdDisplayForFee(
+  tokenAmount: TokenAmount | undefined,
+  feePrices: FeePrices,
+): string | null {
+  if (!tokenAmount || tokenAmount.amount === 0n) return null;
+  const price = feePrices[tokenAmount.token.symbol];
+  if (!price) return null;
```
paulbalaji: **Low — truthy check drops valid `0` price**: `!price` treats `0` as missing. If CoinGecko returns `0` for a depegged/dead token, the USD display silently hides instead of showing `<$0.01`.

```suggestion
  if (price == null) return null;
```

Consistent with this project's preference for `??` over `||`.

src/features/balances/useFeePrices.ts
```diff
@@ -0,0 +1,69 @@
...
+    refetchInterval: FEE_PRICE_REFRESH_INTERVAL,
+    refetchOnWindowFocus: false,
+    refetchOnMount: false,
+  });
+
+  const result: FeePrices = {};
+  for (const [symbol, id] of Object.entries(symbolToId)) {
+    // Prefer batch price, fall back to separately fetched price
+    const price = batchPrices[id] ?? data?.[id];
+    if (price) result[symbol] = price;
```
paulbalaji: **Low — same truthy issue**: `if (price)` drops `0`. Use `if (price != null)` to match the nullish-check convention.

```suggestion
    if (price != null) result[symbol] = price;
```

src/features/balances/useFeePrices.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (resolvedSymbols[feeToken.symbol]) return resolvedSymbols[feeToken.symbol];
+  for (const token of knownTokens) {
+    if (!token.coinGeckoId) continue;
+    if (token.chainName === feeToken.chainName && token.symbol === feeToken.symbol)
+      return token.coinGeckoId;
+  }
+  // Fall back to symbol-only match (native tokens like ETH across chains share a CoinGecko ID)
+  for (const token of knownTokens) {
+    if (token.symbol === feeToken.symbol && token.coinGeckoId) return token.coinGeckoId;
+  }
```
paulbalaji, line 29: **Low — symbol-only fallback**: Reasonable for native gas tokens (ETH, SOL share a CoinGecko ID across chains), but could mismatch for wrapped/bridged tokens that share a symbol with a fundamentally different asset. Consider restricting to native tokens or adding a comment documenting the assumption.

src/features/balances/useFeePrices.ts
Xaroz: Fixed in 6a4fbfbd. Fee tokens are native gas tokens (ETH, SOL, etc.) where symbol maps 1:1 to CoinGecko ID across chains — added a comment documenting this assumption. `resolveCoinGeckoId` already tries chain+symbol match first (line 21-22) before falling back to symbol-only.

src/features/balances/feeUsdDisplay.ts
Xaroz: Fixed in 6a4fbfbd — changed to `price == null`.

src/features/balances/useFeePrices.ts
Xaroz: Fixed in 6a4fbfbd — changed to `price \!= null`.

src/features/balances/useFeePrices.ts
Xaroz: Added comment in 6a4fbfbd documenting the assumption: fee tokens are native gas tokens whose symbol maps 1:1 to CoinGecko ID across chains.

paulbalaji (approved):

PR #954, opened 2026-02-18
feat: show IGP deficit in insufficient gas error

paulbalaji:
## Summary
Improve transfer validation UX for interchain gas shortfalls.

When SDK returns `Insufficient <TOKEN> for interchain gas`, UI now shows exact missing amount (e.g. `need 0.0500 more TIA`).

## What Actually Changed
- In `validateForm`, post-process non-null `warpCore.validateTransfer(...)` results via `enrichBalanceError`.
- `enrichBalanceError` now runs only for IGP error shape (`/^Insufficient (\S+) for interchain gas$/`).
- IGP enrichment uses `getInterchainTransferFee(...).igpQuote` (not `tokenFeeQuote`).
- Missing amount is computed from IGP quote token balance and appended to the message.
- If enrichment fails, fall back to original SDK validation result.
- Removed the extra `gas+transfer` enrichment branch to avoid additional simulation-heavy async calls.
- Preserve sibling validation keys by returning `{ ...result, amount: <enriched message> }`.

## Why
- Correctness: IGP insufficiency must be derived from IGP quote/token, not token fee quote.
- Perf: avoid extra async/RPC load on validation-failure path.
- Safety: enrichment no longer drops non-`amount` validation fields.

## Test Plan
- [x] `pnpm typecheck`
- [x] `pnpm eslint src/features/transfer/TransferTokenForm.tsx`
- [ ] Manually trigger `Insufficient <TOKEN> for interchain gas` and verify `need X more <TOKEN>` message
- [ ] Manually verify non-IGP validation errors are unchanged

---

> [!NOTE]
> **Low Risk**
> UI-side validation message enrichment only; it adds extra fee quote/balance RPCs on a specific error path but doesn’t change transfer execution logic.
> 
> **Overview**
> Improves transfer form validation UX by enriching the SDK’s `Insufficient <TOKEN> for interchain gas` error with the *exact additional amount needed*.
> 
> `validateForm` now post-processes non-null `warpCore.validateTransfer` results via `enrichBalanceError`, which fetches an `igpQuote` from `warpCore.getInterchainTransferFee`, compares against the sender’s relevant token balance, and rewrites only the `amount` field while preserving other validation keys; failures fall back to the original error.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a37b8c8ea9b20a1c7b50dae0b27436aeae5f1333. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/features/analytics/refiner.ts
+++ b/src/features/analytics/refiner.ts
@@ -2,7 +2,7 @@ import { config } from '../../consts/config';
 import { logger } from '../../utils/logger';
 
 const REFINER_PROJECT_ID = process.env.NEXT_PUBLIC_REFINER_PROJECT_ID || '';
-const REFINER_TRANSFER_FORM_ID = process.env.NEXT_PUBLIC_REFINER_TRANSFER_FORM_ID || '';
+const REFINER_RESEARCH_FORM_ID = process.env.NEXT_PUBLIC_REFINER_RESEARCH_FORM_ID || '';
 
 type RefinerFn = (method: string, ...args: unknown[]) => void;
 
@@ -31,7 +31,7 @@ export function refinerIdentifyAndShowTransferForm(params: {
   protocol: string;
   chain: string;
 }): void {
-  if (!config.enableTrackingEvents || !refiner || !REFINER_TRANSFER_FORM_ID) return;
+  if (!config.enableTrackingEvents || !refiner || !REFINER_RESEARCH_FORM_ID) return;
 
   refiner('identifyUser', {
     id: params.walletAddress,
@@ -40,5 +40,5 @@ export function refinerIdentifyAndShowTransferForm(params: {
     protocol: params.protocol,
     chain: params.chain,
   });
-  refiner('showForm', REFINER_TRANSFER_FORM_ID);
+  refiner('showForm', REFINER_RESEARCH_FORM_ID);
 }


src/features/transfer/TransferTokenForm.tsx
```diff
@@ -909,3 +913,37 @@ async function validateForm(
...
+      originTokenAmount,
+      destination,
+      sender,
+      recipient,
+    });
+    const balance = originTokenAmount.token.isFungibleWith(igpQuote.token)
+      ? await originTokenAmount.token.getBalance(warpCore.multiProvider, sender)
+      : await igpQuote.token.getBalance(warpCore.multiProvider, sender);
+    const deficit = igpQuote.amount - balance.amount;
+    if (deficit > 0n) {
```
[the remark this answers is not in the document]
paulbalaji: ⏺ This is a false positive. Let me explain why.                                                                                                                
                                                                                                                                                             
  The SDK's validateTransfer runs checks sequentially. The error "Insufficient X for interchain gas" only fires from Check 2 (SDK line 554):                   
   
  // Check 2: balance < IGP fee alone                                                                                                                          
  const interchainQuoteTokenBalance = originToken.isFungibleWith(interchainQuote.token)                                                                      
      ? senderBalanceAmount   // ← full wallet balance
      : await interchainQuote.token.getBalance(this.multiProvider, sender);
  if (interchainQuoteTokenBalance.amount < interchainQuote.amount) {
      return { amount: `Insufficient ${interchainQuote.token.symbol} for interchain gas` };
  }

  This only triggers when the total wallet balance < IGP fee alone — the transfer amount is irrelevant here because the user can't even cover the fee by
  itself. Our deficit (igpQuote.amount - balance.amount) matches exactly what the SDK checked.

  The case the review describes — enough for IGP but not IGP + transfer — is caught by Check 5, which returns "Insufficient balance for gas and transfer", a
  different message that our igpErrorPattern doesn't match.

  No change needed.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -909,3 +921,64 @@ async function validateForm(
...
+  if (!result.amount) return result;
+  try {
+    if (igpErrorPattern.test(result.amount)) {
+      const { tokenFeeQuote } = await warpCore.getInterchainTransferFee({
+        originTokenAmount,
+        destination,
+        sender,
+        recipient,
+      });
+      if (!tokenFeeQuote) return result;
```
Xaroz: is this correct? what if token fee is not set but theres an IGP quote, don't we wanna show how much is missing?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -909,3 +921,64 @@ async function validateForm(
     return [{ form: errorMsg }, null];
   }
 }
+
+const igpErrorPattern = /^Insufficient (\S+) for interchain gas$/;
+const gasAndTransferError = 'Insufficient balance for gas and transfer';
+
+async function enrichBalanceError(
```
Xaroz, line 926: I think overall it looks good but my biggest complaint about this is that validateTransfer already does a bunch of async calls and this function also adds around 2 or 3 extra (which inside might include more) async calls

src/features/transfer/TransferTokenForm.tsx
paulbalaji: isn't that just the type that all quotes come under tokenFeeQuote? or maybe i'm misunderstanding

src/features/transfer/TransferTokenForm.tsx
paulbalaji: hm true but i'm also unsure how we'd get around that, since wherever we'd put the checks it'd have to be somewhere in the stack right - but i'll have a look

paulbalaji: trimmed it to just do an IGP check and ignore the 2nd set of checks for now

src/features/transfer/TransferTokenForm.tsx
paulbalaji: Good catch. Fixed in `f3f52f0d`.

We now use `igpQuote` for this path (not `tokenFeeQuote`), so if token fee is unset but IGP exists we still compute and show `need X more SYMBOL`. The old `if (!tokenFeeQuote) return result` path is removed.

src/features/transfer/TransferTokenForm.tsx
paulbalaji: Agreed. Also fixed in `f3f52f0d`.

I removed the heavy `gas+transfer` enrichment branch. Now enrichment runs only for IGP errors, with early return for everything else. Extra cost is now just: 1 fee-quote call + 1 balance call (only on IGP error).

src/features/transfer/TransferTokenForm.tsx
Xaroz: nope, token fees are the fees set in the origin based on amount and then there is the IGP

Xaroz (approved):
LGTM other than perhaps resolving agents comment

PR #955, opened 2026-02-18
feat: dark mode

paulbalaji:
## Summary

Adds a dark mode theme system to Warp UI, toggled via header button with system-preference fallback and localStorage persistence. All dark styling is driven by `html[data-theme-mode='dark']` CSS selectors — no runtime style injection.

## Theme infrastructure

- **Pre-hydration init** (`public/theme-init.js`) — synchronous script in `<Head>` sets `data-theme-mode` before first paint, eliminating flash-of-wrong-theme
- **Source order**: localStorage override → system `prefers-color-scheme` → default
- **Toggle**: "Lights on / Lights out" button in header, persists preference
- **Instant switch**: `data-theme-switching` attribute suppresses all CSS transitions during toggle

## Dark mode logo variants

Convention-based system in `src/utils/imageBrightness.ts`:
- Rewrites image URLs to `darkmode-<filename>` in dark mode
- Error fallback to original when dark variant doesn't exist
- LRU cache (500 entries) remembers missing variants to avoid repeated 404s
- `MutationObserver` handles late-loaded widget images (chain edit modal, etc.)

## Dark-styled surfaces

Extensive `globals.css` overrides (~985 lines) covering:
- App shell, header/nav menu, footer
- Transfer form, chain/token fields, text inputs, swap button, max button
- Chain picker modal, token picker modal, chain edit modal
- Transfer details modal (+ timeline arrow clip-paths)
- Fee details modal
- Wallet dropdown, wallet protocol modal
- Sidebar (connected wallets, transfer history)
- Tip card, floating buttons, warning banners

## Component changes

- **New**: `WalletProtocolModal` — custom grid-based protocol selector replacing widget's `MultiProtocolWalletModal`
- **New**: `src/utils/theme.ts` — `parseUiThemeMode()`, `getSystemUiThemeMode()` helpers
- **Modified**: 30+ components to add CSS hook classes (`.transfer-card`, `.sidebar-menu-item`, etc.) enabling dark mode selectors without touching widget internals
- **Removed**: Inline background styles from `AppLayout` (moved to CSS custom properties)
- **Removed**: `BACKGROUND_COLOR`/`BACKGROUND_IMAGE` constants from `app.ts`

## UI fixes & improvements

- **Transfer details layout** — flex → CSS grid (`grid-cols-[1fr_auto_1fr]`) for centered chevrons
- **Duplicate token symbol** — suppressed in details modal and sidebar when origin/dest match
- **Warning banner clipping** — removed clipping context, responsive spacer (`pb-12 sm:pb-10`)
- **Relative timestamps** — sidebar shows `Xs/Xm/Xh ago` for transfers <24h old
- **Border-radius unification** — token picker rows `rounded-md`, token button `rounded-[7px]` to match chain fields
- **Sidebar wallet rows** — full-width flat rows with dividers instead of padded rounded cards
- **Redundant `data-theme-mode`** — removed duplicate attribute from `#app-content` div

## Tests

- `src/utils/date.test.ts` — relative timestamp thresholds, boundary cases
- `src/utils/imageBrightness.test.ts` — dark URL generation, fallback, cache, container processing

## Validation

```
pnpm prettier && pnpm lint && pnpm typecheck && pnpm test
```

---

> [!NOTE]
> **Medium Risk**
> Adds a cross-app theming system (pre-hydration script, context provider, large global CSS overrides) that touches many UI surfaces and third-party widget styling, so regressions are possible despite limited business-logic impact.
> 
> **Overview**
> Introduces a **light/dark theme system** driven by `html[data-theme-mode]`, with a pre-hydration `/theme-init.js` script (wired in `_document`) to avoid flash, a new `ThemeProvider` (wired in `_app`) to toggle/persist the mode, and a header toggle button.
> 
> Adds a **dark-logo variant pipeline** (`imageBrightness.ts`) that rewrites image URLs to `darkmode-*`, falls back on load errors, caches missing variants, and observes late-loaded widget images; `TokenIcon`, `ChainLogo`, and `ChainEditModal` are updated to apply this behavior.
> 
> Reworks styling to support dark mode by moving app background to CSS variables (removing `BACKGROUND_*` constants/inline styles), adding many targeted class hooks across transfer/token/chain pickers, modals, header/footer, and sidebar, and adding UI tweaks like relative timestamps in transfer history (`relativeTimeTicker` + `formatTransferHistoryTimestamp`) and a new custom `WalletProtocolModal` replacing the widget modal.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 15f9e93582fca7fd593f8b6fe2e30901a337b9d6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/features/wallet/SideBarMenu.tsx
+++ b/src/features/wallet/SideBarMenu.tsx
@@ -26,6 +26,7 @@ import { TokenChainIcon } from '../tokens/TokenChainIcon';
 import { TransfersDetailsModal } from '../transfer/TransfersDetailsModal';
 import { TransferContext, TransferStatus } from '../transfer/types';
 import { getIconByTransferStatus, STATUSES_WITH_ICON } from '../transfer/utils';
+import { startRelativeTimeTicker } from './relativeTimeTicker';
 
 export function SideBarMenu({
   onClickConnectWallet,
@@ -146,9 +147,9 @@ export function SideBarMenu({
 
   useEffect(() => {
     if (!isMenuOpen) return;
-    setNowMs(Date.now());
-    const intervalId = window.setInterval(() => setNowMs(Date.now()), 30000);
-    return () => window.clearInterval(intervalId);
+    return startRelativeTimeTicker({
+      onTick: () => setNowMs(Date.now()),
+    });
   }, [isMenuOpen]);
 
   return (
--- a/src/features/wallet/WalletProtocolModal.tsx
+++ b/src/features/wallet/WalletProtocolModal.tsx
@@ -52,7 +52,7 @@ export function WalletProtocolModal({
       dialogClassname="wallet-protocol-dialog"
       panelClassname="wallet-protocol-modal max-w-[44rem] p-4"
     >
-      <div className="wallet-protocol-grid grid grid-cols-2 gap-2.5 py-2 sm:grid-cols-3">
+      <div className="wallet-protocol-grid flex flex-wrap justify-center gap-2.5 py-2">
         {PROTOCOL_OPTIONS.filter((option) => includesProtocol(option.protocol)).map((option) => {
           const Logo = PROTOCOL_TO_LOGO[option.protocol];
           if (!Logo) return null;
@@ -60,7 +60,7 @@ export function WalletProtocolModal({
             <button
               key={option.protocol}
               onClick={() => onClickProtocol(option.protocol)}
-              className="wallet-protocol-card flex w-full flex-col items-center space-y-2.5 rounded-lg border border-gray-200 py-3.5 transition-all hover:bg-gray-100 active:scale-95"
+              className="wallet-protocol-card flex basis-[calc(50%-0.5rem)] flex-col items-center space-y-2.5 rounded-lg border border-gray-200 py-3.5 transition-all hover:bg-gray-100 active:scale-95 sm:basis-[calc(33.333%-0.5rem)]"
             >
               <Logo width={34} height={34} className={option.logoClassName} />
               <div className="wallet-protocol-title tracking-wide text-gray-800">
--- a/src/features/wallet/relativeTimeTicker.test.ts
+++ b/src/features/wallet/relativeTimeTicker.test.ts
@@ -1,0 +1,73 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { startRelativeTimeTicker } from './relativeTimeTicker';
+
+function createFakeDocument(initialVisibilityState: DocumentVisibilityState = 'visible') {
+  let visibilityState = initialVisibilityState;
+  const listeners = new Set<() => void>();
+
+  return {
+    get visibilityState() {
+      return visibilityState;
+    },
+    setVisibilityState(nextVisibilityState: DocumentVisibilityState) {
+      visibilityState = nextVisibilityState;
+    },
+    addEventListener(type: 'visibilitychange', listener: () => void) {
+      if (type === 'visibilitychange') listeners.add(listener);
+    },
+    removeEventListener(type: 'visibilitychange', listener: () => void) {
+      if (type === 'visibilitychange') listeners.delete(listener);
+    },
+    dispatchVisibilityChange() {
+      listeners.forEach((listener) => listener());
+    },
+  };
+}
+
+describe('startRelativeTimeTicker', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('ticks immediately and on the configured interval', () => {
+    const onTick = vi.fn();
+    const documentObject = createFakeDocument();
+    const stop = startRelativeTimeTicker({
+      onTick,
+      intervalMs: 30000,
+      documentObject,
+      timerApi: globalThis,
+    });
+
+    expect(onTick).toHaveBeenCalledTimes(1);
+
+    vi.advanceTimersByTime(30000);
+    expect(onTick).toHaveBeenCalledTimes(2);
+
+    stop();
+  });
+
+  it('ticks when the tab becomes visible again', () => {
+    const onTick = vi.fn();
+    const documentObject = createFakeDocument('hidden');
+    const stop = startRelativeTimeTicker({
+      onTick,
+      intervalMs: 30000,
+      documentObject,
+      timerApi: globalThis,
+    });
+
+    expect(onTick).toHaveBeenCalledTimes(1);
+
+    documentObject.setVisibilityState('visible');
+    documentObject.dispatchVisibilityChange();
+
+    expect(onTick).toHaveBeenCalledTimes(2);
+
+    stop();
+  });
+});
--- a/src/features/wallet/relativeTimeTicker.ts
+++ b/src/features/wallet/relativeTimeTicker.ts
@@ -1,0 +1,41 @@
+interface VisibilityDocument {
+  visibilityState: DocumentVisibilityState;
+  addEventListener: (
+    type: 'visibilitychange',
+    listener: () => void,
+    options?: boolean | AddEventListenerOptions,
+  ) => void;
+  removeEventListener: (
+    type: 'visibilitychange',
+    listener: () => void,
+    options?: boolean | EventListenerOptions,
+  ) => void;
+}
+
+interface StartRelativeTimeTickerOptions {
+  onTick: () => void;
+  intervalMs?: number;
+  documentObject?: VisibilityDocument;
+  timerApi?: Pick<typeof globalThis, 'setInterval' | 'clearInterval'>;
+}
+
+export function startRelativeTimeTicker({
+  onTick,
+  intervalMs = 30000,
+  documentObject = document,
+  timerApi = globalThis,
+}: StartRelativeTimeTickerOptions) {
+  onTick();
+
+  const intervalId = timerApi.setInterval(onTick, intervalMs);
+  const handleVisibilityChange = () => {
+    if (documentObject.visibilityState === 'visible') onTick();
+  };
+
+  documentObject.addEventListener('visibilitychange', handleVisibilityChange);
+
+  return () => {
+    timerApi.clearInterval(intervalId);
+    documentObject.removeEventListener('visibilitychange', handleVisibilityChange);
+  };
+}


next.config.js
```diff
@@ -20,7 +20,7 @@ const STYLE_SRC_HOSTS = ['https://js.refiner.io', 'https://storage.refiner.io'];
 const IMG_SRC_HOSTS = [
   'https://*.walletconnect.com',
   'https://*.githubusercontent.com',
-  'https://cdn.jsdelivr.net/gh/hyperlane-xyz/hyperlane-registry@main/',
+  'https://cdn.jsdelivr.net/gh/hyperlane-xyz/hyperlane-registry@pb/darklogos/',
```
paulbalaji: revert after dark mode logos are merged https://github.com/hyperlane-xyz/hyperlane-registry/pull/1394

src/consts/links.ts
```diff
@@ -13,7 +13,7 @@ export const links = {
   privacyPolicy: 'https://hyperlane.xyz/privacy-policy',
   bounty:
     'https://github.com/search?q=org:hyperlane-xyz+label:bounty+is:open+is:issue&type=issues&s=&o=desc',
-  imgPath: 'https://cdn.jsdelivr.net/gh/hyperlane-xyz/hyperlane-registry@main',
+  imgPath: 'https://cdn.jsdelivr.net/gh/hyperlane-xyz/hyperlane-registry@pb/darklogos',
```
paulbalaji: revert after dark mode logos are merged https://github.com/hyperlane-xyz/hyperlane-registry/pull/1394

src/utils/imageBrightness.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+}
+
+function joinPath(basePath: string, suffixPath: string): string {
+  const base = basePath.endsWith('/') ? basePath.slice(0, -1) : basePath;
+  const suffix = suffixPath.startsWith('/') ? suffixPath : `/${suffixPath}`;
+  return `${base}${suffix}`;
+}
+
+// Some chain logo URIs can point at a different registry branch/base than links.imgPath.
```
paulbalaji: Temporary: this registry URL normalization is only needed until `hyperlane-xyz/hyperlane-registry#1394` merges and dark assets are available on the canonical registry path/branch.

paulbalaji:
Addressed follow-up bot findings in `3d25e00a`:

- `src/components/icons/TokenIcon.tsx`
  - Added guard for `logoDarkFailed` + restored-original state so React `onError` does not race against native fallback handler and incorrectly flip to text fallback.
- `src/features/transfer/TransferTokenForm.tsx`
  - Restored capped warning-banner wrapper height (`max-h-12 sm:max-h-10 overflow-hidden`) to avoid stacked spacer padding when multiple warnings are present.

Validation:
- `pnpm -s prettier`
- `pnpm typecheck`
- `pnpm -s test`
- `pnpm -s lint`

src/pages/_document.tsx
```diff
@@ -29,6 +30,7 @@ export default function Document() {
         <meta property="og:type" content="website" />
         <meta property="og:image" content={`${APP_URL}/logo.png`} />
         <meta property="og:description" content={APP_DESCRIPTION} />
+        <Script src="/theme-init.js" strategy="beforeInteractive" />
```
Xaroz: one thing to clarify is that `beforeInteractive` will still attempt to fetch the script over the network, so if you want it to be really non-flashing on slower network then you would set this as an inline script instead, but I think we are ok with this tradeoff

next.config.js
paulbalaji: Done — registry PR #1394 merged, branch now uses `@main`.

src/consts/links.ts
paulbalaji: Done — reverted to `@main`.

src/utils/imageBrightness.ts
paulbalaji: Done — `normalizeRegistryAssetSrc` and `joinPath` removed now that registry PR #1394 is merged.

src/pages/_document.tsx
paulbalaji: Addressed — switched from `next/script beforeInteractive` to a raw `<script src>` tag for reliable synchronous execution. Inlining would be blocked by CSP (`script-src` has no `unsafe-inline`), so keeping it as an external same-origin fetch.

src/components/icons/TokenIcon.tsx
```diff
@@ -22,18 +27,50 @@ export function TokenIcon({ token, size = 32 }: Props) {
...
-    </Circle>
+    <span className="inline-flex">
+      <Circle size={size} bgColorSeed={bgColorSeed} title={title}>
+        {imageSrc && !fallbackToText ? (
+          <img
+            src={imageSrc}
+            className="h-full w-full p-0.5"
+            onLoad={(event) => scheduleDarkLogoProcessing(event.currentTarget)}
+            onError={(event) => {
+              const img = event.currentTarget;
```
Xaroz: can we have this function as a named function instead?

src/consts/app.ts
```diff
@@ -1,7 +1,23 @@
...
+  if (value === 'light' || value === 'dark') return value;
+  return null;
+}
+
+export function getSystemUiThemeMode(): UiThemeMode {
+  if (typeof window === 'undefined' || typeof window.matchMedia !== 'function') {
+    return DEFAULT_UI_THEME_MODE;
+  }
+  return window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+}
```
Xaroz: I don't think these functions belong in the const folder

src/styles/globals.css
```diff
@@ -262,3 +298,641 @@ Needs to be higher than Intercom chatbox (z-index: 2147483000)
 #refiner-widget-wrapper {
   z-index: 2147483001 !important;
 }
+
+/*
+Theme mode
+==========
+Many selectors below target widget internals (.htw-*); revisit during @hyperlane-xyz/widgets upgrades.
+*/
+.theme-toggle {
```
Xaroz: while this is fine, I'd prefer if we stick to already established tailwind-css architecture for css classes, if you need to define a group that is reusable what we do is have them in a const

```tsx
const styles = {
   themeTogle: "border color-red"
}

<div className = {`${styles.themeToggle}`}> </div>
```

src/styles/globals.css
```diff
@@ -262,3 +298,641 @@ Needs to be higher than Intercom chatbox (z-index: 2147483000)
...
+  border: 1px solid rgba(154, 13, 255, 0.35);
+  background: rgba(255, 255, 255, 0.85);
+  color: #4d0f7b;
```
Xaroz: as for colors, its better if they are defined the tailwind theme so that its reusable and if any value needs to be changed then we know where

src/components/layout/AppLayout.tsx
```diff
@@ -31,6 +77,36 @@ export function AppLayout({ children }: PropsWithChildren) {
...
+    const knownLogoImages = document.querySelectorAll(
+      'img[data-logo-handlers-bound="true"], img[data-logo-original-src], img[data-logo-dark-src]',
+    );
+    knownLogoImages.forEach((img) => processDarkLogoImage(img as HTMLImageElement));
+
+    const frame = window.requestAnimationFrame(() => {
+      delete document.documentElement.dataset.themeSwitching;
+    });
+    return () => window.cancelAnimationFrame(frame);
+  }, [themeMode, hasExplicitThemePreference]);
```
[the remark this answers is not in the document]
paulbalaji: claude says "⏺ That's not a real issue — data-theme-mode on <html> is the source of truth for CSS and is always kept in sync by this effect whenever themeMode changes.     
  There's no scenario where clearing localStorage externally would cause staleness, since the React state still drives the attribute. The effect correctly runs
   on every themeMode change regardless of preference state. This is a false positive from the bot — no action needed."

src/features/wallet/SideBarMenu.tsx
```diff
@@ -249,89 +261,151 @@ function TransferSummary({
...
-      destToken: tryFindToken(warpCore, destChain, msg.recipient),
-    };
-  }, [item.type, item.data, multiProvider, warpCore, routerAddressesByChainMap]);
+  // Format amount using wire decimals from precomputed map
+  let formattedAmount = '';
+  if (msg.warpTransfer?.amount && token) {
+    const normalizedSender = normalizeAddress(msg.sender);
+    const routerInfo = routerAddressesByChainMap[originChain]?.[normalizedSender];
+    const wireDecimals = routerInfo?.wireDecimals ?? token.decimals;
+    formattedAmount = fromWei(msg.warpTransfer.amount, wireDecimals);
```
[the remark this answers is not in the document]
paulbalaji: False positive — the try/catch is present in the final code (lines 298-302 of SideBarMenu.tsx). The bot likely read an intermediate rebase diff rather than the resolved state.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -249,89 +261,151 @@ function TransferSummary({
...
+  }
```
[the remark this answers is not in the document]
paulbalaji: Same false positive as the previous comment — the try/catch with `logger.error` is present in the final code at lines 298-302. The bot is diffing against an intermediate rebase state, not the actual HEAD.

src/components/icons/TokenIcon.tsx
```diff
@@ -21,19 +27,56 @@ export function TokenIcon({ token, size = 32 }: Props) {
...
+
+    // Dark-variant misses should fall back to original logo, not text.
+    if (isDarkFallbackError || isDarkVariantSrc) {
+      markDarkLogoMissing(current);
+      const nextSrc = fallbackSrc || original;
+      if (nextSrc) {
+        img.src = nextSrc;
+        return;
+      }
+    }
```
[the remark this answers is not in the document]
paulbalaji: Re: cursor bot "Stale markDarkLogoMissing cache when original also fails" on `TokenIcon.tsx`:

Not a real issue. The dark variant URL IS missing — that cache entry is correct regardless of whether the original also fails. The cache prevents re-fetching a known-missing dark variant on subsequent renders, which is the intended behavior. The entry is ephemeral (in-memory Map, cleared on page reload) so there's no persistence concern.

src/features/transfer/TransfersDetailsModal.tsx
```diff
@@ -252,10 +252,10 @@ export function TransfersDetailsModal({
 
         <div>
           <div className="mt-4 flex w-full items-center justify-center rounded-sm border border-gray-400/25 bg-card-gradient py-2 shadow-card">
-            <div className="flex items-baseline font-secondary text-sm font-normal">
+            <div className="flex items-center font-secondary text-sm font-normal">
               <span>{amount}</span>
               <span className="ml-1">{token?.symbol || ''}</span>
-              {destToken && (
+              {destToken && destToken.symbol !== token?.symbol && (
```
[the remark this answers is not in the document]
paulbalaji: intentional because "X USDC -> X USDC" is meaningless given we already say "originchain -> destinationchain" right below it

the "amount ticker -> amount ticker" is only really useful if there are differences

src/features/wallet/WalletProtocolModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+  protocols?: ProtocolType[];
+  onProtocolSelected?: (protocol: ProtocolType) => void;
+}
+
+const PROTOCOL_OPTIONS = [
+  { protocol: ProtocolType.Ethereum, title: 'EVM', subtitle: 'an EVM' },
+  { protocol: ProtocolType.Sealevel, title: 'Solana', subtitle: 'a Solana' },
+  { protocol: ProtocolType.Cosmos, title: 'Cosmos', subtitle: 'a Cosmos' },
+  { protocol: ProtocolType.Starknet, title: 'Starknet', subtitle: 'a Starknet' },
+  { protocol: ProtocolType.Radix, title: 'Radix', subtitle: 'a Radix' },
```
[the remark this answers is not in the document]
paulbalaji: Intentional — `PROTOCOL_OPTIONS` defines the known set of protocols with their display metadata (name, icon). The component already skips protocols without a `PROTOCOL_TO_LOGO` entry. New protocols need UI metadata to be displayable, same as the widget it replaced. Not a regression.

src/components/layout/AppLayout.tsx
```diff
@@ -31,6 +77,38 @@ export function AppLayout({ children }: PropsWithChildren) {
...
+    if (hasExplicitThemePreference) return;
+    if (typeof window.matchMedia !== 'function') return;
+    const mediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
+    const handleChange = (event: MediaQueryListEvent) => {
+      const nextMode: UiThemeMode = event.matches ? 'dark' : 'light';
+      setThemeMode(nextMode);
+    };
+
+    mediaQuery.addEventListener('change', handleChange);
+    return () => mediaQuery.removeEventListener('change', handleChange);
```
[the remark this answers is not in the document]
paulbalaji: `addEventListener` on `MediaQueryList` is supported since Safari 14+ (2020). This app targets modern browsers (Next.js 15, ES2020+). The legacy `addListener` API is itself deprecated. No polyfill needed.

src/components/layout/AppLayout.tsx
```diff
@@ -23,6 +48,27 @@ export function AppLayout({ children }: PropsWithChildren) {
...
+  const [themeMode, setThemeMode] = useState<UiThemeMode>(() => {
+    if (typeof window === 'undefined') return DEFAULT_UI_THEME_MODE;
+    const docTheme = parseUiThemeMode(document.documentElement.dataset.themeMode);
+    if (docTheme) return docTheme;
+    const storedTheme = getStoredThemeMode();
+    return storedTheme ?? getSystemUiThemeMode();
+  });
+  const [hasExplicitThemePreference, setHasExplicitThemePreference] = useState(() => {
+    return getStoredThemeMode() !== null;
+  });
```
Xaroz: A suggestion here would probably be move everything theme related to a `ThemeContext` file (or maybe even simpler set it in the zustand store) and wrap it in `_app.tsx` or inside `AppLayout.tsx`

This way you can separate the theming code from the layout and not have the layout bloated with theming code

Xaroz: Then inside the Header file you can call the hooks related to the theme like
`const { somethingHere, toggleThemeMode } = useTheme()`

src/features/transfer/TransfersDetailsModal.tsx
```diff
@@ -260,10 +260,10 @@ export function TransfersDetailsModal({
...
```
Xaroz: I think we still wanna showcase the dest token even if its the same symbol as the origin

src/features/wallet/SideBarMenu.tsx
```diff
@@ -310,29 +328,54 @@ function TransferSummary({
...
+                {amount}
+              </span>
+            )}
+            <span
+              className={`sidebar-menu-token-text text-sm font-normal text-gray-800 ${amount ? 'ml-1' : ''}`}
+            >
               {token?.symbol || 'Unknown token'}
             </span>
-            {destToken && (
+            {destToken && destToken.symbol !== token?.symbol && (
```
Xaroz: ditto

src/styles/globals.css
```diff
@@ -303,3 +340,995 @@ ChainDetailsMenu to match v2 styling
...
+
+html[data-theme-mode='dark'] .sidebar-menu .text-gray-600,
+html[data-theme-mode='dark'] .sidebar-menu .text-gray-500,
+html[data-theme-mode='dark'] .sidebar-menu .text-gray-400 {
+  color: var(--dark-text-secondary) !important;
+}
+
+.tip-card {
+  border: 0;
+}
```
Xaroz: can we revisit some of the classes inside the .css file, for example this one doesn't have to be here

src/styles/globals.css
```diff
@@ -303,3 +340,995 @@ ChainDetailsMenu to match v2 styling
...
+
+html[data-theme-mode='dark'] .tip-card-more {
+  border-color: rgba(154, 13, 255, 0.8) !important;
+  background: rgba(154, 13, 255, 0.2) !important;
+  color: #ffffff !important;
+}
+
+html[data-theme-mode='dark'] .tip-card-more:hover {
+  background: rgba(154, 13, 255, 0.3) !important;
```
Xaroz, line 520: and some of these colors seems repeated my ideal scenario would be to set dark mode with tailwind and call stuff like this:
`dark:bg-primary-color/90`

(you can ignore this if its too much work or can be done in a different PR)

src/utils/imageBrightness.test.ts
```diff
@@ -0,0 +1,140 @@
+import { afterEach, beforeEach, describe, expect, test } from 'vitest';
+import {
+  markDarkLogoMissing,
+  processDarkLogoImage,
+  processDarkLogosInContainer,
+  resetDarkLogoCache,
+  toOriginalVariantSrc,
+} from './imageBrightness';
+
+class FakeImage {
```
Xaroz, line 10: maybe worth adding e2e tests for some of these stuff instead?

src/features/wallet/SideBarMenu.tsx
paulbalaji: Updated here too. Sidebar transfer summaries now keep the destination token visible whenever `destToken` is available.

src/components/layout/AppLayout.tsx
paulbalaji: Moved theme state/effects into `ThemeProvider` in `src/features/theme/ThemeContext.tsx` and wrapped it in `_app.tsx`. `Header` now consumes `useTheme()` directly, so `AppLayout` no longer carries the theming state.

src/features/transfer/TransfersDetailsModal.tsx
paulbalaji: Updated. Transfer details now renders the destination token summary whenever `destToken` exists, even when the origin/destination symbols match.

src/styles/globals.css
paulbalaji: Agreed on the direction, but I kept this PR scoped to the behavior fixes / small cleanup above. A broader move toward Tailwind dark utilities or shared CSS vars is better as a follow-up instead of mixing it into this branch.

src/utils/imageBrightness.test.ts
paulbalaji: Makes sense, but for this branch I kept coverage at the unit level since the changed logic is deterministic src switching/fallback. Visual theme/logo e2e coverage is a good follow-up, not something I wanted to bolt onto this PR.

src/styles/globals.css
paulbalaji: Removed that redundant `.tip-card { border: 0; }` override.

src/styles/globals.css
```diff
@@ -31,10 +31,47 @@ Place font files in public/fonts/ directory
...
+}
+
+body,
+.warp-init-gate,
+#app-content {
+  background-color: var(--app-bg-color);
+  background-image: var(--app-bg-image);
+  background-size: cover;
+  background-repeat: no-repeat;
+  background-position: center;
```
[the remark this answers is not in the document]
paulbalaji: Fixed in e6f9da37. `body` now only carries the background-color fallback; the full background image/gradient stays on `.warp-init-gate` and `#app-content`, which are the actual full-screen app containers.

PR #964, opened 2026-02-20
feat: batch CoinGecko price fetching with 5min cache

Xaroz:
## Summary
- With the addition of token balances and USD values in the token selection modal, CoinGecko's free API (5-15 req/min dynamic limit) was getting rate limited due to multiple independent price-fetching hooks each polling every 60s
- Consolidated `useTokenPrice` and `useTokenPrices` into a single `useTokenPrices()` hook that batch-fetches all token prices in one API call (~2KB response)
- Deduplicated CoinGecko IDs are extracted during warp context init and stored in Zustand, so no repeated iteration over tokens on every render
- Increased stale time from 60s to 5min — CoinGecko's free tier only updates prices every 1-2 min server-side anyway, and these are informational estimates not execution prices

## Test plan
- [ ] `pnpm typecheck` passes
- [ ] `pnpm test` passes (60/60)
- [ ] Amount input shows ≈$X.XX USD estimate
- [ ] Token selection modal shows USD values and sorts by value
- [ ] Network tab: single CoinGecko request on load, next one after 5 min

## Summary by CodeRabbit

* **Refactor**
  * Improved token price fetching mechanism for better performance and efficiency

### DIFF
--- a/src/features/tokens/useTokenPrice.tsx
+++ b/src/features/tokens/useTokenPrice.tsx
@@ -1,11 +1,8 @@
-import { Token } from '@hyperlane-xyz/sdk';
 import { useQuery } from '@tanstack/react-query';
-import { useMemo } from 'react';
 import { logger } from '../../utils/logger';
-import { TransferFormValues } from '../transfer/types';
-import { getTokenByIndex, useWarpCore } from './hooks';
+import { useStore } from '../store';
 
-const TOKEN_PRICE_REFRESH_INTERVAL = 60_000; // 60s
+const PRICE_STALE_TIME = 300_000; // 5 min
 
 type CoinGeckoResponse = Record<string, { usd: number }>;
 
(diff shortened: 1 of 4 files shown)


src/features/tokens/useTokenPrice.tsx
paulbalaji, line 1: what if we just had an api in the back which kept track of prices and the frontend just fetched that (maybe overkill tho)

Xaroz: I am not opposed if we do have an API outside of nextjs, one thing to remember is that Vercel serverless function do not really have a cache so you would essentially be fetching from coingecko each time the nextjs API was called

paulbalaji: yeah i thought as much tbh, probs fine as-is for now

PR #973, opened 2026-02-26
chore: update dependencies to latest

yjamin:


package.json
```diff
@@ -21,9 +21,9 @@
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^2.2.0",
     "@hyperlane-xyz/registry": "23.14.0",
```
paulbalaji: maybe update this after releasing https://github.com/hyperlane-xyz/hyperlane-registry/pull/1400

PR #980, opened 2026-02-26
feat: v2 chain filter, sort, edit & disabled chain support

Xaroz:
## Summary

Restores and enhances chain panel features for the v2 unified token/chain picker (`UnifiedTokenChainModal` → `ChainFilterPanel` → `ChainList`).

- **V2 chain metadata editing**: Restore ability to edit chain metadata (RPCs, block explorers) that was lost in v2's new chain panel. Pencil icon in chain toolbar + "Edit" CTA on unhealthy chain warning banner. Reuses `ChainDetailsMenu` from `@hyperlane-xyz/widgets` with v2 CSS overrides.
- **V2 chain filter & sort**: Add icon-driven filter/sort controls to v2's compact chain panel (282px), mirroring widgets' `chainSearch()` logic:
  - Filter by protocol (EVM, SVM, Cosmos, etc.) and type (Mainnet/Testnet)
  - Sort by name, chain ID, or protocol with asc/desc toggle
  - FunnelIcon / ArrowIcon / PencilIcon compact toolbar
- **V2 disabled chains**: Chains with `availability.status === Disabled` are grayed out (opacity-50), unclickable, and sorted to the bottom. Gated behind `config.shouldDisableChains`.
- **Config cleanup**: Remove unused v1 flags `showAddRouteButton`, `showAddChainButton`, `showDisabledTokens`.

### DIFF
--- a/src/features/chains/MobileChainQuickSelect.tsx
+++ b/src/features/chains/MobileChainQuickSelect.tsx
@@ -33,8 +33,8 @@ export function MobileChainQuickSelect({
         .map((name) => allChains.find((c) => c.name === name))
         .filter((c): c is ChainInfo => !!c);
 
-      // Fill remaining slots with other chains (not in preferred list)
-      const remaining = allChains.filter((c) => !preferredSet.has(c.name));
+      // Fill remaining slots with non-preferred, non-disabled chains
+      const remaining = allChains.filter((c) => !preferredSet.has(c.name) && !c.disabled);
       const slotsToFill = DEFAULT_MAX_VISIBLE_CHAINS - preferred.length;
       const fillers = remaining.slice(0, Math.max(0, slotsToFill));
 
@@ -46,7 +46,7 @@ export function MobileChainQuickSelect({
       };
     }
 
-    // Otherwise use first N chains alphabetically
+    // Otherwise use first N chains
     return {
       visibleChains: allChains.slice(0, DEFAULT_MAX_VISIBLE_CHAINS),
       hasMore: allChains.length > DEFAULT_MAX_VISIBLE_CHAINS,
--- a/src/features/chains/chainFilterSort.test.ts
+++ b/src/features/chains/chainFilterSort.test.ts
@@ -29,6 +29,12 @@ const chains: ChainInfo[] = [
     chainId: 1399811149,
     protocol: ProtocolType.Sealevel,
   }),
+  makeChain({
+    name: 'cosmoshub',
+    displayName: 'Cosmos Hub',
+    chainId: 'cosmoshub-4',
+    protocol: ProtocolType.Cosmos,
+  }),
   makeChain({ name: 'sepolia', displayName: 'Sepolia', chainId: 11155111, isTestnet: true }),
   makeChain({ name: 'disabled-chain', displayName: 'Disabled', chainId: 999, disabled: true }),
 ];
@@ -68,6 +74,13 @@ describe('chainSearch', () => {
       expect(result).toHaveLength(1);
       expect(result[0].name).toBe('polygon');
     });
+
+    it('matches broadly by chainId substring', () => {
+      // '1' appears in chainIds: 1, 137, 42161, 1399811149, 11155111, 999 → all except 999
+      const result = search({ query: '1' });
+      expect(result.map((c) => c.name)).not.toContain('disabled-chain');
+      expect(result.length).toBeGreaterThan(1);
+    });
   });
 
   describe('filter', () => {
@@ -99,7 +112,14 @@ describe('chainSearch', () => {
     it('sorts by name ascending (default)', () => {
       const result = search({});
       const names = result.filter((c) => !c.disabled).map((c) => c.name);
-      expect(names).toEqual([...names].sort());
+      expect(names).toEqual([
+        'arbitrum',
+        'cosmoshub',
+        'ethereum',
+        'polygon',
+        'sepolia',
+        'solanamainnet',
+      ]);
     });
 
     it('sorts by name descending', () => {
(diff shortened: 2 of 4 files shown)


Xaroz:
Addressing third round:

**1. Missing ChainName import in ChainEditModal** — Not an issue. `ChainName` is declared globally in `src/global.d.ts` (along with `ChainId`, `Address`, etc.), so no import needed. Typecheck passes clean.

**2. Filter dropdown clipping on mobile** — Valid, fixing now. Changed filter dropdown to `right-0` on mobile with `max-w-[calc(100vw-2rem)]` to prevent overflow, and `md:left-0 md:right-auto` for desktop positioning.

Xaroz:
@claude more reviews 4th round but try to address everything at once now please

src/features/chains/chainFilterSort.ts
```diff
@@ -0,0 +1,102 @@
...
+        return included;
+      })
+      // Sort options
+      .sort((c1, c2) => {
+        // Disabled chains always at the bottom
+        if (c1.disabled && !c2.disabled) return 1;
+        if (!c1.disabled && c2.disabled) return -1;
+
+        if (sort.sortBy === ChainSortBy.ChainId) {
+          const result = Number(c1.chainId) - Number(c2.chainId);
```
paulbalaji: **Bug: `Number(chainId)` produces NaN for non-EVM string chainIds, making sort non-deterministic.**

`ChainId` is typed `number | string` (per `global.d.ts`). For Cosmos/Sealevel chains, `chainId` can be a non-numeric string like `"cosmoshub-4"`. `Number("cosmoshub-4")` returns `NaN`, and `NaN - NaN` is `NaN`, which violates the `Array.sort()` comparator contract.

The upstream widgets `chainSearch` ([ChainSearchMenu.tsx:299-301](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/widgets/src/chains/ChainSearchMenu.tsx#L299-L301)) handles this with string comparison:

```ts
sortValue1 = c1.chainId.toString();
sortValue2 = c2.chainId.toString();
// ...
sortValue1.localeCompare(sortValue2)
```

Suggested fix:
```diff
-          const result = Number(c1.chainId) - Number(c2.chainId);
-          return sort.sortOrder === SortOrder.Asc ? result : -result;
+          const result = c1.chainId.toString().localeCompare(c2.chainId.toString());
+          return sort.sortOrder === SortOrder.Asc ? result : -result;
```

src/features/chains/MobileChainQuickSelect.tsx
```diff
@@ -30,14 +30,8 @@ export function MobileChainQuickSelect({
...
-        .map((name) => {
-          const chain = allChains.find((c) => c.name === name);
-          return {
-            name,
-            displayName: chain?.displayName || name,
-            chainId: chain?.chainId || 1,
-          };
-        });
+        .map((name) => allChains.find((c) => c.name === name))
+        .filter((c): c is ChainInfo => !!c);
```
paulbalaji, line 34: **Bug: Disabled-chain bypass on mobile.**

This refactor correctly switches to `ChainInfo` objects (which now carry `disabled`), but `visibleChains` is rendered at line 72-76 without any disabled check. Desktop `ChainList` passes `disabled={chain.disabled}` to block interaction, but mobile pills remain fully clickable.

When `config.shouldDisableChains` is enabled, users can still select disabled chains from the mobile quick-select.

Suggested fix — filter disabled chains out of `visibleChains`, or add disabled styling + click guard:
```ts
// Option A: filter
const remaining = allChains.filter((c) => !preferredSet.has(c.name) && !c.disabled);

// Option B: guard in render
<button disabled={chain.disabled} className={chain.disabled ? 'opacity-50' : ...}>
```

Also: the comment on line 49 says "first N chains alphabetically" but after the pre-sort removal from `useChainInfos` (round 2 fix), the array is now in `Object.values(chainMetadata)` insertion order — not alphabetical.

src/features/chains/ChainEditModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+    },
+    [chainName, setChainMetadataOverrides],
+  );
+
+  const displayName = metadata?.displayName ?? metadata?.name ?? chainName;
+
+  if (!metadata) return null;
+
+  return (
+    <Modal isOpen={isOpen} close={close} panelClassname="p-0 max-w-lg overflow-hidden">
```
paulbalaji, line 38: **Nit: Redundant double header.**

This renders `<ModalHeader>Edit ${displayName}</ModalHeader>`, then the embedded `ChainDetailsMenu` widget renders its own internal header with a "Back" link and an h2 `"${displayName} Metadata"`. The user sees two headers:

```
┌──────────────────────────┐
│ Edit Ethereum             │  ← ModalHeader
│ ← Back                    │
│ [Logo] Ethereum Metadata  │  ← Widget header
└──────────────────────────┘
```

Consider removing the `ModalHeader` and relying on the widget's built-in header, or hiding the widget's header via CSS.

src/features/tokens/UnifiedTokenChainModal.tsx
```diff
@@ -63,6 +66,11 @@ export function UnifiedTokenChainModal({
     setShowMobileChainList(false);
   };
 
+  const handleEditChain = (chainName: string) => {
+    onClose();
```
paulbalaji: **Nit: `onClose()` resets internal state visibly during close animation.**

`onClose()` resets `chainSearch`, `tokenSearch`, `selectedChain`, and `showMobileChainList` AND triggers the modal close animation. The user may briefly see the modal content snap to defaults before the animation finishes closing.

Consider calling just the parent's `close()` (i.e., `setIsModalOpen(false)`) here instead of `onClose()`, or deferring the state reset until after the animation completes.

src/features/chains/chainFilterSort.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        filter: { type: FilterTestnet.Mainnet, protocol: ProtocolType.Ethereum },
+      });
+      expect(result.every((c) => !c.isTestnet && c.protocol === ProtocolType.Ethereum)).toBe(true);
+    });
+  });
+
+  describe('sort', () => {
+    it('sorts by name ascending (default)', () => {
+      const result = search({});
+      const names = result.filter((c) => !c.disabled).map((c) => c.name);
```
paulbalaji, line 114: **Test quality: Weak sort assertion.**

This extracts names from the result, sorts them, and compares to the same array sorted — which always passes regardless of input order. Stronger assertion:
```ts
expect(names).toEqual(['arbitrum', 'ethereum', 'polygon', 'sepolia', 'solanamainnet']);
```

Same pattern in the descending and chainId tests.

Also missing: Protocol sort test (`ChainSortBy.Protocol` path at `chainFilterSort.ts:95-97` has zero coverage).

src/features/chains/chainFilterSort.ts
```diff
@@ -0,0 +1,102 @@
...
+  const q = query.trim().toLowerCase();
+  return (
+    data
+      // Query search
+      .filter(
+        (chain) =>
+          !q ||
+          chain.name.toLowerCase().includes(q) ||
+          chain.displayName.toLowerCase().includes(q) ||
+          chain.chainId.toString().includes(q),
```
paulbalaji, line 69: **Test gap: Broad chainId substring matching.**

`.includes(q)` means a query of `"1"` matches every chain whose chainId contains digit 1 (1, 137, 42161, 1399811149, etc.). This mirrors the widgets pattern so may be intentional, but no test documents this behavior. Consider adding a test for `search({ query: '1' })` to make the expected result count explicit.

src/features/chains/chainFilterSort.ts
Xaroz: Fixed in 73608a1. Switched from `Number(chainId)` to `localeCompare` for chainId sorting — matches the widgets implementation and handles string chainIds like `"cosmoshub-4"`. Added test with cosmoshub (string chainId) to cover this.

src/features/chains/MobileChainQuickSelect.tsx
Xaroz: Fixed in 73608a1. Disabled chains are now filtered out of `remaining` slots in `MobileChainQuickSelect`, so they never appear in the quick-select bar.

src/features/chains/ChainEditModal.tsx
Xaroz: Keeping as-is — `ModalHeader` provides v2 design consistency (shared header style across all modals). The widget's internal header is nested inside the content area. Can revisit if we want to suppress the widget header via CSS.

src/features/tokens/UnifiedTokenChainModal.tsx
Xaroz: Fixed in 73608a1. `handleEditChain` now calls `close()` instead of `onClose()` so state isn't reset visibly during the close animation.

src/features/chains/chainFilterSort.test.ts
Xaroz: Fixed in 73608a1. Assertions now use explicit expected arrays instead of self-referential sorts. Added protocol sort test and cosmoshub chain with string chainId `"cosmoshub-4"` to cover localeCompare sort path.

src/features/chains/chainFilterSort.ts
Xaroz: Intentional — matches widgets' `chainSearch()` behavior (substring `.includes()`). Added a test case documenting this broad matching behavior so it's explicitly covered.

PR #996, opened 2026-03-09
feat: predicate warp route support

troykessler:
## Summary

- Adds attestation fetching from the Nexus UI for warp routes with a `PredicateRouterWrapper` configured
- Server-side proxy (`/api/predicate/attestation`) keeps `PREDICATE_API_KEY` off the client and prevents SSRF via HTTPS enforcement and domain allowlist
- Attestation is fetched before fee estimation and transfer execution; errors surface as user-facing toasts rather than silent fallbacks
- Transfer status tracks a `FetchingAttestation` stage so users see progress

## How it works

1. On fee estimation and transfer, `fetchPredicateAttestation` calls `warpCore.isPredicateSupported(token, destination)` to detect predicate routes on-chain (traverses the hook tree for a `PredicateRouterWrapper`)
2. If required, a temp transaction is built to obtain calldata, which is posted to the Predicate API via the Next.js proxy to get an attestation
3. The attestation is passed to `getTransferRemoteTxs` / `estimateTransferRemoteFees`, which routes through `transferRemoteWithAttestation` on the wrapper

## Configuration

```
PREDICATE_API_KEY=your_predicate_api_key_here       # server-side only
PREDICATE_API_URL=https://api.predicate.io/v2/attestation  # optional override
```

## Notes

- Draft: depends on the SDK predicate implementation being merged (`warpCore.isPredicateSupported` is guarded with a typeof check for older SDKs)
- Non-predicate routes are unaffected — `isPredicateSupported` returns `false` and the existing flow runs unchanged


src/pages/api/predicate/attestation.ts
```diff
@@ -0,0 +1,37 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+        'x-api-key': PREDICATE_API_KEY,
+      },
+      body: JSON.stringify(req.body),
+    });
+
+    const data = await response.json();
+
+    if (!response.ok) {
```
paulbalaji: This forwards `req.body` directly to the Predicate API with no input validation. Any caller can send arbitrary JSON that gets relayed with the server's API key.

**Suggestion:** Validate `req.body` against a schema with the 5 expected fields (`to`, `from`, `data`, `msg_value`, `chain` -- all strings). Reject anything that doesn't match.

Also consider:
- Rate limiting per IP to prevent quota abuse
- Validating `PREDICATE_API_URL` is HTTPS + allowed domain to prevent SSRF
- Sanitizing error responses before forwarding (line 31 forwards the full Predicate error body to the client, which may leak internals)

src/features/transfer/useTokenTransfer.ts
```diff
@@ -170,11 +172,63 @@ async function executeTransfer({
...
+    try {
+      const needsAttestation = await warpCore.isPredicateSupported(originToken, destination);
+      logger.debug('Predicate support check:', { needsAttestation, origin, destination });
+
+      if (needsAttestation) {
+          logger.debug('Route requires Predicate attestation, fetching...');
+          updateTransferStatus(transferIndex, (transferStatus = TransferStatus.FetchingAttestation));
+
+          // Get wrapper address for attestation request
+          const adapter = originToken.getAdapter(multiProvider) as any;
```
paulbalaji: `as any` casts here (`adapter as any`, `tempTx as any`) will crash at runtime on non-EVM chains where `getPredicateWrapperAddress()` doesn't exist on the adapter.

Per the monorepo type safety guidelines, `as any` should be replaced with:
1. A type guard or predicate-specific adapter interface for `getPredicateWrapperAddress()`
2. Narrowing `tempTxs[0]` via the `type` discriminant (`ProviderType.EthersV5`) before accessing `.transaction.data`

Also, `wrapperAddress!` and `tempTx.transaction.data!` should be `assert()` calls:
```typescript
assert(wrapperAddress, 'Predicate wrapper address not found');
assert(tempTx.transaction.data, 'Transfer tx missing calldata');
```

src/features/transfer/useFeeQuotes.ts
```diff
@@ -103,12 +104,52 @@ async function fetchFeeQuotes(
   }
 
   const originTokenAmount = transferToken.amount(amountWei);
+
+  // Check if predicate attestation needed for fee estimation
+  let attestation: PredicateAttestation | undefined;
```
paulbalaji, line 148: This attestation fetch block is nearly identical to `useTokenTransfer.ts:175-230` and `TransferTokenForm.tsx:964-1003`.

Two concerns:
1. **DRY**: Extract into a shared function (e.g., `fetchPredicateAttestation()`).
2. **Error handling mismatch**: This logs a warning and continues without attestation, but `useTokenTransfer.ts` throws and blocks the transfer. If a predicate route proceeds to fee estimation without attestation, the estimate will be wrong (or the on-chain call will just revert). Consider either failing consistently or skipping the attestation fetch here entirely -- the fee estimation can likely proceed without it, or just use `isPredicateSupported()` as a lightweight signal.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -960,12 +961,52 @@ async function validateForm(
     }
 
     const originTokenAmount = transferToken.amount(amountWei);
+
+    // Check if predicate attestation needed for validation
+    let attestation: PredicateAttestation | undefined;
+    const predicateClient = getPredicateClient();
```
paulbalaji: Same DRY concern as the other two attestation fetch sites. This is the third copy of the same pattern.

For validation specifically: fetching a real attestation here seems unnecessary. The attestation will be discarded (a new one is fetched at transfer time). Consider replacing with just the `isPredicateSupported()` check to confirm the route requires attestation, without actually fetching one.

scripts/link-monorepo.js
```diff
@@ -145,6 +146,63 @@ args.forEach((folder) => {
   console.log(`      Location: ${relativePath}\n`);
 });
 
+/**
+ * Hardcoded: Pack @hyperlane-xyz/core from solidity directory
+ */
```
paulbalaji, line 154: This hardcoded block has a few issues:

1. **Always runs** even when the user only wants to link specific TS packages -- surprising behavior
2. **Duplicates** the existing pack-move-track pattern from the `args.forEach` loop above. Should extract a shared `packAndMove(packageDir)` function.
3. **Reads like a dev workaround** (the comment says "Hardcoded") rather than a permanent feature

**Suggestion:** Either generalize the script to accept `solidity` as a CLI argument, or split this into a separate PR.

paulbalaji (commented):
## Review Summary

Ran multiple review passes (security, code quality, script changes). The overall approach is sound -- server-side proxy for the API key, client-side singleton, status tracking in the transfer flow. But there are several issues to address before merge.

### Findings

| # | Finding | Severity |
|---|---------|----------|
| 1 | **No input validation on API proxy** -- forwards arbitrary `req.body` to Predicate API with server API key attached. Open relay risk. | HIGH |
| 2 | **`as any` casts** on adapter and transaction will crash on non-EVM chains. `getPredicateWrapperAddress()` doesn't exist on `ITokenAdapter`. | HIGH |
| 3 | **Non-null assertions** (`wrapperAddress!`, `tempTx.transaction.data!`) on nullable values -- runtime crash risk | HIGH |
| 4 | **Attestation logic duplicated 3x** with inconsistent error handling -- `useTokenTransfer` throws on failure (correct), but `useFeeQuotes` and `validateForm` silently continue without attestation (will cause late-stage revert) | HIGH |
| 5 | **Triple attestation fetch per transfer** -- validation, fee estimation, and execution each fetch a fresh attestation. Wastes API quota, adds latency, and the first two are discarded anyway | MEDIUM |
| 6 | **SSRF via `PREDICATE_API_URL`** env var -- if compromised, can point proxy at internal services | MEDIUM |
| 7 | **No rate limiting** on `/api/predicate/attestation` proxy -- anyone can abuse the API key | MEDIUM |
| 8 | **`link-monorepo.js` changes are unrelated** -- hardcoded solidity packing duplicates existing pack logic and runs unconditionally. Should be a separate PR. | MEDIUM |
| 9 | **Error responses forward Predicate API internals** to client | LOW |
| 10 | **API key properly server-side** -- `PREDICATE_API_KEY` never exposed to client bundle | OK |

### Recommendations (priority order)

1. **Extract** attestation logic into a shared function (e.g., `fetchPredicateAttestation()` in `src/features/transfer/predicate.ts`). Fixes DRY, inconsistent error handling, and `as any` in one place.
2. **Fetch attestation once** at transfer execution time only. Remove from `validateForm` and `useFeeQuotes` -- use `isPredicateSupported()` as a lightweight check instead.
3. **Add input validation** on the API proxy route (Zod schema for the 5 expected fields).
4. **Replace `as any`** with type guards or a predicate-specific adapter interface.
5. **Replace `!` assertions** with `assert()` calls per codebase convention.
6. **Split** `link-monorepo.js` changes into a separate PR.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -960,12 +961,52 @@ async function validateForm(
...
+
+    try {
+      const needsAttestation = await warpCore.isPredicateSupported(transferToken, destination);
+
+      if (needsAttestation) {
+        // Get wrapper address for attestation request
+        const adapter = transferToken.getAdapter(warpCore.multiProvider) as any;
```
Xaroz: what kind of token standard do we expect here? Ideally we don't just set `any`

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -960,12 +961,52 @@ async function validateForm(
...
+          originTokenAmount,
+          destination,
+          sender: sender || '',
+          recipient,
+          senderPubKey: await senderPubKey,
+        });
+        const tempTx = tempTxs[0] as any;
+
+        const attestationRequest = {
+          to: wrapperAddress!,
```
Xaroz: I think you should create an interface/ type for the attestation body request, and probably handle undefined or null values instead of non-null asserting here

src/features/transfer/useTokenTransfer.ts
```diff
@@ -170,11 +172,63 @@ async function executeTransfer({
 
     updateTransferStatus(transferIndex, (transferStatus = TransferStatus.CreatingTxs));
 
+    // Check if Predicate attestation is needed
+    let attestation: PredicateAttestation | undefined;
+    const predicateClient = getPredicateClient();
+
+    try {
+      const needsAttestation = await warpCore.isPredicateSupported(originToken, destination);
```
Xaroz: minor nit: one thing to note is that this just calls `getPredicateWrapperAddress` so you are essentially calling it twice, once here and once inside the `if`

src/lib/predicateClient.ts
```diff
@@ -0,0 +1,58 @@
...
+    uuid: string;
+    [key: string]: any;
+  };
+}
+
+/**
+ * Proxy client for Predicate API that routes through Next.js API route
+ * to avoid CORS issues
+ */
+class PredicateProxyClient {
```
Xaroz: not sure if we need a class here and we can just do functional programming as we don't really use classes inside the warp-ui itself

src/pages/api/predicate/attestation.ts
troykessler: added validation

src/features/transfer/useTokenTransfer.ts
troykessler: fixed

src/features/transfer/TransferTokenForm.tsx
troykessler: fixed

scripts/link-monorepo.js
troykessler: it is intended that it always should run, tested and verified it with the new predicate contract changes

src/features/transfer/TransferTokenForm.tsx
troykessler: fixed

src/features/transfer/TransferTokenForm.tsx
troykessler: done

src/features/transfer/useTokenTransfer.ts
troykessler: fixed

src/lib/predicateClient.ts
troykessler: changed it now

src/features/transfer/predicate.ts
```diff
@@ -0,0 +1,102 @@
...
+  const tempTxs = await warpCore.getTransferRemoteTxs({
+    originTokenAmount: amount,
+    destination,
+    sender,
+    recipient,
+    destinationToken,
+  });
+
+  assert(tempTxs.length > 0, 'No transactions returned for transfer');
+
```
paulbalaji, line 96: **🔴 High — attested calldata diverges from submitted calldata.**

This builds `tempTxs` *without* an `attestation` to extract calldata for the Predicate request. But downstream `useTokenTransfer.ts:193-200` and `useFeeQuotes.ts:151-179` call `getTransferRemoteTxs` again with the fetched `attestation`, and in upstream `WarpCore.getTransferRemoteTxs` the `attestation` is threaded into `hypAdapter.populateTransferRemoteTx(...)` — producing different calldata than what was attested.

For a compliance-gated route, the on-chain verifier will see calldata the attestation doesn't cover. Either:
- Build `tempTxs` once, attest over the actual transfer-tx calldata, then submit those same tempTxs (don't rebuild with attestation), or
- Expose (on the SDK side) a `populateForAttestation` primitive that returns the exact calldata that will ultimately be signed — attestation included — so round-trip attestation-in-calldata is coherent.

src/features/transfer/predicate.ts
```diff
@@ -0,0 +1,102 @@
...
+  assert(wrapperAddress, 'Predicate wrapper address not found');
+
+  // Build temporary transaction to get calldata
+  const tempTxs = await warpCore.getTransferRemoteTxs({
+    originTokenAmount: amount,
+    destination,
+    sender,
+    recipient,
+    destinationToken,
+  });
```
paulbalaji, line 93: **🔴 High — wrong index: `tempTxs[0]` is the approval, not the transfer, on approve-required routes.**

Confirmed against `hyperlane-monorepo/typescript/sdk/src/warp/WarpCore.ts:553-600`: `preTransferRemoteTxs` (optional revoke + approval) are pushed onto `transactions` *before* the transfer tx. Whenever `isApproveRequired` is true (most first-time ERC20 routes, plus USDT-style revoke+approve), `tempTxs[0]` is an approval and `tempTxs.length - 1` is the transfer.

Use `tempTxs[tempTxs.length - 1]` (or filter by `category === WarpTxCategory.Transfer`). Also consider gating this function on `token.protocol === ProtocolType.Ethereum` — the `(tempTx as any).transaction.data` shape assumes EVM; non-EVM adapters would break here.

src/features/transfer/predicate.ts
```diff
@@ -0,0 +1,102 @@
...
+  token,
+  destination,
+  sender,
+  recipient,
+  amount,
+  destinationToken,
+}: FetchAttestationParams): Promise<PredicateAttestation | undefined> {
+  const warpCoreAny = warpCore as any;
+  const needsAttestation: boolean =
+    typeof warpCoreAny.isPredicateSupported === 'function'
```
paulbalaji: `warpCore as any` + `PredicateAttestation = any` are now avoidable — SDK 33 (bumped in this same PR) has real types and `WarpCore.isPredicateSupported`. Suggest importing `PredicateAttestation` from `@hyperlane-xyz/sdk` (or wherever #8224 landed it) and dropping all three `as any` casts in this file; the last one at `:48` (`token.getAdapter(... as any) as any`) should become a typed call plus a narrow type guard on the capability.

src/lib/predicateClient.ts
```diff
@@ -0,0 +1,37 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(request),
+  });
+
+  if (!response.ok) {
+    const error = await response.json();
+    throw new Error(error.error || 'Failed to fetch attestation');
+  }
```
paulbalaji: **🟠 Medium — bypasses the SDK's attestation validation.**

Upstream `PredicateApiClient` (`hyperlane-monorepo/typescript/sdk/src/predicate/PredicateApiClient.ts:85-100`) does three things this client skips:
1. `PredicateAttestationResponseSchema.parse(...)` — shape validation via Zod.
2. Rejects responses where `is_compliant === false`.
3. Rejects responses where `attestation.expiration <= now`.

Returning raw `response.attestation` means an expired or non-compliant payload can flow into `warpCore.validateTransfer` / `estimateTransferRemoteFees` / `getTransferRemoteTxs`, where the SDK won't re-verify. Either mirror those three checks here, or (cleaner) call `PredicateApiClient` directly in the Next.js API route on the server — it's a pure-TS class with light deps. A narrow SDK subpath would make the latter option clean.

src/lib/predicateClient.ts
```diff
@@ -0,0 +1,37 @@
...
+  data: string;
+  msg_value: string;
+  chain: string;
+}
+
+export interface AttestationResponse {
+  attestation: {
+    uuid: string;
+    [key: string]: any;
+  };
```
paulbalaji: `attestation: { uuid: string; [key: string]: any }` is effectively `any`. Import `PredicateAttestation` (or the response type) from the SDK now that 33.0.0 is the pinned version in `package.json:49-51`.

src/features/transfer/useFeeQuotes.ts
```diff
@@ -154,6 +175,7 @@ export async function fetchFeeQuotes(
       destination,
       sender: EVM_FEE_QUOTE_FALLBACK_ADDRESS,
       recipient: recipient,
+      attestation,
       destinationToken: connectedDestinationToken,
```
paulbalaji, line 218: **🔴 High — attestation reused across sender change.**

The attestation fetched at `:140-150` is signed against calldata for the connected `sender`. When this fallback path retries with `sender: EVM_FEE_QUOTE_FALLBACK_ADDRESS` (`0x…dEaD`), the same `attestation` is passed through. Fee estimation on an RPC that actually checks the attestation against the tx's `from` will either silently drift (stale estimate) or fail.

Either: (a) skip `attestation` entirely in the fallback retry (fees are only for display; the real submit path fetches a fresh one), or (b) refetch the attestation for the fallback sender — noting this compounds the divergence issue flagged in `predicate.ts`.

src/features/transfer/useFeeQuotes.ts
```diff
@@ -124,13 +126,31 @@ export async function fetchFeeQuotes(
...
+      warpCore,
+      token: transferToken,
+      destination,
+      sender,
+      recipient,
+      amount: originTokenAmount,
+      destinationToken: connectedDestinationToken,
+    });
+  } catch (error: any) {
+    toast.error('Compliance verification failed. Unable to estimate fees.');
```
paulbalaji: Prior reviews (mine + CodeRabbit) flagged that `useFeeQuotes` refetches an attestation every 30s, which both burns quota and adds a full `getTransferRemoteTxs` call to the quote polling path. Still present. A lightweight `warpCore.isPredicateSupported(...)` gate at the top of `fetchFeeQuotes` to *skip* attestation entirely in the quote path (fees are display-only; submit path refetches) would sidestep both problems.

src/pages/api/predicate/attestation.ts
```diff
@@ -0,0 +1,82 @@
...
+      )
+    );
+  } catch {
+    return false;
+  }
+}
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  if (req.method !== 'POST') {
+    return res.status(405).json({ error: 'Method not allowed' });
```
paulbalaji, line 31: **🟠 Medium — still an unauthenticated public proxy.**

SSRF hostname hardening here is correct. But anyone who can reach the deployment can still POST arbitrary `{to, from, data, msg_value, chain}` triples and spend the operator's Predicate quota. Add one or more of: per-IP rate-limit (e.g., Vercel edge middleware or `next-rate-limit`), `Origin` allowlist, or an auth-session gate. `Origin` alone isn't sufficient against scripted callers — prefer a rate-limit as the primary control. Unresolved across two prior reviews.

src/pages/api/predicate/attestation.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    const responseData = await response.json();
+
+    if (!response.ok) {
+      // Sanitize error response - don't leak Predicate API internals
+      console.error('Predicate API error:', response.status, responseData);
+      return res.status(response.status >= 500 ? 502 : response.status).json({
+        error: 'Failed to fetch attestation',
+      });
+    }
```
paulbalaji: `response.status >= 500 ? 502 : response.status` passes through 4xx codes from the Predicate API verbatim. Minor, but it lets a caller distinguish 400 (bad shape) from 401/403 (auth/allowlist) states at the operator's expense. Consider normalizing non-2xx to a single 502 with a generic error body. Low-severity.

paulbalaji (commented):
## Follow-up review on head `b66b60a`

Re-reviewed after the force-push. Cross-checked my prior findings, CodeRabbit's, and a codex pass — consolidating below.

### Fixed since the last round
- Typecheck, tests, lint, format all green (SDK bumped to `33.0.0` in `package.json:49-51`, `useFeeQuotes.test.ts` stubs `isPredicateSupported`).
- `seismiclocal` dev chain + warp route removed from `chains.yaml` / `warpRoutes.yaml`.
- `validateForm` no longer silently swallows attestation errors — it now throws into the outer catch.
- SSRF hostname allowlist tightened (`api.predicate.io` / `predicate.io` exact-or-subdomain).
- `assert()` used for `wrapperAddress` and `txData` — non-null assertions gone.
- Input validation (5-field string check) added to the proxy.

### Still unresolved

**🔴 High — attested calldata ≠ submitted calldata.** `predicate.ts:65-73` builds a provisional `getTransferRemoteTxs(...)` *without* an attestation to extract calldata, gets an attestation signed against that calldata, then `useTokenTransfer.ts:193-200` and `useFeeQuotes.ts:151-179` rebuild the transfer *with* the attestation. In `WarpCore.getTransferRemoteTxs` the `attestation` flows into `hypAdapter.populateTransferRemoteTx(...)`, so the final calldata differs from the one that was attested. For a Predicate-gated route this is a protocol-level mismatch — the on-chain verifier receives calldata the attestation doesn't cover.

**🔴 High — `tempTxs[0]` is the wrong tx for approve/revoke routes.** `predicate.ts:70`. Verified upstream in `hyperlane-monorepo/typescript/sdk/src/warp/WarpCore.ts:553-600`: `preTransferRemoteTxs` (revoke + approval) are pushed to `transactions` **before** the transfer tx. So `tempTxs[0]` is the approval, not the transfer, any time `isApproveRequired` is true. The attestation is then requested for approval calldata.

**🔴 High — fee-quote fallback reuses an attestation bound to the wrong sender.** `useFeeQuotes.ts:140-179`. Attestation is fetched once with the real `sender`, then on failure the retry re-estimates with `sender: EVM_FEE_QUOTE_FALLBACK_ADDRESS` (`0x…dEaD`) while passing the *same* attestation through. The attestation was signed for the real sender's calldata.

**🟠 Medium — client bypasses SDK's attestation validation.** `predicateClient.ts` just checks `response.ok` and returns `response.attestation`. The upstream SDK (`hyperlane-monorepo/typescript/sdk/src/predicate/PredicateApiClient.ts:85-100`) parses the response through a Zod schema, rejects `!is_compliant`, and rejects expired attestations. UI skips all three — non-compliant or expired responses propagate into `warpCore.validateTransfer` / `estimateTransferRemoteFees` / `getTransferRemoteTxs`.

**🟠 Medium — proxy still unauthenticated.** Covered by my prior review and CodeRabbit; SSRF allowlist fix is great, but absent rate-limit / origin / auth, anyone reaching the deployment can drain `PREDICATE_API_KEY` quota.

### Dependency / bundle impact

- **Intentional dep change**: `@hyperlane-xyz/sdk|utils|widgets` 32.0.0 → 33.0.0.
- The rest of the ~14k-line `pnpm-lock.yaml` diff is incidental caret re-resolution (Next 16.2.1 → 16.2.3, React 19.2.4 → 19.2.5, Playwright 1.58 → 1.59, metro patch bumps). Not malicious, but noisy and will conflict with #1030. Consider `pnpm install --lockfile-only --prefer-frozen-lockfile` from `main`'s lock so the diff reflects only the intentional 33.0.0 bump.
- Local build comparison (codex ran it): PR branch 104s vs main 140s; emitted `.next/static/chunks` basically flat (~315 MB both sides). The ~36s build-time win is almost certainly from the Next/React bump, not from the predicate feature itself.
- **Tree-shaking note (informational, not blocking this PR)**: `@hyperlane-xyz/sdk` is `sideEffects: false` and #8224 landed `WarpCore.isPredicateSupported` / `EvmTokenAdapter.getPredicateWrapperAddress` on the existing `./warp/*` and `./token/*` subpath entries — so those are tree-shakeable. However #8224 did **not** add a `./predicate` subpath, did **not** re-export `PredicateAttestation` types from `./types`, and still re-exports `PredicateWrapperDeployer` (deploy-only, pulls typechain) from the root barrel (`typescript/sdk/src/index.ts:902-912`). Root-barrel consumers therefore widen the module graph with deploy baggage. Warp-ui #1030 plus a monorepo follow-up (narrow `./predicate/PredicateApiClient` subpath + drop deployer from root barrel + re-export attestation types from `./types`) would close the loop. This PR currently uses root-barrel imports in `predicate.ts:1-3` / `useFeeQuotes.ts:1` / `useTokenTransfer.ts:1-7` — aligned with current convention but a re-touch target for #1030.

### CI

- `typecheck`, `lint`, `test`, `format`, `build`, `security-review` all pass.
- `e2e` pending / `code-review` + `interactive` skipping at review time.

### Suggested minimum pre-merge

1. Consolidate attestation to **one** `getTransferRemoteTxs` call whose outputs are both attested and submitted — e.g. build tempTxs once, fetch attestation using the tempTxs calldata (on the transfer tx, i.e. the last one), then submit those same tempTxs. If `attestation` actually changes the calldata, then redesign: the Predicate API needs to attest over calldata that already includes a pre-committed attestation field, or the adapter needs to expose a `populateForAttestation` primitive that returns the calldata that will ultimately be submitted.
2. Move the attestation fetch **out** of `useFeeQuotes` entirely; at minimum, don't reuse an attestation across the primary-sender → fallback-sender retry.
3. Mirror the SDK's validation in `predicateClient.ts` (or have the UI just delegate to `PredicateApiClient` once it's exposed via a narrow subpath).
4. Add a rate-limit on the proxy route (unresolved from prior reviews).
5. Revert the unintentional lockfile drift to reduce merge friction with #1030.

### Observations outside this PR (monorepo follow-up)
- `typescript/sdk/package.json` — add narrow `./predicate/PredicateApiClient` subpath (single file, not wildcard, to avoid re-exposing `PredicateDeployer`).
- `typescript/sdk/src/index.ts:909-912` — drop `PredicateWrapperDeployer` / `PredicateWrapperDeploymentResult` from root barrel; expose via a separate `./predicate/PredicateDeployer` subpath for CLI.
- `typescript/sdk/src/types.ts` — `export type { PredicateAttestation, PredicateAttestationRequest, PredicateAttestationResponse }` so readonly consumers can import types-only via `sdk/types`.

package.json
paulbalaji, line 1: most of this diff feels quite unnecessary since it's just reordering things, can we trim this

src/pages/api/predicate/attestation.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    return res.status(200).json(responseData);
+  } catch (error) {
+    console.error('Predicate API proxy error:', error);
```
paulbalaji: theres some eslint warnings from gh itself, maybe worth looking at

src/features/transfer/predicate.ts
```diff
@@ -0,0 +1,102 @@
+import type { IToken, Token, WarpCore } from '@hyperlane-xyz/sdk';
+import { TokenAmount } from '@hyperlane-xyz/sdk';
+import { assert } from '@hyperlane-xyz/utils';
+
+import { fetchAttestation as fetchAttestationFromProxy } from '../../lib/predicateClient';
+import { logger } from '../../utils/logger';
+
+// TODO: Import from SDK when available
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type PredicateAttestation = any;
```
paulbalaji: old todo / unwanted any ?

paulbalaji:
Great clarification, thanks for pushing back. I dug into the Solidity and you're right that it's not exact outer-calldata match — and codex followed up with a sharper description of the model, which is worth being precise about. Updating this comment in place.

## Verifier model — exact reconstructed inner statement

Not exact outer tx bytes, and not loose semantic params either. The wrapper signs an **exact reconstructed inner Statement**: `(encodedSigAndArgs, msgSender, target, msgValue, policyId, uuid, expiration)`.

Evidence:

```solidity
// hyperlane-monorepo/solidity/contracts/token/libs/AbstractPredicateWrapper.sol:152-163
bytes memory encodedSigAndArgs = abi.encodeWithSelector(
    ITokenBridge.transferRemote.selector,
    _destination, _recipient, _amount
);
// then in _executeAttested(...):207-214 →
// _authorizeTransaction(_attestation, encodedSigAndArgs, msg.sender, msg.value)
```

```solidity
// solidity/dependencies/@predicate-contracts-2.2.2/src/mixins/PredicateClient.sol:184-200
// builds Statement { msgSender, target, msgValue, encodedSigAndArgs, policyId, uuid, expiration }
// fields defined at @predicate-contracts-2.2.2/src/interfaces/IPredicateRegistry.sol:11-26
```

So the wrapper never signs the outer `transferRemoteWithAttestation(...)` calldata. The "circular dependency" on outer calldata isn't real — my H1 as written is withdrawn.

## What still survives — because the Statement binds more than just `data`

### Still ⛔ H (`predicate.ts:63-83`, tempTxs[0] bug)

For approve-required routes `tempTxs[0]` is the approval. The client then attests over `approve(spender, amount)` calldata; the wrapper reconstructs `transferRemote(dest, recipient, amount)` calldata. Different `encodedSigAndArgs` → different Statement preimage → signer recovery fails → revert. Fix: `tempTxs[tempTxs.length - 1]` or filter by `category === WarpTxCategory.Transfer`.

### Still ⛔ H (`useFeeQuotes.ts:140-179`, fallback sender)

The Statement binds `msgSender`. The retry path estimates with `sender: EVM_FEE_QUOTE_FALLBACK_ADDRESS` while reusing the attestation signed for the real sender. On-chain reconstruction would use `msg.sender = 0x…dEaD` → Statement hash differs → verification fails. The fallback cannot produce a quote for a Predicate-gated route. Fix: refetch for the fallback sender, or drop attestation from the fallback estimate entirely (fees are display-only; submit path refetches).

## Three subtle encoding requirements (any divergence = on-chain revert)

Because the Statement is a canonical reconstruction, the client-side `data`/`from`/`to`/`msg_value` sent to the Predicate API must reconstruct byte-for-byte identically to what the wrapper builds:

1. **Selector overload.** Wrapper uses 3-arg `ITokenBridge.transferRemote.selector` (`transferRemote(uint32,bytes32,uint256)`). `TokenRouter` has 5-arg overloads with hook/metadata — if `getTransferRemoteTxs` ever emits the 5-arg one, selectors diverge and the attestation won't verify. Worth asserting.
2. **Cross-collateral routes encode a *different* inner call.** `PredicateCrossCollateralRouterWrapper.sol:101-107` reconstructs `CrossCollateralRouter.transferRemoteTo.selector` with an extra `targetRouter` arg:
   ```solidity
   abi.encodeWithSelector(
     CrossCollateralRouter.transferRemoteTo.selector,
     destination, recipient, amount, targetRouter
   )
   ```
   The UI's `predicate.ts` doesn't branch on wrapper type today — it pulls `tempTx.transaction.data` verbatim from `getTransferRemoteTxs`. If for cross-collateral routes the tempTx calldata isn't `transferRemoteTo(dest, recipient, amount, targetRouter)` in that exact shape, the Statement won't match. Needs verification against what the `PredicateCrossCollateralRouter` adapter actually populates.
3. **`msg_value` binding.** Wrapper authorizes against outer `msg.value`, not inner. UI currently passes `transaction.value` from the provisional inner tx (`predicate.ts:91-98`) — that value must be identical to what the user will send to `transferRemoteWithAttestation` (≈ interchain-gas quote), and the client currently has no asserting guarantee of that.

## Suggested direction

No SDK primitive needed to break an outer-calldata circular dependency — that problem doesn't exist. But an SDK helper that returns the **canonical attested inner Statement inputs** (`{ encodedSigAndArgs, target, msgValue }` as the wrapper will reconstruct them, branched by wrapper kind) would be the cleanest fix. The UI is deriving these indirectly from provisional `tempTxs` today — which is the proximate cause of both surviving bugs and the three encoding risks above. Something like `adapter.getAttestationStatementInputs({ amount, destination, recipient, ... })` on the Predicate-capable adapter would let the UI hand the API exactly what the wrapper will sign against.

## Unchanged from prior review

- **🟠 Medium** — `predicateClient.ts` skips the SDK's Zod parse + `is_compliant` + expiry checks (`PredicateApiClient.ts:85-100`).
- **🟠 Medium** — proxy still unauthenticated / no rate-limit.

## Net for merge

1. Fix `tempTxs[0]` → use the transfer tx, and source `data` from adapter-canonical inputs (once available) rather than provisional tx bytes.
2. Fix fee-quote fallback attestation reuse.
3. Verify the cross-collateral path emits `transferRemoteTo(...)` calldata matching the wrapper's reconstruction.
4. Mirror `is_compliant` + expiry in `predicateClient.ts`.
5. Rate-limit the proxy.

cc @troykessler

src/pages/api/predicate/attestation.ts
troykessler: took care of the warnings

src/features/transfer/predicate.ts
troykessler: this was already flagged by the ai-review, this was already removed

src/features/transfer/predicate.ts
```diff
@@ -0,0 +1,104 @@
...
+    from: sender,
+    chain: token.chainName,
+  });
+
+  // Fetch attestation from Predicate API via proxy
+  const response = await fetchAttestationFromProxy({
+    to: wrapperAddress,
+    from: sender,
+    data: txData,
+    msg_value: transaction.value?.toString() || '0',
```
paulbalaji, line 126: **🟠 Medium — IGP-drift exposure on `msg_value` binding.**

`tempTxs` is built here without a pinned IGP quote. At line 98 the attestation is bound to `transaction.value?.toString()` from that tempTx. Downstream, `useTokenTransfer.ts:196-203` calls `getTransferRemoteTxs` again and `useFeeQuotes.ts:167-173` calls `estimateTransferRemoteFees` again — both re-quote IGP. The with-attestation and no-attestation branches of `EvmTokenAdapter.populateTransferRemoteTx` (`hyperlane-monorepo typescript/sdk/src/token/adapters/EvmTokenAdapter.ts:510-600`) derive `.value` from the same `igpQuote + optional tokenFeeQuote` formula, so this is correct *by construction* — but exposed to any IGP change between the two calls.

`AbstractPredicateWrapper._executeAttested` (`solidity/contracts/token/libs/AbstractPredicateWrapper.sol:198-243`) hashes outer `msg.value` into the Statement preimage, so a drifted fee → `_authorizeTransaction` revert.

Fix: capture the `interchainGas` quote produced during the attestation fetch and pass it into the submit and fee-estimation calls so the same `nativeValue` is used throughout.

src/features/transfer/predicate.ts
```diff
@@ -0,0 +1,104 @@
...
+  assert(tempTxs.length > 0, 'No transactions returned for transfer');
+
+  // Use last tx: preTransferRemoteTxs (approval/revoke) are prepended before the transfer tx
+  const tempTx =
+    tempTxs.find((tx) => (tx as any).category === WarpTxCategory.Transfer) ??
+    tempTxs[tempTxs.length - 1];
+  assert(tempTx && typeof tempTx === 'object', 'Invalid transaction object');
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const transaction = (tempTx as any).transaction;
```
paulbalaji: Two stray `as any` casts: `(tx as any).category` on :76 and `(tempTx as any).transaction` on :81. `WarpTypedTransaction` has both fields, so typed access should work and would remove the `eslint-disable-next-line` on :80. Since we import `WarpTxCategory` and the call site is already Predicate-specific (gated on `ProtocolType.Ethereum`), a narrow type annotation or assertion (`tempTx as WarpTypedTransaction`) is cleaner than silencing the linter.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -171,11 +173,32 @@ async function executeTransfer({
...
+      logger.error('Predicate attestation error:', error);
+      throw new Error(`Predicate attestation failed: ${message}`);
+    }
+
     const txs = await warpCore.getTransferRemoteTxs({
       originTokenAmount,
       destination,
       sender,
       recipient,
+      attestation,
```
paulbalaji: This is the second `getTransferRemoteTxs` call. The tempTxs built in `predicate.ts:64-70` to bind the attestation to `msg_value` are thrown away and rebuilt here — meaning any IGP drift between the two calls drifts the `.value` away from what the attestation was bound to. Passing the `interchainGas` quote from the first call through here (and into `useFeeQuotes.ts` likewise) would pin the value. See the thread on `predicate.ts:64-98`.

src/features/transfer/useFeeQuotes.ts
```diff
@@ -139,6 +170,7 @@ export async function fetchFeeQuotes(
       sender,
       senderPubKey: senderPubKeyValue,
       recipient: recipient,
```
paulbalaji, line 199: Same IGP-drift concern as `useTokenTransfer.ts:201`: the attestation was bound to the tempTx's `value` inside `fetchPredicateAttestation`, and this `estimateTransferRemoteFees` call re-quotes IGP. For a predicate-gated route, a drifted fee here will either make fee display flaky (revert) or, if the drifted value is also used at submit, fail the on-chain attestation check. Threading the `interchainGas` quote through addresses both.

src/pages/api/predicate/attestation.ts
```diff
@@ -0,0 +1,87 @@
...
+const ipRequestLog = new Map<string, number[]>();
+
+function isRateLimited(ip: string): boolean {
+  const now = Date.now();
+  const timestamps = (ipRequestLog.get(ip) ?? []).filter((t) => now - t < RATE_LIMIT_WINDOW_MS);
+  if (timestamps.length >= RATE_LIMIT_MAX) return true;
+  timestamps.push(now);
+  ipRequestLog.set(ip, timestamps);
+  return false;
+}
```
paulbalaji: **🟡 Low — rate-limiter is a speed bump, not a control.**

Three issues worth noting:

1. **Per-instance state.** The `ipRequestLog` Map is process-local; on Vercel serverless each warm instance has its own 10-req/60s budget. Scale-out multiplies the effective quota.
2. **Spoofable IP header.** `x-real-ip` is trusted without validating that the upstream is a known proxy. A direct caller can set that header freely.
3. **Unbounded growth.** Old timestamps get filtered on read but the Map entry itself is never removed — one entry per seen IP indefinitely.

Fine as a mitigation; inadequate if Predicate quota protection is load-bearing. Consider backing with Vercel KV / Redis and adding sweep logic.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -951,12 +952,24 @@ async function validateForm(
...
     const originTokenAmount = transferToken.amount(amountWei);
+
+    const attestation = await fetchPredicateAttestation({
+      warpCore,
+      token: transferToken,
+      destination,
+      sender: sender || '',
+      recipient,
+      amount: originTokenAmount,
+      destinationToken: connectedDestinationToken,
```
paulbalaji: Validate-form now fetches a full Predicate attestation on every debounced form change, passes it into `validateTransfer`, and discards the result. For a predicate-gated route this doubles the attestation API calls per transfer (validate + submit, each with their own `getTransferRemoteTxs` inside). A lightweight gate on `warpCore.isPredicateSupported(...)` (skipping the fetch and letting `validateTransfer` run without attestation), or deferring attestation to the submit path only, would halve the API spend and tighten the IGP-drift window.

paulbalaji (commented):
## Re-review at head `316c5855`

Cross-checked against codex. The earlier review queue has cleared — nothing previously flagged remains open. Two smaller items surface on this pass.

### Previously flagged, now fixed (not re-raising)

- `tempTxs[0]` wrong index → fixed with `find(category === WarpTxCategory.Transfer) ?? tempTxs[tempTxs.length - 1]` in `src/features/transfer/predicate.ts:75-77`.
- Fee-quote fallback reusing real-sender attestation → fixed: predicate routes with fallback-only sender return `null` (no fees); the retry path passes `attestation: undefined` (`src/features/transfer/useFeeQuotes.ts:140-162, 189`).
- Client-side skipping SDK's Zod / `is_compliant` / expiry validation → proxy now instantiates `PredicateApiClient` server-side (`src/pages/api/predicate/attestation.ts:82-84`), so those checks run.
- `as any` on adapter / WarpCore / attestation type → replaced with `isPredicateCapableAdapter(...)`, `token.getHypAdapter(...)`, and the real `PredicateAttestation` type throughout `predicate.ts`, `useFeeQuotes.ts`, `useTokenTransfer.ts`, `predicateClient.ts`.
- Proxy input validation / SSRF hostname hardening / dev-fixture chain removal / prior lint/type/test issues → all addressed in the fix commits.

### Open — worth addressing before merge

**🟠 Medium — attestation `msg_value` binds to a provisional quote that is later re-quoted.**

`src/features/transfer/predicate.ts:64-70` builds `tempTxs` via `warpCore.getTransferRemoteTxs(...)` and at line 98 binds the attestation to `transaction.value` from that tempTx. The real execution paths (`useTokenTransfer.ts:196-203` and `useFeeQuotes.ts:167-173`) call `getTransferRemoteTxs` / `estimateTransferRemoteFees` *again*, and both re-quote IGP when no quote is pinned. Verified in `hyperlane-monorepo/typescript/sdk/src/token/adapters/EvmTokenAdapter.ts:510-600`: the with-attestation and no-attestation branches derive `.value` from the *same* `nativeValue = igpQuote + optional tokenFeeQuote` formula, so the binding is right *in principle*; the risk is entirely IGP drift between the two calls.

On a congested origin chain, the inner IGP fee at attestation-fetch time can differ from the IGP fee at submit time. `AbstractPredicateWrapper._executeAttested` (monorepo `solidity/contracts/token/libs/AbstractPredicateWrapper.sol:198-243`) hashes the outer `msg.value` into the Statement preimage, so any drift → on-chain `_authorizeTransaction` revert.

Fix: cache the `interchainGas` quote produced during the attestation fetch and pass it through to both the submit path's `getTransferRemoteTxs` and the fee path's `estimateTransferRemoteFees` so the same `nativeValue` is used everywhere.

**🟡 Low — rate-limiter is a speed bump, not a control.**

`src/pages/api/predicate/attestation.ts:26-39`. Three issues:
1. Process-local `Map` — Vercel serverless scale-out multiplies the 10/min limit by the number of warm instances.
2. Trusts `x-real-ip` without validating upstream is a known proxy (spoofable by a direct caller).
3. `ipRequestLog` never deletes entries — one entry per seen IP indefinitely is an unbounded memory footprint.

Fine as mitigation. If Predicate quota abuse matters in production, move to shared storage (Vercel KV / Redis) and add sweep logic.

### Also worth a look

- **Validate-form fetches attestation to discard it.** `src/features/transfer/TransferTokenForm.tsx:955-963` now calls `fetchPredicateAttestation` on every debounced form change; the attestation is passed to `validateTransfer` and the result is immediately dropped. On a Predicate-gated route this doubles the API calls per transfer (validate + submit). Gating on `warpCore.isPredicateSupported(...)` and skipping the fetch during validation, or reusing the submit-time attestation, would cut this in half.
- **Two stray `as any` in `predicate.ts:76, 81`.** `WarpTypedTransaction` has `category` and `transaction` fields; typed access would work and would remove the `eslint-disable-next-line` on `:80`.
- **`package.json` missing trailing newline** (`\ No newline at end of file`). Trivial.
- **`pnpm-lock.yaml` churn not reverted** — still 14k lines, dominated by Next 16.2.1 → 16.2.3 / React 19.2.4 → 19.2.5 / Playwright 1.58 → 1.59 caret re-resolution, not by the intentional 32→33 Hyperlane bump. Will still collide with #1030. Rerunning `pnpm install --lockfile-only --prefer-frozen-lockfile` against main's lock would leave only the intentional change in the diff.

### Deps / perf

Direct dep delta is just `@hyperlane-xyz/sdk|utils|widgets 32.0.0 → 33.0.0` (`package.json:23-25`). Emitted client bundle is flat vs `main` (`.next/static/chunks` within kilobytes across multiple runs). Local build-time numbers vary run-to-run — I wouldn't read anything into them; the machine was running multiple agents in parallel and the timing is noise. No perf regression attributable to the predicate feature.

### Verified end-to-end correctness

One thing I wanted to double-check given the earlier discussion: whether cross-collateral Predicate routes encode the right selector. `WarpCore.getTransferRemoteTxs` dispatches on `isCrossCollateralTransfer(...)` *before* the attestation flag is considered (`typescript/sdk/src/warp/WarpCore.ts:494`), routing to `getCrossCollateralTransferTxs` → `adapter.populateTransferRemoteToTx` → `transferRemoteTo(dest, recipient, amount, targetRouter)` — which matches `PredicateCrossCollateralRouterWrapper.sol:101-107`'s reconstruction. Both wrapper variants are wired up correctly end-to-end.

### Net

No blockers. IGP-drift fix is the sharpest of the remaining items and is a small change. The rest are polish.

scripts/link-monorepo.js
```diff
@@ -157,6 +158,63 @@ args.forEach((folder) => {
...
+console.log('📦 Packing hardcoded package: @hyperlane-xyz/core (from solidity/)');
+
+const solidityCorePath = SOLIDITY_DIR;
+if (fs.existsSync(solidityCorePath)) {
```
Xaroz: this looks similar to the function above no? no way to group it, refactor it?

src/features/transfer/useFeeQuotes.ts
```diff
@@ -124,13 +131,37 @@ export async function fetchFeeQuotes(
...
   if (!connectedDestinationToken) return null;
   const isEvmToEvmRoute =
     originToken.protocol === ProtocolType.Ethereum &&
     destinationToken.protocol === ProtocolType.Ethereum;
   const senderPubKeyValue = await senderPubKey;
 
+  // Predicate routes require the wrapper path (transferRemoteWithAttestation) — estimating
+  // fees without a valid attestation reverts on-chain. But attestation needs a real sender
+  // address; skip fee display when only the fallback address is available.
+  let attestation: PredicateAttestation | undefined;
```
Xaroz, line 148: can we add tests for this new branch?

src/pages/api/predicate/attestation.ts
```diff
@@ -0,0 +1,87 @@
...
+const RATE_LIMIT_MAX = 10;
+const ipRequestLog = new Map<string, number[]>();
+
+function isRateLimited(ip: string): boolean {
+  const now = Date.now();
+  const timestamps = (ipRequestLog.get(ip) ?? []).filter((t) => now - t < RATE_LIMIT_WINDOW_MS);
+  if (timestamps.length >= RATE_LIMIT_MAX) return true;
+  timestamps.push(now);
+  ipRequestLog.set(ip, timestamps);
+  return false;
```
Xaroz: instead of doing a custom rate limiting, you can set a custom rule in Vercel `firewall/rules` to do the rate limiting

also from what I remember mapping/storage might not work the way you think it does, because these are serverless function which create new instances every time it calls

src/features/transfer/useFeeQuotes.ts
troykessler: already addressed

src/pages/api/predicate/attestation.ts
troykessler: moving this to vercel

src/features/transfer/useFeeQuotes.ts
troykessler: added tests

src/features/transfer/useFeeQuotes.ts
```diff
@@ -124,23 +131,61 @@ export async function fetchFeeQuotes(
...
       recipient: recipient,
+      attestation,
       destinationToken: connectedDestinationToken,
     });
+    // Override interchainQuote with the value pinned at attestation-build time.
+    // estimateTransferRemoteFees re-quotes IGP internally; if the price drifted since
+    // the attestation was fetched, the displayed fee would diverge from the attested
+    // msg_value and the submit tx would revert on _authorizeTransaction.
+    if (pinnedInterchainFee) {
+      return { ...feeEstimate, interchainQuote: pinnedInterchainFee, tokenFeeQuote: pinnedTokenFeeQuote ?? feeEstimate.tokenFeeQuote };
```
paulbalaji: **🟡 Low — fee simulation still re-quotes IGP, even though display uses the pinned value.**

`estimateTransferRemoteFees` in SDK 33.0.0 (see `node_modules/@hyperlane-xyz/sdk/dist/warp/WarpCore.js:696` and upstream `typescript/sdk/src/warp/WarpCore.ts:1099-1115`) accepts `attestation` but **not** `interchainFee` / `tokenFeeQuote`. It calls `getInterchainTransferFee` internally and runs gas simulation against the fresh quote.

When IGP drifts meaningfully between attestation fetch and this simulation:
- The attestation was signed over the *pinned* `msgValue`.
- Internal `populateTransferRemoteTx` builds the sim tx with the *fresh* IGP → different outer `msg.value`.
- `AbstractPredicateWrapper._authorizeTransaction` Statement-hash check fails → estimateGas reverts.
- Falls to the catch on :189; fallback retry can't salvage it (predicate wrapper rejects non-attested transfers) → fees disappear for the user.

Submit path stays correct (pinned values flow through `getTransferRemoteTxs`). This is UX flake only, not a correctness bug.

Cleanest fix is monorepo-side: extend `estimateTransferRemoteFees` to accept `interchainFee` + `tokenFeeQuote` the same way `getTransferRemoteTxs` already does (`WarpCore.ts:451-468`), then thread the pinned values through here too. Worth filing as a follow-up on the SDK. Not a blocker for this PR.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -951,14 +951,26 @@ async function validateForm(
...
+        recipient,
+        sender: sender || '',
+        senderPubKey: await senderPubKey,
+        destinationToken: connectedDestinationToken,
+      });
+    } catch (error) {
+      const isPredicateRoute = await warpCore.isPredicateSupported(transferToken, destination);
+      if (!isPredicateRoute) throw error;
+      result = null;
+    }
```
paulbalaji, line 978: **🟡 Low — catch is broader than the comment suggests.**

The comment (:955-958) targets "gas simulation errors" from running `validateTransfer` without an attestation on a Predicate route. But the `catch` is un-narrowed — *any* error thrown from `validateTransfer` on a Predicate route sets `result = null` (= valid). That includes provider/RPC errors, wallet-side failures, or any future non-attestation throw the SDK might add.

Concrete failure mode: if the RPC goes sideways mid-validation on a Predicate route, the user sees "validation passed" and learns the real problem only at submit-time as a revert.

Tighter alternatives:
- Match on the specific Predicate wrapper revert signature (`0x8af999c4` per the monorepo discussion on this topic) before swallowing — rethrow anything else.
- Or check `isPredicateRoute` up front and skip `validateTransfer`'s simulation step via a dedicated SDK entrypoint (`validateTransferWithoutSimulation`?), if one can be added monorepo-side.

Acceptable as-is for a first cut; worth tightening in a follow-up.

paulbalaji (approved):
## LGTM — approving with two Low follow-ups

Cross-checked against codex; we converge on the same picture. The IGP-pinning redesign is clean and every substantive item from prior rounds (mine, CodeRabbit's, Xaroz's) is resolved on head `14cb297d`. Submit-time correctness is solid; the remaining items are low-severity UX flake and a defensive-scoping nit.

### Resolved on this head (not re-raising)

- **IGP-drift pinning** — `predicate.ts:72-78` quotes `igpQuote` + `tokenFeeQuote` once via `getInterchainTransferFee`, threads both into `getTransferRemoteTxs` at `predicate.ts:85-94`, and the new `PredicateAttestationResult` propagates them to the submit path (`useTokenTransfer.ts:200-204`). Clean design.
- Stage reset after attestation success (`useTokenTransfer.ts:195`).
- `(tx as any).category` / `(tempTx as any).transaction` → `tx.category` + `tempTx as WarpTypedTransaction`; `eslint-disable` removed.
- Server-side `PredicateApiClient` handles Zod/`is_compliant`/expiry validation (`src/pages/api/predicate/attestation.ts:82`).
- In-memory rate limiter removed in favor of Vercel firewall rules (with a comment to prevent reintroduction).
- 3 new tests in `useFeeQuotes.test.ts` covering predicate paths — fallback-sender skip, attestation fetch failure, and pinned-interchainFee override.
- `link-monorepo.js` refactor: shared `packPackage()` helper deduplicates the two pack paths.

### Two Low follow-ups (inline)

1. **Fee-display is still drift-sensitive** — `estimateTransferRemoteFees` in installed SDK 33.0.0 (verified in `node_modules/@hyperlane-xyz/sdk/dist/warp/WarpCore.js:696` and upstream `typescript/sdk/src/warp/WarpCore.ts:1099-1115`) does not accept pinned `interchainFee` / `tokenFeeQuote` — only `attestation`. It re-quotes IGP internally and gas-sims against that fresh quote. Under drift, the sim reverts (attested msgValue ≠ simulated msg.value), fees disappear. Submit path remains correct. Monorepo follow-up to extend the SDK signature would close this.

2. **Predicate-route `validateForm` swallows all thrown errors** — the try/catch on `TransferTokenForm.tsx:960-973` catches every throw from `validateTransfer` on Predicate routes, not just the attestation-less simulation failure it's targeting. Tighter scoping (by error code / match string) would prevent silently masking real provider/wallet errors.

### Deps / perf

Nothing changed on this axis since `316c5855`. Direct dep delta remains the intentional `@hyperlane-xyz/sdk|utils|widgets 32.0.0 → 33.0.0`. Emitted client bundle still flat vs `main`. `pnpm-lock.yaml` still carries ~8k lines of incidental caret-re-resolution drift unrelated to the Hyperlane bump — will conflict with #1030; worth a rebase-from-main-lock pass if that's easy.

### CI

`build`, `lint`, `test`, `typecheck`, `format`, `security-review`, Vercel, Socket — all green. `e2e` pending as of review time.

pnpm-lock.yaml
paulbalaji, line 1: i think you probably need to reset both this + package.json to `main` state to make the diff accurate

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -951,14 +951,31 @@ async function validateForm(
...
+        originTokenAmount,
+        destination,
+        recipient,
+        sender: sender || '',
+        senderPubKey: await senderPubKey,
+        destinationToken: connectedDestinationToken,
+      });
+    } catch (error) {
+      const isPredicateRoute = await warpCore.isPredicateSupported(transferToken, destination);
+      if (!isPredicateRoute) throw error;
```
Xaroz, line 971: does this throw make it so that second catch happens? just making sure because we don't wanna swallow all the errors that comes from validateTransfer

troykessler: yes it rethrows so we dont swallow it

troykessler:
@Xaroz let me know if you have any other concerns, would also wait for your approval on this

Xaroz (approved):

PR #1029, opened 2026-03-24
feat: adopt offchain fee quoting service API

yorhodes:
## Summary

- Integrate offchain fee quoting service from hyperlane-xyz/hyperlane-monorepo#8404
- Add Next.js API route proxy (`/api/quote`) to keep `FEE_QUOTING_API_KEY` server-side
- Add `useQuotedCalls` hook for offchain fee quoting via `QuotedCalls` contract
- Thread `QuotedCallsParams` from fee display through to `warpCore.getTransferRemoteTxs({ quotedCalls })`
- Falls back to existing onchain fee estimation when not configured or for non-EVM chains

## Test plan

- [ ] Set `NEXT_PUBLIC_FEE_QUOTING_URL` and `FEE_QUOTING_API_KEY`, verify offchain quotes fetched
- [ ] Verify QuotedCalls.execute() tx sent instead of direct transferRemote
- [ ] Without env vars: verify existing onchain flow unchanged
- [ ] Non-EVM chains fall back to onchain quoting


src/features/tokens/approval.ts
```diff
@@ -14,9 +19,20 @@ export function useIsApproveRequired(token?: IToken, amount?: string, enabled =
...
+      token?.addressOrDenom,
+      quotedCallsParams?.address,
+    ],
     queryFn: async () => {
       if (!token || !owner || !amount) return false;
+      // QuotedCalls: approval target is the quotedCalls contract, not the router
+      if (quotedCallsParams?.address) {
+        const adapter = token.getAdapter(warpCore.multiProvider);
+        return adapter.isApproveRequired(owner, quotedCallsParams.address, amount);
+      }
```
Xaroz, line 38: should this be inside the warpCore instead?

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,189 @@
...
+  sender: string | undefined,
+  amount: string | undefined,
+  recipient: string | undefined,
+): Promise<QuotedCallsFeeResult | null> {
+  if (!originToken || !destinationToken || !destination || !sender || !amount || !recipient)
+    return null;
+
+  // Look up quotedCalls address from registry
+  const addresses = await registry.getAddresses();
+  const chainAddresses = addresses[originToken.chainName];
```
Xaroz: probably better to call `await registry.getChainAddresses(originToken.chainName)`

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,189 @@
...
+
+function generateClientSalt(): Hex {
+  const bytes = new Uint8Array(32);
+  crypto.getRandomValues(bytes);
+  return toHex(bytes);
+}
+
+async function fetchQuotedCallsFees(
+  warpCore: WarpCore,
+  registry: import('@hyperlane-xyz/registry').IRegistry,
```
Xaroz: why not import `IRegistry` at the top of the file?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -697,20 +709,38 @@ function ReviewDetails({
...
+  // Approval check: uses quotedCalls address as spender when offchain quoting is active
+  const { isLoading: isApproveLoading, isApproveRequired } = useIsApproveRequired(
+    originToken,
+    amountWei,
+    isReview,
+    quotedCallsParams,
+  );
+
+  // Pass quotedCallsParams to parent via ref for use in transfer execution
+  quotedCallsParamsRef.current = quotedCallsParams ?? null;
```
Xaroz: you are bypassing React's data flow by using the useRef this way, so no re-renders will happen and unexpected behavior could occur, I think my recommendation is that at the parent you declare a state for quoted calls params

```tsx
const [quotedCallsParams, setQuotedCallsParams] = useState<QuotedCallsParams | null>(null);
```

then for your other components

```tsx
<ReviewDetaills
  ...
  onQuotedCallsParams={setQuotedCallsParams}
/>

<ButtonSection
  ...
  quotedCallsParams={quotedCallsParams}
/>
```
and inside review
```tsx
useEffect(() => {
  onQuotedCallsParams(quotedCallsParams ?? null);
}, [quotedCallsParams, onQuotedCallsParams]);
```

Xaroz:
@claude review

src/features/tokens/approval.ts
Xaroz: Fixed in 6a3bfc4e — added a TODO noting the parallel adapter call should be replaced once `WarpCore.isApproveRequired` accepts a spender override (SDK 33.1.0 doesn't support it yet).

src/features/transfer/useQuotedCalls.ts
Xaroz: Done in d948e2dc — went one further and read `quotedCalls` from the store-resolved `chainAddresses` (populated once at warp context init via the chainAddresses refactor in #1086). No registry call inside the queryFn anymore.

src/features/transfer/useQuotedCalls.ts
Xaroz: Resolved naturally in d948e2dc — `fetchQuotedCallsFees` no longer takes `IRegistry` (now takes the resolved `quotedCallsAddress` directly), so the inline import-type is gone.

src/features/transfer/TransferTokenForm.tsx
Xaroz: Fixed in 6a3bfc4e — added a `TransferCheckout` wrapper component that owns `useQuotedCallsFeeQuotes` and passes the result down to both `ReviewDetails` and `ButtonSection` as props. Drops the useRef entirely; fully React-tracked, no write-during-render.

package.json
```diff
@@ -20,10 +20,11 @@
...
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^2.2.0",
+    "@hyperlane-xyz/fee-quoting": "link:../monorepo/typescript/fee-quoting",
     "@hyperlane-xyz/registry": "24.1.0",
-    "@hyperlane-xyz/sdk": "29.0.0",
-    "@hyperlane-xyz/utils": "29.0.0",
-    "@hyperlane-xyz/widgets": "29.0.0",
+    "@hyperlane-xyz/sdk": "link:../monorepo/typescript/sdk",
+    "@hyperlane-xyz/utils": "link:../monorepo/typescript/utils",
+    "@hyperlane-xyz/widgets": "link:../monorepo/typescript/widgets",
```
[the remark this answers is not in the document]
Xaroz: Resolved — these moved to the published versions when main was merged in (`@hyperlane-xyz/sdk@33.1.0`, `utils@33.1.0`, `widgets@33.1.0`, `registry@24.3.0`). No more `link:` refs or local overrides.

src/consts/config.ts
```diff
@@ -13,6 +13,8 @@ const chainWalletWhitelists = JSON.parse(process?.env?.NEXT_PUBLIC_CHAIN_WALLET_
 const rpcOverrides = process?.env?.NEXT_PUBLIC_RPC_OVERRIDES || '';
 const explorerApiUrl =
   process?.env?.NEXT_PUBLIC_EXPLORER_API_URL || 'https://explorer4.hasura.app/v1/graphql';
+const feeQuotingUrl =
+  process?.env?.NEXT_PUBLIC_FEE_QUOTING_URL || 'https://quoting.services.hyperlane.xyz';
```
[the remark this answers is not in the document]
Xaroz: Fixed in 6a3bfc4e — `feeQuotingUrl` now resolves to `undefined` when the env var isn't set, so the hook short-circuits and `/api/quote` is never called. Also did the same on the server side (`pages/api/quote.ts`) so the `!apiKey || !baseUrl` 503 actually triggers.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -96,6 +97,8 @@ export function TransferTokenForm() {
   // This state is used for when the formik token is different from
   // the token with highest collateral in a multi-collateral token setup
   const [routeOverrideToken, setRouteTokenOverride] = useState<Token | null>(null);
+  // Ref for passing QuotedCallsParams from ReviewDetails to ButtonSection
+  const quotedCallsParamsRef = useRef<QuotedCallsParams | null>(null);
```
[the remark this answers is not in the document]
Xaroz: Fixed in 6a3bfc4e — see reply on the Xaroz thread above. New `TransferCheckout` wrapper owns the hook and passes `quotedCallsParams` down as props; ref is gone.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,189 @@
...
+      : null,
+    quotedCallsParams: data?.quotedCallsParams ?? null,
+  };
+}
+
+function generateClientSalt(): Hex {
+  const bytes = new Uint8Array(32);
+  crypto.getRandomValues(bytes);
+  return toHex(bytes);
+}
```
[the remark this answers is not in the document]
Xaroz: Skipping. Verified the flagged scenario doesn't actually fail: signed quotes have an `expiry = issuedAt + 300s` (5min) window enforced on-chain (`AbstractOffchainQuoter.submitQuote`), and the 30s `refetchInterval` keeps a fresh quote with ≥4.5min validity. The user always submits the latest *settled* quote (react-query `data` only updates on settle), so there's no salt/quote mismatch race. Salts themselves don't expire — they're just nonces. `generateClientSalt()` is sub-millisecond CPU, so per-refetch regeneration is essentially free; the stability optimization didn't justify the added complexity.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,189 @@
...
+        sender,
+        debouncedAmount,
+        recipient,
+      ),
+    enabled: shouldFetch,
+    refetchInterval: FEE_QUOTE_REFRESH_INTERVAL,
+  });
+
+  return {
+    isLoading: isLoading || isFetching,
```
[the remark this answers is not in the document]
Xaroz: Fixed in e0619ff4 — dropped `isFetching` from the return so background refetches don't flash the spinner.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,189 @@
...
+  const { isLoading, isError, data, isFetching } = useQuery({
+    // eslint-disable-next-line @tanstack/query/exhaustive-deps
+    queryKey: [
+      'useQuotedCallsFeeQuotes',
+      originTokenKey,
+      destinationTokenKey,
+      sender,
+      debouncedAmount,
+      recipient,
+    ],
```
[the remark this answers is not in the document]
Xaroz: Documented in e0619ff4. The closed-over `(warpCore, originToken, destinationToken, destination)` are class instances and can't be safely stringified into a query key. Their identity is already covered: `originToken`/`destinationToken`/`destination` via the `originTokenKey`/`destinationTokenKey` proxies in the key, and `warpCore` is a session-stable store singleton. Added a multi-line comment on the eslint-disable explaining this so a future reader doesn't second-guess it.

src/features/transfer/useQuotedCalls.ts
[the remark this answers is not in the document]
Xaroz: Resolved in d948e2dc, going further than `getChainAddresses` — `quotedCalls` is now read synchronously from the store's pre-loaded `chainAddresses` (populated once at warp context init via the chainAddresses store refactor in #1086). No registry call inside the queryFn at all.

src/features/transfer/useQuotedCalls.ts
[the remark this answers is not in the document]
Xaroz: Resolved in d948e2dc — `IRegistry` is no longer used inside this hook (registry call replaced by store read), so the inline import-type was removed.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,189 @@
...
+  const { igpQuote, tokenFeeQuote, feeQuotes } = await warpCore.getQuotedTransferFee({
+    originTokenAmount,
+    destination,
+    sender,
+    recipient,
+    quotedCalls: quotedCallsParams,
+  });
+
+  // Attach feeQuotes to params for later use in getTransferRemoteTxs
+  quotedCallsParams.feeQuotes = feeQuotes;
```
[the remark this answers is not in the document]
Xaroz: Fixed in e0619ff4 — switched to a spread (`{ ...baseQuotedCallsParams, feeQuotes }`) instead of mutating the original after construction.

src/pages/api/quote.ts
```diff
@@ -0,0 +1,34 @@
...
+const apiKey = process.env.FEE_QUOTING_API_KEY;
+const baseUrl = process.env.NEXT_PUBLIC_FEE_QUOTING_URL || 'https://quoting.services.hyperlane.xyz';
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  if (req.method !== 'GET') return res.status(405).json({ message: 'Method not allowed' });
+  if (!apiKey || !baseUrl) return res.status(503).json({ message: 'Fee quoting not configured' });
+
+  const { command, origin, router, destination, salt, recipient } = req.query;
+
+  if (!command || !origin || !router || !destination || !salt) {
```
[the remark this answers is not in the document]
Xaroz: Recipient validation fixed in e0619ff4 — added a check using the SDK's `WARP_FEE_COMMANDS` set so warp commands without a recipient return 400 instead of failing downstream. Skipped the module-level `FeeQuotingClient` instance: in serverless (Vercel), warm containers reuse modules but cold starts get fresh ones, and the constructor itself is just two assignments + a regex strip — sub-millisecond. Per-request instantiation is negligible.

src/features/tokens/approval.ts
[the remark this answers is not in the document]
Xaroz: Fixed in 6a3bfc4e — added a TODO noting the parallel adapter call should be replaced once `WarpCore.isApproveRequired` accepts a spender override. SDK 33.1.0 doesn't expose one yet; tracked as upstream follow-up.

Xaroz:
@claude review again

src/pages/api/quote.ts
```diff
@@ -0,0 +1,48 @@
...
+  // Warp commands (transferRemote / transferRemoteTo) require a recipient.
+  if (WARP_FEE_COMMANDS.has(command as FeeQuotingCommand) && !recipient) {
+    return res.status(400).json({ message: 'recipient required for warp commands' });
+  }
+
+  const client = new FeeQuotingClient({ baseUrl, apiKey });
+
+  try {
+    const response = await client.getQuote({
+      command: command as FeeQuotingCommand,
```
paulbalaji, line 65: **[Blocker]** `command` reaches `client.getQuote` via a type-cast only — no runtime allowlist. `FeeQuotingClient.getQuote` interpolates it raw into the upstream URL path: `${baseUrl}/quote/${params.command}?...` (sdk `quoted-calls/client.js:19`). The `WARP_FEE_COMMANDS.has(...)` check above only gates the `recipient` requirement; non-matching commands fall through.

A request like `command=foo/../admin` would resolve to `${baseUrl}/admin` on the upstream, bypassing path scoping while spending the server-side API key. The base URL is fixed so this isn't full SSRF, but it's still a meaningful path-injection surface on a public endpoint.

```ts
import { FeeQuotingCommand } from '@hyperlane-xyz/sdk';
const ALLOWED_COMMANDS = new Set(Object.values(FeeQuotingCommand));
if (!ALLOWED_COMMANDS.has(command as FeeQuotingCommand)) {
  return res.status(400).json({ message: 'Invalid command' });
}
```

Do this before constructing the client.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -453,18 +453,77 @@ function TokenBalance({
...
+  const tokenMap = useTokenByKeyMap();
+  const isRouteSupported = useIsRouteSupported();
+
+  const originTokenByKey =
+    routeOverrideToken || getTokenByKeyFromMap(tokenMap, values.originTokenKey);
+  const destinationTokenByKey = getTokenByKeyFromMap(tokenMap, values.destinationTokenKey);
+  const originToken =
+    destinationTokenByKey && originTokenByKey
+      ? findRouteToken(warpCore, originTokenByKey, destinationTokenByKey)
+      : undefined;
```
paulbalaji: **[Blocker]** `TransferCheckout` resolves `originToken` via `findRouteToken(...)`, which can return a different multi-collateral variant than the input. But `ButtonSection` (line 531) and `executeTransfer` (`useTokenTransfer.ts:130`) both use `routeOverrideToken || getTokenByKey(...)` directly — no second `findRouteToken` pass.

The offchain quote sends `router: originToken.addressOrDenom` (`useQuotedCalls.ts:168`). If `findRouteToken` selects a different token here than what `executeTransfer` ultimately calls, the signed quote is bound to the wrong router and `QuotedCalls.execute()` will fail on-chain.

In the steady state `validateForm` already produced the resolved token as `routeOverrideToken`, so they tend to agree. But that's an implicit invariant that depends on `findRouteToken` being idempotent on validated input. Two safer fixes:

- Drop the `findRouteToken` call here so all three sites share one resolution path, or
- Hoist a single `(originToken, destinationToken)` resolution into `TransferTokenForm` and pass it down.

Related: `useQuotedCalls.ts:87` keys the query on `originTokenKey` (the form-value proxy), not the resolved `originToken.addressOrDenom`. When `routeOverrideToken` flips during validation, the cache entry stays under the same key and serves a quote built against a stale router. Including the resolved router/destination addresses in the queryKey would close that gap regardless of which fix above you take.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,229 @@
...
+  const clientSalt = generateClientSalt();
+  const salt = computeScopedSalt(sender as Address, clientSalt);
+
+  // Get destination domain ID
+  const destinationDomainId = warpCore.multiProvider.getDomainId(destination);
+
+  // Fetch quotes from API proxy
+  const recipientBytes32 = addressToBytes32(recipient) as Hex;
+  const params = new URLSearchParams({
+    command: 'transferRemote',
```
paulbalaji, line 221: **[Warning]** `command` is hardcoded to `'transferRemote'` and `targetRouter` is never sent. Cross-collateral QuotedCalls flows use `TRANSFER_REMOTE_TO` with an explicit `targetRouter` — for those routes this hook will request a quote that doesn't match the call the SDK will actually make.

Either derive `command`/`targetRouter` from the route shape (when origin and destination tokens have different routers in the warp config), or short-circuit `shouldFetch` for cross-collateral routes so they fall through to the onchain path until offchain support lands.

paulbalaji (commented):
## Review Summary

Round 1 feedback was addressed cleanly — the useRef-during-render shuttle is gone (replaced by the `TransferCheckout` wrapper), `feeQuotingUrl` defaults to `undefined` so offchain quoting is opt-in, the registry call became a sync store read, link: deps are unpinned. The salt-regeneration skip rationale checks out: react-query v5 preserves `data` on refetch error, the on-chain quote expiry is 5min, and the 30s refetch interval keeps a fresh signed quote with ample validity headroom.

Two blockers to address before merge:

1. **`/api/quote` `command` is not validated** — it's interpolated raw into the upstream URL path, allowing path-segment injection under the configured host while burning the server-side API key. Inline below.
2. **Quoted router can drift from executed router** — `TransferCheckout` runs `findRouteToken` an extra time that `ButtonSection` and `executeTransfer` don't, so the offchain quote can be bound to a different router than the actual transfer. The queryKey identity gap compounds this. Inline below.

Three more inline items: hardcoded `transferRemote` blocks cross-collateral routes, the local-gas estimator failure is silently swallowed (collides with the CLAUDE.md "no silent fallbacks" rule), and a minor UX regression where the collateral address is hidden under offchain quoting.

## Operational concern (not inline)

Unlike the sibling `pages/api/predicate/attestation.ts`, `pages/api/quote.ts` has no comment about edge-layer rate limiting. Anyone can hit `/api/quote` and consume the project's upstream quota. Suggest adding the same Vercel-firewall-rules note that `attestation.ts` carries, so deployers know the file expects external rate-limiting to be configured.

## Acknowledging existing comments

Not duplicating CodeRabbit's three unresolved comments — silent offchain fallback (`TransferTokenForm.tsx:760`), `Number(destination)` NaN (`quote.ts:21`), and upstream error message verbatim forwarding (`quote.ts:43`). All valid; defer to those threads.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,229 @@
...
+      sender,
+      interchainFee: igpQuote,
+      tokenFeeQuote,
+      amount: originTokenAmount.amount,
+      destinationToken,
+      quotedCalls: quotedCallsParams,
+    });
+  } catch (e) {
+    logger.warn('Failed to estimate local gas for QuotedCalls', e);
+  }
```
paulbalaji: **[Warning]** Silent fallback — on failure `localQuote` stays at `igpQuote.token.amount(0n)`. The review UI hides the local-gas row when `amount > 0n` (`TransferTokenForm.tsx:854`), so users don't see "0 ETH", but they also see no signal that the estimator failed.

CLAUDE.md is explicit: "NEVER add silent fallbacks for unexpected issues — they mask bugs." Two reasonable directions:

- Re-throw so the query fails and the consumer falls back to onchain quoting (which has its own local-gas path).
- Or surface the failure as `localQuote: undefined` and have the UI render "Local gas estimate unavailable" rather than silently omitting the row.

Current behavior makes a regressed estimator look indistinguishable from a route that legitimately has no local gas.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -755,9 +823,9 @@ function ReviewDetails({
               {isApproveRequired && (
                 <div>
                   <h4>Transaction 1: Approve Transfer</h4>
-                  <div className="ml-1.5 mt-1.5 space-y-1.5 border-l border-gray-300 pl-2 text-xs dark:border-primary-300/25">
-                    <p>{`Router Address: ${originToken?.addressOrDenom}`}</p>
-                    {originToken?.collateralAddressOrDenom && (
+                  <div className="ml-1.5 mt-1.5 space-y-1.5 border-l border-gray-300 pl-2 text-xs">
+                    <p>{`Spender: ${quotedCallsParams?.address ?? originToken?.addressOrDenom}`}</p>
+                    {!quotedCallsParams && originToken?.collateralAddressOrDenom && (
```
paulbalaji: **[Suggestion]** When `quotedCallsParams` is set, the "Collateral Address" line is suppressed. The new "Spender" row tells the user *who* can spend; the collateral address tells them *what* ERC20 they're approving. Both are useful pre-approval — users double-checking against a token contract still want to see the underlying ERC20 even when the spender is a quoted-calls dispatcher.

Suggest showing both unconditionally — the spender changes between the two paths, but the collateral address doesn't.

src/pages/api/quote.ts
Xaroz: Fixed in b2aad8c0 — added a strict `ALLOWED_COMMANDS` allowlist (built from `FeeQuotingCommand` enum) that rejects unknown commands with 400 before constructing the client. `foo/../admin` no longer reaches the upstream URL path.

src/pages/api/quote.ts
```diff
@@ -0,0 +1,48 @@
...
+  const origin = firstString(req.query.origin);
+  const router = firstString(req.query.router);
+  const destination = firstString(req.query.destination);
+  const salt = firstString(req.query.salt);
+  const recipient = firstString(req.query.recipient);
+  const targetRouter = firstString(req.query.targetRouter);
+
+  if (!command || !origin || !router || !destination || !salt) {
+    return res.status(400).json({ message: 'Missing required query parameters' });
+  }
```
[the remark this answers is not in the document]
Xaroz: Fixed in b2aad8c0 — `destination` is now coerced to a number once and validated with `Number.isInteger(...) && > 0` before being passed into `client.getQuote`. Non-numeric input returns 400 with a clear message.

src/pages/api/quote.ts
```diff
@@ -0,0 +1,48 @@
...
+      destination: Number(destination),
+      salt: salt as `0x${string}`,
+      recipient: recipient as `0x${string}` | undefined,
+      targetRouter: targetRouter as `0x${string}` | undefined,
+    });
+    return res.status(200).json(response);
+  } catch (error: any) {
+    const message = error?.message || 'Fee quoting request failed';
+    return res.status(502).json({ message });
+  }
```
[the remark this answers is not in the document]
Xaroz: Fixed in b2aad8c0 — full error is logged server-side via `console.error`; the 502 response now returns a generic `'Fee quoting request failed'` message so upstream URLs / auth hints don't leak to the browser. Added a test covering this.

src/features/transfer/TransferTokenForm.tsx
Xaroz: Fixed in b2aad8c0 — dropped `findRouteToken` from `TransferCheckout` so origin-token resolution matches `ButtonSection` / `executeTransfer` exactly (`routeOverrideToken || getTokenByKey(...)`). `validateForm` (which formik runs on each change) already calls `getTransferToken` to produce the multi-collateral / cross-asset optimal token and stores it in `routeOverrideToken`, so re-resolving here was redundant and risked picking differently than executeTransfer. The destination still uses `findConnectedDestinationToken` for display. queryKey unchanged — form-value keys are sufficient because the resolved token agrees with the caller-passed origin.

src/features/transfer/useQuotedCalls.ts
Xaroz: Partial fix in b2aad8c0 — added the cross-collateral short-circuit you suggested as the second option. The hook now skips offchain quoting for cross-collateral routes (`warpCore.isCrossCollateralTransfer(...)`) and falls through to onchain quoting. Left a TODO to wire `command='transferRemoteTo'` + `targetRouter` through to `/api/quote` properly in a follow-up. Avoids shipping incorrect quotes for cross-collateral users today.

src/features/transfer/useQuotedCalls.ts
Xaroz: Fixed in b2aad8c0 — dropped the silent catch around `getLocalTransferFeeAmount`. A throw now drops the whole offchain result (`fees` + `quotedCallsParams` together), so the consumer falls through to the plain `transferRemote` path. The onchain `useFeeQuotes` runs and shows a local-gas estimate for the actual tx the user will sign — no mixed offchain/onchain mode.

src/features/transfer/TransferTokenForm.tsx
Xaroz: Fixed in b2aad8c0 — collateral address now shows unconditionally in the approve summary, regardless of whether `quotedCallsParams` is set. The spender changes between paths but the collateral address doesn't.

src/pages/api/quote.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  test('returns 502 on upstream error', async () => {
+    mockGetQuote.mockRejectedValue(new Error('upstream failure'));
+
+    const { req, res } = mockReqRes('GET', validQuery);
+    await handler(req, res);
+
+    expect(res.status).toHaveBeenCalledWith(502);
+    expect(res.json).toHaveBeenCalledWith({ message: 'upstream failure' });
+  });
+});
```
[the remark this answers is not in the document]
Xaroz: Fixed in b2aad8c0 — added the warp-recipient test plus two more covering the round-2/round-3 validation paths: command allowlist (path-injection guard) and `destination` integer guard. Total 7 tests on `quote.ts` now.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,229 @@
...
+        debouncedAmount,
+        recipient,
+      ),
+    enabled: shouldFetch,
+    refetchInterval: FEE_QUOTE_REFRESH_INTERVAL,
+  });
+
+  return {
+    isLoading,
+    isError,
```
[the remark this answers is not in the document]
Xaroz: Fixed in b2aad8c0 — dropped `isError` from the return type and the `useQuery` destructure. With the silent-catch removal in `getLocalTransferFeeAmount`, failures now propagate through React Query naturally; server-side errors are logged in `/api/quote`'s catch block. No surface for `isError` to be mis-consumed.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -684,20 +737,35 @@ function ReviewDetails({
...
-    isRouteSupported,
+    isRouteSupported && offchainUnavailable,
     originToken,
     destinationToken,
     !isReview,
   );
 
+  const feeQuotes = offchainFeeQuotes ?? onchainFeeQuotes;
+  const isQuoteLoading = offchainUnavailable ? isOnchainQuoteLoading : isOffchainQuoteLoading;
+
```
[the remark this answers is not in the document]
Xaroz: Largely covered by b2aad8c0 — three of the four silent-failure surfaces are now addressed: server logs full upstream errors (`/api/quote` catch), `getLocalTransferFeeAmount` no longer silently catches, and cross-collateral routes short-circuit instead of getting incorrect quotes. The remaining gap is *user-facing* surfacing of offchain failures (toast/banner). Skipping that for now — adding a toast on every transient RPC blip during background refetches is more UX noise than signal, and operators get the failure via `console.error` server-side. Open to revisiting if production telemetry shows users hitting silent fallbacks regularly.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -199,6 +207,7 @@ async function executeTransfer({
       destination,
       sender,
       recipient,
+      quotedCalls: quotedCallsParams,
```
paulbalaji: 🔴 **HIGH — `quotedCalls` and `attestation` are mutually exclusive in the SDK.**

`WarpCore.getTransferRemoteTxs` asserts this explicitly (`@hyperlane-xyz/sdk` `dist/warp/WarpCore.js:261`):
```js
assert(!(quotedCalls && attestation), 'quotedCalls and attestation cannot be used together. The QuotedCalls path does not support attestation-gated transfers.');
```

For any chain that has both a predicate-wrapped route *and* a `quotedCalls` chain address with `feeQuotingUrl` configured, both branches activate: `useQuotedCallsFeeQuotes` populates `quotedCallsParams`, `fetchPredicateAttestation` populates `attestationResult`, both get passed here, the SDK throws, and the transfer never gets sent.

Minimal fix at this site:
```ts
quotedCalls: attestationResult ? undefined : quotedCallsParams,
```

Semantically: an attested predicate route requires the wrapper path; the QuotedCalls atomic path doesn't support attestation gating, so it has to lose. Worth adding a brief comment naming the exclusion. Alternatively, gate `shouldFetch` in `useQuotedCalls.ts` on `!isPredicateSupported(...)` so the offchain quote never gets requested for predicate routes — but that's an async check, so threading it through is more work; the one-liner here covers the same ground.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,237 @@
...
+    // warpCore is a store singleton that's stable for the session.
+    queryKey: [
+      'useQuotedCallsFeeQuotes',
+      originTokenKey,
+      destinationTokenKey,
+      sender,
+      debouncedAmount,
+      recipient,
+      quotedCallsAddress,
+    ],
```
paulbalaji: 🔴 **HIGH — cache-identity gap survives `routeOverrideToken` flips and the `isCrossCollateral` gate.**

Round-3 dropped the extra `findRouteToken` in `TransferCheckout`, which closes the render-time divergence between `TransferCheckout` and `ButtonSection`/`executeTransfer`. Good. But the query cache key still uses `originTokenKey` (the form value), not the resolved router address — so a cached entry survives the very transitions that should invalidate it:

**Scenario A — `routeOverrideToken` flip (multi-collateral resolution):**
1. User picks token A. Quote fetched, cached under key `[..., originTokenKey: "chainA-A", ...]`. `data.quotedCallsParams.quotes` are signed against router A.
2. `validateForm` runs, sets `routeOverrideToken = B`. Re-render: `originToken` is now B; `originTokenKey` is unchanged.
3. queryKey unchanged → react-query serves cached `data`. UI surfaces quotes for router A while `originToken` is B.
4. `executeTransfer` calls `getTransferRemoteTxs` with router-B amount and router-A quotes → on-chain `QuotedCalls.execute()` reverts.

**Scenario B — `shouldFetch` goes false:**
When `isCrossCollateral` flips true (or `enabled` flips false, or `quotedCallsAddress` becomes undefined), `shouldFetch = false` disables the query, but react-query preserves `data` and the hook returns `{quotedCallsParams: data.quotedCallsParams, ...}`. The consumer happily passes a stale quote into a route it wasn't built for.

Two orthogonal fixes — I'd suggest doing both:

1. Include the resolved identities in the queryKey:
   ```ts
   queryKey: [
     'useQuotedCallsFeeQuotes',
     originToken?.addressOrDenom,
     destinationToken?.addressOrDenom,
     sender,
     debouncedAmount,
     recipient,
     quotedCallsAddress,
     isCrossCollateral,
   ],
   ```
   You can drop `originTokenKey`/`destinationTokenKey` — `addressOrDenom` already differentiates them.

2. Null out the return when `!shouldFetch`, so disabled queries can't leak stale data downstream:
   ```ts
   return {
     isLoading: shouldFetch ? isLoading : false,
     fees: shouldFetch && data ? { ... } : null,
     quotedCallsParams: shouldFetch ? (data?.quotedCallsParams ?? null) : null,
   };
   ```

(2) alone closes Scenario B; (1) is needed for Scenario A and is also more in line with the eslint-disable's stated reasoning — "identity is covered by the *Key proxies" — which is no longer accurate now that resolved tokens can diverge from form values via `routeOverrideToken`.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,237 @@
...
+        destination,
+        sender,
+        debouncedAmount,
+        recipient,
+      ),
+    enabled: shouldFetch,
+    refetchInterval: FEE_QUOTE_REFRESH_INTERVAL,
+  });
+
+  return {
```
paulbalaji, line 161: 🟠 **MEDIUM — refetch error preserves stale signed quotes past their on-chain expiry.**

The round-1 salt-refetch skip rationale was: signed quotes have a 5min on-chain expiry, the 30s `refetchInterval` keeps a fresh quote with ≥4.5min validity, react-query only updates `data` on settle, so the user always submits the latest settled quote. That holds **as long as refetches succeed**.

Under an upstream outage longer than the 5min expiry, react-query (v5 default) preserves `data` from the last successful fetch through subsequent failed refetches. The UI keeps surfacing the same `quotedCallsParams`; if the user clicks Send, the on-chain expiry check rejects the quote and the tx reverts. With `isError` dropped from the return, the consumer has no signal to display.

A few mitigations, in increasing invasiveness:
- Stamp `issuedAt: Date.now()` into `QuotedCallsFetchResult`, and have the consumer treat the quote as missing when `Date.now() - issuedAt > 4.5 * 60_000` (gives 30s headroom under the 5min on-chain expiry). Falls back to onchain quoting cleanly.
- Or set `gcTime: 4.5 * 60_000` so react-query drops the entry after that window even on consecutive failures.
- Or surface `isError`/`isStale` and let the UI decide.

First option is the cleanest IMO — server time is irrelevant to the on-chain check (which is `block.timestamp` based), but a generous local cap is a reasonable approximation.

src/features/tokens/approval.ts
```diff
@@ -15,9 +20,22 @@ export function useIsApproveRequired(token?: IToken, amount?: string, enabled =
...
+      quotedCallsParams?.address,
+    ],
     queryFn: async () => {
       if (!token || !owner || !amount) return false;
+      // QuotedCalls: approval target is the quotedCalls contract, not the router
+      // TODO: upstream a `spender` override to WarpCore.isApproveRequired so this
+      // parallel adapter call can be removed.
+      if (quotedCallsParams?.address) {
+        const adapter = token.getAdapter(warpCore.multiProvider);
+        return adapter.isApproveRequired(owner, quotedCallsParams.address, amount);
```
paulbalaji, line 37: 🟠 **MEDIUM — preview ignores token fees that the SDK includes in `totalTokenNeeded`.**

The SDK's actual approval check (`@hyperlane-xyz/sdk` `dist/warp/WarpCore.js:639-650` in `getQuotedCallsTransferTxs`) computes:
```js
const { tokenTotals } = extractQuoteTotals(feeQuotes);
const totalTokenNeeded = tokenTotals.get(transferParams.token.toLowerCase()) ?? 0n;
// ...
adapter.isApproveRequired(sender, quotedCalls.address, totalTokenNeeded)
```
so it approves `transfer amount + token fees`. This UI preview only checks `amount`.

Edge case: a user whose existing allowance is exactly `amount` (e.g. they re-bridged the same value previously) sees "no approval required" in the review, but `getTransferRemoteTxs` computes `totalTokenNeeded > allowance`, prepends an approval tx, and the user gets an unexpected wallet popup.

Fix needs the fee-quotes total to be available here. Two options:
- Thread the resolved `feeQuotes` (or just `totalTokenNeeded` as a wei amount) from `useQuotedCallsFeeQuotes` through to `useIsApproveRequired`.
- Or push for the upstream `WarpCore.isApproveRequired` spender override that the existing TODO already calls out — and add an amount override at the same time so the wrapper goes away entirely.

Low likelihood, but it's exactly the class of bug the round-2 "parallel approval path" comment was warning about.

src/pages/api/quote.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { NextApiRequest, NextApiResponse } from 'next';
+
+const apiKey = process.env.FEE_QUOTING_API_KEY;
+const baseUrl = process.env.NEXT_PUBLIC_FEE_QUOTING_URL || undefined;
+// Strict allowlist — `command` is interpolated raw into the upstream URL path
+// by FeeQuotingClient (`${baseUrl}/quote/${command}?...`), so a bare type-cast
+// would be a path-injection surface.
+const ALLOWED_COMMANDS = new Set<FeeQuotingCommand>(Object.values(FeeQuotingCommand));
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
```
paulbalaji, line 22: 🟠 **MEDIUM — still no rate-limit / quota invariant comment** matching the sibling `pages/api/predicate/attestation.ts`:
```ts
// Rate limiting is handled by Vercel firewall rules at the edge layer.
// Do not add in-memory rate limiting here — serverless instances are ephemeral
// and load-balanced, so module-level state is not shared across invocations.
```
This endpoint forwards an unauthenticated request to a paid upstream service using a server-side API key. Without an explicit comment, deployers who fork this template are unlikely to realize they need to configure edge-layer rate limiting before exposing the endpoint, and the project's quota is one curl loop away from being drained.

Not a code change beyond a comment, but worth carrying the same invariant the predicate proxy carries.

paulbalaji (commented):
## Re-review (commit `b2aad8c0`)

Round 3 cleanly resolves all five of my prior inline comments and the three CodeRabbit threads. Specifically:

- ✅ `command` allowlist on `/api/quote` (path-injection blocker)
- ✅ `destination` integer guard
- ✅ Generic 502 message + server-side full error log
- ✅ Single origin-token resolution path (TransferCheckout no longer calls `findRouteToken`)
- ✅ Collateral address shown unconditionally
- ✅ Cross-collateral routes short-circuit to onchain quoting
- ✅ `getLocalTransferFeeAmount` no longer silently catches
- ✅ `useFeeQuotes` drops `isFetching` from return — refetches no longer flash spinners (matches the offchain hook)

## Still open

Thanks to codex for surfacing two of these on the latest pass.

1. 🔴 **HIGH — predicate × quoted-calls is a runtime assert** (inline on `useTokenTransfer.ts:210`). SDK throws when both are passed.
2. 🔴 **HIGH — cache-identity gap on the offchain quote** (inline on `useQuotedCalls.ts:91-106`). The fix in round 3 made TransferCheckout / ButtonSection / executeTransfer share one resolution path, which closes the *render-time* divergence — but the query cache still keys on `originTokenKey` (the form value), not on the resolved router address or the `isCrossCollateral` gate. When `routeOverrideToken` flips post-validation, or when the route becomes cross-collateral and `shouldFetch` goes `false`, react-query keeps serving the cached `quotedCallsParams` from the previous router.
3. 🟠 **MEDIUM — refetch error preserves stale signed quotes past on-chain expiry** (inline on `useQuotedCalls.ts:122`). The salt-refetch skip rationale assumed refetches succeed; under a 5min+ upstream outage, react-query keeps the last successful `data` and the UI can submit an expired quote.
4. 🟠 **MEDIUM — approval preview ignores token fees** (inline on `approval.ts:37`). SDK's `getQuotedCallsTransferTxs` checks `amount + tokenFees` via `extractQuoteTotals`; the UI preview only checks `amount`.
5. 🟠 **MEDIUM — `api/quote.ts` still lacks the rate-limit invariant comment** that `predicate/attestation.ts` carries (inline on `api/quote.ts:11`).

## Bottom line

Almost there. Items 1 and 2 are the gating ones — both are real on-chain failure paths, both are small fixes. Items 3-5 are smaller and could ship in a follow-up if needed.

src/features/transfer/useTokenTransfer.ts
Xaroz: Fixed in d8374a75 — `quotedCalls` is now skipped when an attestation is present:
```ts
quotedCalls: attestationResult ? undefined : quotedCallsParams,
```

src/features/transfer/useQuotedCalls.ts
Xaroz: Fixed in d8374a75. Two changes in `useQuotedCallsFeeQuotes`:
1. Added `originToken?.addressOrDenom` and `destinationToken?.addressOrDenom` to the queryKey so a token swap that keeps the `*Key` proxies stable still busts the cache.
2. Suppress the cached `data` when `shouldFetch` is false (form invalid) so a stale quote doesn't leak into the consumer between fetches.

src/features/transfer/useQuotedCalls.ts
Xaroz: Fixed in d8374a75. Stamp `issuedAt` on the cached result and gate consumption on `Date.now() - issuedAt < MAX_QUOTE_AGE_MS` (4 min — 1 min margin under the 5 min upstream expiry).

When stale, the hook reports `isLoading: isFetching`:
- Stale + a fresh fetch is in flight (focus refetch, retry): keep the form in offchain mode.
- Stale + nothing in flight (paused interval, exhausted retries): `isFetching: false` flows through to the consumer's `offchainSettled` branch and the form falls back to onchain quoting instead of spinning forever.

src/features/tokens/approval.ts
Xaroz: Fixed in d8374a75 — approval check now inflates by `tokenFeeQuote.amount` when the fee is denominated in the transferred token:
```ts
const sameTokenFeeAmount =
  quotedCallsParams && feeQuotes?.tokenFeeQuote?.token.equals(originToken)
    ? feeQuotes.tokenFeeQuote.amount
    : 0n;
const approvalAmountWei = isNft
  ? amountWei
  : (BigInt(amountWei || '0') + sameTokenFeeAmount).toString();
```
Limited the inflation to the QuotedCalls path so the onchain path keeps its existing `warpCore.isApproveRequired` semantics.

src/pages/api/quote.ts
Xaroz: Fixed in d8374a75 — added an invariant comment documenting that rate limiting lives at the Vercel platform layer + upstream Hyperlane fee quoting service, and warning future maintainers not to add an app-layer limiter that would double-count against the same budget. (Vercel rate-limiting will be enabled separately on this route.)

src/features/transfer/useQuotedCalls.ts
```diff
@@ -0,0 +1,259 @@
...
+    !!destinationToken &&
+    warpCore.isCrossCollateralTransfer(originToken, destinationToken);
+  const isFormValid = !!(originToken && destination && debouncedAmount && recipient && sender);
+  const shouldFetch =
+    enabled &&
+    isFormValid &&
+    isEvm &&
+    !isCrossCollateral &&
+    !!config.feeQuotingUrl &&
+    !!quotedCallsAddress;
```
paulbalaji, line 99: 🟠 **MEDIUM — `shouldFetch` doesn't gate on predicate-route status, so predicate routes still preview offchain quotes that won't be used.**

For a predicate route with `feeQuotingUrl` configured and a `quotedCalls` chain address present:

1. `shouldFetch = true` → offchain quote is fetched and stored in `data`.
2. `ReviewDetails` consumes `quotedCalls.fees` (offchain) and `quotedCallsParams` flows to `ButtonSection`. The review shows the offchain fee breakdown and the QuotedCalls contract as the approval spender.
3. User clicks Send. `executeTransfer` calls `fetchPredicateAttestation`, gets `attestationResult`, and `useTokenTransfer.ts:212` drops `quotedCallsParams` (`attestationResult ? undefined : quotedCallsParams`).
4. Approval popup is for the predicate router (different address than the previewed QuotedCalls spender). Fee amounts may also differ since the predicate path goes through `transferRemoteWithAttestation` with pinned IGP from the attestation, not the offchain quote.

The round-3 fix at the SDK call site is correct (it's the only place that *can* know whether attestation is present, since attestation is fetched at submit time). But because the hook eagerly fetches, the review previews a path the executor will reject.

Most direct fix: mirror the `useFeeQuotes` pattern and short-circuit inside `fetchQuotedCallsFees`:
```ts
const isPredicateRoute = await warpCore.isPredicateSupported(originToken, destination);
if (isPredicateRoute) return null;
```
That keeps `shouldFetch` synchronous. Cost is one extra adapter check per fetch, identical to what `useFeeQuotes.ts:149` already does. The `data: null` then makes `offchainUnavailable = true` in `ReviewDetails`, the onchain quote takes over, and the review matches what the user actually signs.

(Alternatively, surface a sync flag derived from the warp registry config — `warpCore.hasPredicateWrapper(token)` style — but that's an SDK-side change. The async-in-queryFn approach is the small one and matches the precedent.)

paulbalaji (commented):
## Re-review (commit `d8374a75`)

Round 4 cleanly resolves all five of my round-3 items:

- ✅ `quotedCalls`/`attestation` mutual-exclusion at the SDK call site (one-liner at `useTokenTransfer.ts:212`)
- ✅ Cache identity: `addressOrDenom` added to queryKey + `data` suppressed when `!shouldFetch || isStale`
- ✅ Quote age guard via `MAX_QUOTE_AGE_MS = 4*60_000` + `issuedAt` stamp; thoughtful UX choice to keep stale fees masked while refetch is in flight, only falling back to onchain once retries exhaust
- ✅ Approval amount inflated by `tokenFeeQuote.amount` when fee is denominated in the transferred token (mirrors `extractQuoteTotals` semantics in the SDK)
- ✅ Rate-limit invariant comment added

One remaining item below. The two issues CodeRabbit just raised (block submit while quote is loading, runtime hex/size validation in `quote.ts`) are valid — deferring to those threads rather than duplicating.

## Still open

Thanks again to codex for catching this on the latest pass.

🟠 **MEDIUM — predicate routes still preview QuotedCalls fees and spender, but execution drops them** (inline). The send-time fix at `useTokenTransfer.ts:212` (skip `quotedCalls` when `attestationResult` is set) closes the SDK assert, but the offchain hook itself doesn't gate on predicate-route status. So predicate routes still fetch a signed quote, the review screen shows offchain fees and the QuotedCalls contract as the spender — and at submit time the user gets a different approval popup (predicate router, plus possibly different fee amounts) than what the review previewed.

## Bottom line

With the predicate-preview gate, this is ready to ship. The four prior rounds have been thorough — I'd be happy to approve once that one is addressed (or have a brief discussion if there's reason to prefer a different shape of fix).

src/features/transfer/useQuotedCalls.ts
Xaroz: Fixed in 950435c7 — added a predicate short-circuit at the top of `fetchQuotedCallsFees`:
```ts
if (await warpCore.isPredicateSupported(originToken, destination)) return null;
```
The adapter memoizes the wrapper lookup, so this is one RPC on first quote and free on subsequent fetches (and on the matching `useFeeQuotes` call once it takes over).

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -453,18 +453,77 @@ function TokenBalance({
...
+        quotedCalls={quotedCalls}
+      />
+      <ButtonSection
+        isReview={isReview}
+        isValidating={isValidating}
+        setIsReview={setIsReview}
+        cleanOverrideToken={cleanOverrideToken}
+        routeOverrideToken={routeOverrideToken}
+        quotedCallsParams={quotedCalls.quotedCallsParams}
+      />
```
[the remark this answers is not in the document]
Xaroz: Fixed in 950435c7 — instead of disabling Send (would feel inconsistent with Continue, which also doesn't gate on quote loading), the submit handler now awaits any in-flight offchain quote before triggering transactions:
```ts
// useQuotedCalls return adds:
getQuotedCallsParams: async () => isLoading ? (await refetch()).data?.quotedCallsParams ?? null : cached
```
```ts
// triggerTransactionsHandler:
const quotedCallsParams = await getQuotedCallsParams();
await triggerTransactions(values, routeOverrideToken, quotedCallsParams);
```
Common path (cached): zero extra latency. First-load or stale-refetch: handler waits → submits with correct params → no silent fallback.

src/pages/api/quote.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  try {
+    const response = await client.getQuote({
+      command: command as FeeQuotingCommand,
+      origin,
+      router: router as `0x${string}`,
+      destination: destinationDomainId,
+      salt: salt as `0x${string}`,
+      recipient: recipient as `0x${string}` | undefined,
+      targetRouter: targetRouter as `0x${string}` | undefined,
```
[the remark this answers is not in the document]
Xaroz: Fixed in 950435c7 — added runtime hex validation at the proxy boundary. Note: `isValidAddressEvm` is only correct for `router` (origin is EVM-gated, 20-byte address). `salt` is a keccak hash, `recipient` and `targetRouter` are `addressToBytes32` wire encodings — all 32 bytes — so they're validated against `/^0x[0-9a-fA-F]{64}$/` rather than `isValidAddressEvm` (which would reject the bytes32-encoded form of any non-EVM destination address).
```ts
if (!isValidAddressEvm(router)) return res.status(400).json({ message: 'router must be a valid EVM address' });
if (!HEX_BYTES32.test(salt)) return res.status(400).json({ message: 'salt must be 32-byte hex' });
if (recipient && !HEX_BYTES32.test(recipient)) return res.status(400).json({ message: 'recipient must be 32-byte hex' });
if (targetRouter && !HEX_BYTES32.test(targetRouter)) return res.status(400).json({ message: 'targetRouter must be 32-byte hex' });
```
Tests cover each rejection path.

paulbalaji (approved):
## LGTM (commit `950435c7`)

Round 5 cleanly resolves all three outstanding items:

- ✅ Predicate short-circuit inside `fetchQuotedCallsFees` (`useQuotedCalls.ts:205`) — exactly the pattern `useFeeQuotes.ts:149` uses. Adapter-side memoization keeps it cheap. Review now previews the same path the user signs.
- ✅ `getQuotedCallsParams()` async accessor wired through `ButtonSection` (`TransferTokenForm.tsx:635`) so a Send click during the first-load / stale-refetch window awaits the in-flight quote instead of silently falling through. Good call to keep button availability consistent with Continue rather than disabling — the cached fast-path adds zero wait.
- ✅ Runtime hex validation at the proxy boundary (`quote.ts:48-59`), correctly distinguishing the EVM `router` (20 bytes via `isValidAddressEvm`) from the bytes32 wire encodings (`salt`/`recipient`/`targetRouter` via the `HEX_BYTES32` regex). Tests cover each rejection path.

Five rounds of reviewer feedback addressed thoughtfully, with explicit reasoning on the trade-off calls (salt-refetch skip, refetch UX during stale state, button-availability vs await-at-submit). Approving — happy to ship.

PR #1030, opened 2026-03-25
perf: narrow runtime/bootstrap loading in warp ui

paulbalaji:
## Summary

- rebased on latest `main`
- updated to released packages:
  - `@hyperlane-xyz/registry@24.3.0`
  - `@hyperlane-xyz/sdk@33.0.2`
  - `@hyperlane-xyz/utils@33.0.2`
  - `@hyperlane-xyz/widgets@33.0.2`
- uses the released widgets follow-ups from `hyperlane-xyz/hyperlane-monorepo#8584` and `hyperlane-xyz/hyperlane-monorepo#8587`
- keeps the perf layer reviewable: lighter package surfaces, deferred wallet-provider mount, narrower bootstrap/runtime split

## What changed

1. move Warp UI off fat monorepo root surfaces where lighter subpaths exist
2. keep wallet providers behind a client-only boundary in `_app`
3. split bootstrap/read state from heavier runtime state
4. keep the runtime bridge explicit instead of leaking provider/runtime setup through general app code
5. narrow hot-path wallet address lookup and balance reads without forcing all-VM wrappers back into the initial graph
6. use released widgets helpers/subpaths where they now exist (`walletAddresses`, `animations/Skeleton`)

## Canonical local benchmark

Measured serially, one build at a time, from fresh temp worktrees with:
- `pnpm install`
- `pnpm clean`
- `/usr/bin/time -p pnpm build`

| State | Build time | Compile time | `.next` total | `.next/static/chunks` | static chunk JS bytes | `.next/server` |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| current `main` (`585e45d2`) | `82.70s` | `65s` | `1,866,176 KB` | `315,616 KB` | `322,096,773` | `304,436 KB` |
| this branch (`6787ac55`) | `57.48s` | `51s` | `1,883,412 KB` | `150,224 KB` | `152,526,813` | `409,296 KB` |

Interpretation:
- materially better than latest `main` on clean local build time
- static client chunk JS is much smaller than `main`
- `.next/server` is materially larger than `main`
- overall `.next` total is slightly larger than `main`

## Validation

- [x] `pnpm typecheck`
- [x] `pnpm lint`
- [x] `pnpm build`

## Follow-ups

- stacked regression coverage for route-token/wallet-flow correctness: `#1061`
- if we want broader wallet-flow confidence next, the right follow-up is deterministic mocked-wallet coverage plus a very small Synpress MetaMask smoke layer, not widening this PR further

## Remaining work

- biggest remaining cost is still default-route/runtime exposure through the runtime bridge, not submit-time transfer code
- next meaningful follow-up would be more aggressive metadata-first/default-route runtime separation, not more small import reshuffles

---

> [!NOTE]
> **Medium Risk**
> Moderate risk due to significant changes in core state initialization/hydration and explicit lazy runtime gating across transfer/fee/balance paths; failures could surface as runtime-not-ready errors or incorrect token/address lookups.
> 
> **Overview**
> **Introduces an explicit bootstrap vs runtime split for Warp context**: `store.ts` is refactored to hydrate a minimal context first (`multiProvider` + metadata-derived token graph), then lazily load the heavier runtime (`WarpCore`, protocol provider builders) via new `storeInit.ts`/`storeRuntime.ts` and `ensureWarpRuntime()`.
> 
> **Reduces eager dependencies and improves retryability of lazy loaders** by adding `sdkRuntime.ts`/`sdkTokenRuntime.ts` promise caches that clear on rejection, and by switching many `@hyperlane-xyz/sdk`/`@hyperlane-xyz/widgets` imports to narrower subpaths.
> 
> **Updates transfer/balance/wallet flows to work with the lazy runtime**: read-only views use `useReadyWarpCore`/`routeTokens`, async actions (validation, submit, fee quotes, approvals, transfers) explicitly call `ensureWarpRuntime()`, and EVM balance reads can bypass SDK token fetchers using direct `wagmi` balance queries with chain-aware address normalization.
> 
> **Config/build changes**: adds `@provablehq/sdk` + `@provablehq/wasm`, marks them as `serverExternalPackages`, updates Next route typings import path, and adds new wallet provider composition in `WalletProviders`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c2236bd3bedc314ad7efcb72378396b37f029c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference path="./.next/types/routes.d.ts" />
+import "./.next/types/routes.d.ts";
```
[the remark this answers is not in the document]
paulbalaji: I checked this on the current `#1030` head before changing anything. On this Next 16 line, `pnpm build` regenerates `next-env.d.ts` back to the same `import "./.next/types/routes.d.ts";` form, and the worktree stays clean afterward. So while the general warning is fair, reverting it here would just churn back on the next build. I left it as-is to match the generator output on this branch.

src/features/chains/ChainFilterPanel.tsx
```diff
@@ -1,13 +1,11 @@
-import { ChainName } from '@hyperlane-xyz/sdk';
+import type { ChainName } from '@hyperlane-xyz/sdk/types';
```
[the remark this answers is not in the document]
paulbalaji: Checked this against the current branch and left it as-is. These `ChainName` comments are style-only, and the current review set is internally inconsistent: some comments ask to remove explicit imports in favor of `src/global.d.ts`, while another asks to delete the global declaration and keep explicit imports. I’m avoiding churn until we make one repo-wide decision on type source of truth.

src/features/chains/ChainList.tsx
```diff
@@ -1,5 +1,5 @@
-import { ChainName } from '@hyperlane-xyz/sdk';
-import { PencilIcon } from '@hyperlane-xyz/widgets';
+import type { ChainName } from '@hyperlane-xyz/sdk/types';
```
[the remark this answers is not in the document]
paulbalaji: Same reasoning here: no functional issue, only type-source style. I’m not changing individual files piecemeal while the current review set conflicts on whether `ChainName` should come from globals or explicit imports. Better handled as one repo-wide cleanup, not in this perf PR.

src/features/chains/MobileChainQuickSelect.tsx
```diff
@@ -1,4 +1,4 @@
-import { ChainName } from '@hyperlane-xyz/sdk';
+import type { ChainName } from '@hyperlane-xyz/sdk/types';
```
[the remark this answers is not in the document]
paulbalaji: I checked this and intentionally did not remove the global declaration in this PR. That would be a repo-wide typing migration, not a local perf/correctness fix, and it conflicts with the other CodeRabbit comments on this same PR that ask to remove explicit imports and rely on `src/global.d.ts`. Leaving the current type setup unchanged is the least-churn option here.

src/features/tokens/TokenListPanel.tsx
```diff
@@ -1,4 +1,5 @@
-import { ChainName, Token } from '@hyperlane-xyz/sdk';
+import type { Token } from '@hyperlane-xyz/sdk/token/Token';
+import type { ChainName } from '@hyperlane-xyz/sdk/types';
```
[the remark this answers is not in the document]
paulbalaji: Checked; left unchanged for the same reason as the other `ChainName` comments. This is style-only, not a correctness or perf issue, and the current review set is contradictory on whether globals or explicit imports are the canonical source.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -1,6 +1,9 @@
-import type { ChainName } from '@hyperlane-xyz/sdk';
+import type { ChainName } from '@hyperlane-xyz/sdk/types';
```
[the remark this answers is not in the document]
paulbalaji: Checked and intentionally left as-is. This is not a functional issue, and I don’t want to mix a repo-wide typing-style cleanup into this PR while the current comments disagree on the target convention.

src/features/messages/useMergedTransferHistory.ts
```diff
@@ -1,10 +1,12 @@
-import type { ChainName, MultiProtocolProvider, WarpCore } from '@hyperlane-xyz/sdk';
+import type { MultiProviderAdapter as MultiProtocolProvider } from '@hyperlane-xyz/sdk/providers/MultiProviderAdapter';
+import type { IToken } from '@hyperlane-xyz/sdk/token/IToken';
+import type { ChainName } from '@hyperlane-xyz/sdk/types';
```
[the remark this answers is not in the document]
paulbalaji: Checked; same call as the other `ChainName` threads. This is style-only and not a correctness/perf issue. I’m intentionally not doing piecemeal import churn here while the review set is inconsistent about whether globals or explicit imports are the canonical source.

paulbalaji:
Follow-up from Jason’s Slack review, now checked against the branch.

I pushed one real fix in `99fb92f2`:
- the recipient/sender smart-contract check in `TransferTokenForm` was using the current store `multiProvider`, which can still be metadata-only on this branch
- that means `isSmartContract()` could call `getViemProvider(...)` before the runtime-backed provider existed and fail noisily
- it now awaits `ensureWarpRuntime()` first and runs the contract check against the runtime-backed provider instead

On the other two points:
- the extra wallet-address glue in `balances/hooks.ts` is intentional. It replaced the heavier all-VM `multiProtocol` hot-path lookup with narrower protocol-specific hooks. So yes, there is more app-local coordination code now, but it exists to avoid paying the old generic wrapper path on the default route.
- the reusable part of that address-helper shape is now being upstreamed in `hyperlane-xyz/hyperlane-monorepo#8584`, so this should not need to stay app-local permanently
- `useDirectEvmBalance()` is intentionally EVM-only. EVM has a cheap direct wagmi balance path; the other VMs do not have an equivalent lightweight hook already wired into this app, so they still use the generic SDK balance path. This is not saying other VMs are unimportant, only that EVM was the first protocol where a clear lower-cost fast path already existed.

So current state is:
- smart-contract warning path: fixed
- wallet glue: deliberate hot-path tradeoff, with reusable helper follow-up now in `hyperlane-xyz/hyperlane-monorepo#8584`
- direct balance fast path: EVM-only by design for now

src/features/balances/hooks.ts
```diff
@@ -103,17 +127,129 @@ export function useEvmWalletBalance(
...
 /** Returns a Map<ProtocolType, string> of all connected wallet addresses. */
 function useWalletAddresses(multiProvider: MultiProtocolProvider): Map<ProtocolType, string> {
-  const evmAddress = useEthereumAccount(multiProvider).addresses[0]?.address;
-  const solanaAddress = useSolanaAccount(multiProvider).addresses[0]?.address;
+  const walletAddresses = useDirectWalletAddresses(multiProvider);
 
   return useMemo(() => {
     const map = new Map<ProtocolType, string>();
+    const evmAddress = getAddressForChain(walletAddresses.get(ProtocolType.Ethereum));
+    const solanaAddress = getAddressForChain(walletAddresses.get(ProtocolType.Sealevel));
```
Xaroz: feels like we are adding an extra layer instead of just calling useEthereumAccount/useSolanaAccount now we would have to call useDirectWalletAddress, create a map, and then call getAddressForChain and pass the protocol type

src/features/balances/hooks.ts
```diff
@@ -36,13 +49,23 @@ export function useBalance(chain?: ChainName, token?: IToken, address?: Address)
...
-  useToastError(error, 'Error fetching balance');
+  useToastError(directEvmBalance.error ?? error, 'Error fetching balance');
+
+  if (directEvmBalance.isEnabled) {
+    return {
+      isLoading: directEvmBalance.isLoading,
+      isError: directEvmBalance.isError,
+      balance: directEvmBalance.balance,
+    };
+  }
```
Xaroz, line 79: why do we need to do this and would this be required for every other VMs?

Xaroz: I guess my point is more like with the these new updates we need to individually declare more stuff instead of having the functions pick the correct protocol to use and that introduces overhead and I am not sure if the bundle size reduction compensates for this

src/features/balances/hooks.ts
paulbalaji: Agreed. I flattened this in `692d1c9e`.

I removed the `useDirectWalletAddresses(...)` + intermediate `Map` from the hot path. `useWalletAddressForChain(...)` now reads the protocol-specific wallet hooks directly and switches once at the callsite, and `useWalletAddresses(...)` now only reads the EVM/Solana hooks it actually needs.

So the protocol-scoped behavior stays, but the extra local layer is gone.

src/features/balances/hooks.ts
paulbalaji: This one is only for the EVM fast path.

The early return is there so that once wagmi can answer the balance directly, we do not also run the generic SDK `token.getBalance(multiProvider, ...)` query in parallel and then surface the wrong loading/error state from the slower path.

It is not something we need to copy to every VM. EVM already has a cheap direct balance source in wagmi, so it is worth specializing. The other VMs still stay on the generic SDK balance path until there is an equivalent lightweight direct hook worth using.

On the broader overhead point: I agree we should avoid adding local glue unless it buys us something. That is why I flattened the address-selection layer in `692d1c9e` rather than keeping the extra map indirection.

src/features/hyperlane/sdkRuntime.ts
```diff
@@ -0,0 +1,135 @@
...
+import type { ChainMap } from '@hyperlane-xyz/sdk/types';
+import { type KnownProtocolType, ProtocolType } from '@hyperlane-xyz/utils';
+
+type RuntimeMultiProvider = MultiProviderAdapter<{ mailbox?: string }>;
+
+type SdkRuntime = {
+  createMultiProvider: (
+    chainMetadata: ChainMap<ChainMetadata<{ mailbox?: string }>>,
+  ) => RuntimeMultiProvider;
+  WarpCore: typeof import('@hyperlane-xyz/sdk/warp/WarpCore').WarpCore;
```
Xaroz: rename to `warpCore`?

src/features/hyperlane/sdkRuntime.ts
```diff
@@ -0,0 +1,135 @@
...
+      ) {
+        return (urls, network) => ({
+          type: ProviderType.EthersV5,
+          provider: defaultTronEthersProviderBuilder(urls, network),
+        });
+      }
+
+      return this.providerBuilders[type];
+    }
+  }
```
Xaroz, line 75: move this class out of the function

src/features/hyperlane/sdkTokenRuntime.ts
```diff
@@ -0,0 +1,13 @@
+type TokenClass = typeof import('@hyperlane-xyz/sdk/token/Token').Token;
+
+let tokenPromise: Promise<TokenClass> | undefined;
+
+export function getSdkToken(): Promise<TokenClass> {
```
Xaroz, line 5: this feels a bit iffy, wouldn't it try to load the token each time this gets called and this seems more error prone than just having the token

src/features/tokens/routeTokens.ts
```diff
@@ -0,0 +1,40 @@
...
+      const connectedToken = tokens.find(
+        (candidate) =>
+          candidate.chainName === chainName &&
+          candidate.addressOrDenom === addressOrDenom &&
+          (!token.warpRouteId || candidate.warpRouteId === token.warpRouteId),
+      );
+
+      if (!connectedToken) {
+        throw new Error(`Connected token not found: ${chainName} ${addressOrDenom}`);
+      }
```
Xaroz: use assert here

src/features/tokens/routeTokens.ts
```diff
@@ -0,0 +1,40 @@
+import { parseTokenConnectionId } from '@hyperlane-xyz/sdk/token/TokenConnection';
+import { TokenMetadata } from '@hyperlane-xyz/sdk/token/TokenMetadata';
+import type { WarpCoreConfig } from '@hyperlane-xyz/sdk/warp/types';
+
+export function buildRouteTokens(config: WarpCoreConfig): TokenMetadata[] {
```
Xaroz: this seems exactly what WarpCore.FromConfig, if you already have warpcore.tokens why do you need to do this again?

src/features/transfer/maxAmount.ts
```diff
@@ -22,10 +28,14 @@ interface FetchMaxParams {
 
 export function useFetchMaxAmount() {
   const multiProvider = useMultiProvider();
-  const warpCore = useWarpCore();
+  const ensureWarpRuntime = useStore((s) => s.ensureWarpRuntime);
```
Xaroz, line 31: I think a better name would be getRuntimeWarpCore, but this is just a small nit

src/features/transfer/TransferFeeModal.tsx
```diff
@@ -1,10 +1,15 @@
...
 import Link from 'next/link';
 
 import { links } from '../../consts/links';
 import { UsdLabel } from '../balances/UsdLabel';
-import { FeePrices } from '../balances/useFeePrices';
+import type { FeePrices } from '../balances/useFeePrices';
+
+function LoadingSkeleton({ className }: { className: string }) {
+  return <div className={`rounded bg-gray-200/70 ${className}`} />;
+}
```
Xaroz: any reason we are just not reusing the Skeleton already imported?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -551,7 +568,15 @@ function ButtonSection({
     return () => {
       isMounted = false;
     };
-  }, [recipient, connectedWallet, multiProvider, originToken, destinationToken, chainDisplayName]);
+  }, [
+    recipient,
+    connectedWallet,
+    multiProvider,
```
Xaroz: might need to check the dependencies, you have runtimeMultiProvider inside here but using multiProvider as a dependency too

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -657,16 +682,16 @@ function ReviewDetails({
   routeOverrideToken: Token | null;
 }) {
   const { values } = useFormikContext<TransferFormValues>();
-  const warpCore = useWarpCore();
+  const warpCore = useReadyWarpCore();
```
Xaroz, line 710: maybe need to tighten up the dependencies we have `const ensureWarpRuntime = useStore((s) => s.ensureWarpRuntime);` and then also `useReadyWarpCore`

src/features/hyperlane/sdkRuntime.ts
paulbalaji: Renamed in `ab4f2f99`. The runtime loader now returns `warpCore`, and the store runtime callsite was aligned to match.

src/features/hyperlane/sdkRuntime.ts
paulbalaji: I kept this scoped to `createSdkRuntime()` intentionally.

Moving it to module scope would require a top-level value import of `MultiProviderAdapter`, which would defeat the lazy runtime split and pull that SDK path back into the initial graph. I added a comment in `c050b587` to make that constraint explicit.

src/features/hyperlane/sdkTokenRuntime.ts
paulbalaji: This stays memoized. `getSdkToken()` only creates the dynamic import promise once, returns the same promise to concurrent callers, and clears it on rejection so a transient failure can retry instead of poisoning the cache.

I added that intent comment in `c050b587`.

src/features/tokens/routeTokens.ts
paulbalaji: Switched to `assert(...)` in `c050b587`.

src/features/tokens/routeTokens.ts
paulbalaji: This is the metadata-first path before the runtime `WarpCore` exists.

`WarpCore.FromConfig(...)` gives us the full runtime token graph, but `storeInit` is intentionally trying to avoid building that upfront. `buildRouteTokens(...)` reconstructs the lightweight `TokenMetadata` graph from config so the init path can still assemble token lists, collateral groups, and router-address maps without eagerly loading runtime/provider code.

I added that comment in `c050b587`.

src/features/transfer/maxAmount.ts
paulbalaji: Tightened the local naming in `c050b587` to `runtimeWarpCore` so it is clearer this is the ensured runtime instance, not the store selector.

src/features/transfer/TransferFeeModal.tsx
paulbalaji: I checked this. There is a `Skeleton` in widgets, but not on a narrow exported subpath.

Using it here would mean pulling from the widgets root barrel for a trivial two-line placeholder. Since this PR is explicitly trying to narrow imports on the hot path, I kept the local `LoadingSkeleton` rather than widening the import surface again.

src/features/transfer/TransferTokenForm.tsx
paulbalaji: Yep, good catch. Fixed in `25bda64f`.

That effect only needed `originToken.protocol` / `destinationToken.protocol`, not `multiProvider`, so I removed the metadata lookups and dropped `multiProvider` from the dependency list.

src/features/transfer/TransferTokenForm.tsx
paulbalaji: These two are intentional and I’d keep them separate.

- `useReadyWarpCore()` is the render-safe read path. It only observes whether runtime is already ready and does not trigger the lazy upgrade during render.
- `ensureWarpRuntime()` is the imperative boundary for async paths that are already doing work anyway (`validate`, fee quote fetch, contract checks, submit path).

If I collapse those, either render paths start eagerly forcing runtime again, or the async action paths stop being explicit about where the lazy upgrade happens. So I’m treating them as two different APIs for two different phases: observe vs ensure.

src/features/store.ts
```diff
@@ -42,10 +26,12 @@ export interface RouterAddressInfo {
...
 
 interface WarpContext {
+  isWarpContextReady: boolean;
   registry: IRegistry;
   chainMetadata: ChainMap<ChainMetadata>;
-  multiProvider: MultiProtocolProvider;
-  warpCore: WarpCore;
+  multiProvider: MultiProtocolProvider | undefined;
+  warpCore: WarpCore | undefined;
+  tokensBySymbolChainMap: Record<string, TokenChainMap>;
```
Xaroz: I think I removed this because it is not being used anymore unless you need this

src/features/transfer/TransfersDetailsModal.tsx
```diff
@@ -125,8 +124,8 @@ export function TransfersDetailsModal({
 
   const isAccountReady = !!account?.isReady;
   const connectorName = walletDetails.name || 'wallet';
-  const token = tryFindToken(warpCore, origin, originTokenAddressOrDenom);
-  const destToken = tryFindToken(warpCore, destination, destTokenAddressOrDenom);
+  const token = tryFindTokenInTokens(tokens, origin, originTokenAddressOrDenom);
+  const destToken = tryFindTokenInTokens(tokens, destination, destTokenAddressOrDenom);
```
Xaroz: one thing to note is that theres useTokens and then theres useWarpCoreTokens, 

tokens = deduplicated
warpcore tokens = all of them

so in this case I think its trying to use all the tokens

src/features/wallet/SideBarMenu.tsx
```diff
@@ -101,9 +104,9 @@ export function SideBarMenu({
         if (item.type === TransferItemType.Local) return true;
         const originChain = multiProvider.tryGetChainName(item.data.originDomainId);
         if (!originChain) return false;
-        return !!tryFindToken(warpCore, originChain, item.data.sender);
+        return !!tryFindTokenInTokens(tokens, originChain, item.data.sender);
```
Xaroz: ditto

src/features/store.ts
paulbalaji: We still need it.

This `multiProvider` is the metadata-only provider that gets set during `storeInit`, before runtime `WarpCore` exists. A lot of the app still reads chain metadata / display names / account hooks / explorer URLs off that path before any transfer runtime is needed.

If we remove it here, we either break those read paths or force runtime initialization earlier again, which is the opposite of what this split is trying to do.

src/features/transfer/TransfersDetailsModal.tsx
paulbalaji: Intentional here.

`TransfersDetailsModal` is a read-only/history surface, so I want it to stay on the deduplicated metadata token set from `useTokens()` rather than force full runtime `warpCore.tokens`.

For this modal we only need address -> token display metadata lookup (`symbol`, `icon`, `decimals` fallback, etc.). The deduplicated set is enough for that because the lookup key is still `chain + addressOrDenom`. If this view ever needs route-specific runtime token semantics, then `useWarpCoreTokens()` would make sense, but that would be a different requirement.

src/features/wallet/SideBarMenu.tsx
paulbalaji: Same reasoning here.

`SideBarMenu` is also a read/history path, so it stays on `useTokens()` intentionally. It only needs enough token metadata to render history rows and map router/message addresses back to known tokens. Pulling `useWarpCoreTokens()` into the sidebar would couple it back to the full runtime graph for no extra value on this path.

src/features/transfer/TransferFeeModal.tsx
Xaroz: would prefer then if we can fix the imports in the widgets side, we still wanna surface common components from the library instead of doing single implementations where we can

src/features/transfer/TransferTokenForm.tsx
Xaroz: can we add some comments explaining when to use each?

src/features/transfer/TransfersDetailsModal.tsx
Xaroz: Hmm I think this is incorrect, because when transfers are done the choosing logic would pick the correct token depending on the circumstances (lowest fee, enough collateral, default fpwr) so if you use the deduplicated set it would not find it because the `addressOrDenom` would be different

src/features/store.ts
Xaroz: sorry I am talking about `tokensBySymbolChanMap`

src/features/transfer/TransferTokenForm.tsx
paulbalaji: Added in `7e020cd4`.

- `useReadyWarpCore()` now has an explicit comment that it is the render-safe observe path.
- `ensureWarpRuntime()` is commented at the callsite in `TransferTokenForm` as the imperative async upgrade boundary.

That should make the intended split clearer without collapsing the two APIs.

src/features/transfer/TransferFeeModal.tsx
paulbalaji: Agreed. I opened the narrow widgets follow-up here: hyperlane-xyz/hyperlane-monorepo#8587.

That PR only adds `@hyperlane-xyz/widgets/animations/Skeleton` so consumers can reuse the shared component without widening back to the root widgets barrel. I kept the local placeholder in this PR for now so we do not block on an unpublished package.

src/features/store.ts
paulbalaji: Yep, you were right. `tokensBySymbolChainMap` was dead.

I removed it in `1e8c2e36` from the store shape and from both the init/runtime assembly paths.

src/features/transfer/TransfersDetailsModal.tsx
paulbalaji: Agreed. My earlier reply was too loose.

I fixed this in `1e8c2e36` by switching these history/detail address lookups off the deduplicated token set and onto `routeTokens` instead.

So now:
- pickers/UI still use deduplicated `tokens`
- history/details/sidebar address resolution use raw route tokens

That keeps the read path metadata-only, but preserves exact route-token lookups when same-symbol routes collapse differently.

src/features/transfer/TransferFeeModal.tsx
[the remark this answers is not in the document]
paulbalaji: Deferring this one intentionally. We already opened the narrow widgets follow-up in `hyperlane-xyz/hyperlane-monorepo#8587` to add `@hyperlane-xyz/widgets/animations/*` subpath exports. Once that lands and is released, this modal can switch to the shared `Skeleton` without widening back to the root widgets barrel. I’m leaving the local placeholder here until that package is available, so this PR stays unblocked and does not reintroduce the hot-path import broadening it is trying to remove.

src/features/transfer/TransfersDetailsModal.tsx
Xaroz: not mentioned in your reply but the transfer picking should be done with the raw route tokens to pick the most optimal one

paulbalaji: Yes, agreed. That is the actual intent of the fix.\n\nThe current code in `TransfersDetailsModal` is already doing the lookup against raw route tokens, not the deduplicated token list:\n- `const routeTokens = useRouteTokens();`\n- `tryFindTokenInTokens(routeTokens, origin, originTokenAddressOrDenom)`\n- `tryFindTokenInTokens(routeTokens, destination, destTokenAddressOrDenom)`\n\nSo the important behavior here is exactly what you called out: history/detail lookup needs the raw route-token set so it can resolve the concrete token that was actually picked at transfer time, including cases where route selection differed because of collateral / fee / fpwr choice. My earlier reply was too imprecise, but the code path is now on the correct raw-token behavior.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -910,7 +937,7 @@ async function validateForm(
     }
 
     const { address: sender, publicKey: senderPubKey } = getAccountAddressAndPubKey(
-      warpCore.multiProvider,
+      multiProvider,
```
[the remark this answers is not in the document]
paulbalaji: Fixed in `fbc7e7e8`.

All three router-address lookup sites now normalize with the token protocol before hitting `routerAddressesByChainMap`, so case-sensitive non-EVM addresses use the same protocol-aware key shape as the map builder.

src/features/transfer/TransfersDetailsModal.tsx
```diff
@@ -271,7 +266,7 @@ export function TransfersDetailsModal({
               {destToken && (
                 <>
                   <Image className="mx-2" src={ArrowRightIcon} width={10} height={10} alt="" />
-                  <span>{destAmount || amount}</span>
+                  <span>{amount}</span>
```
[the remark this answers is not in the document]
paulbalaji: Fixed in `fbc7e7e8`.

`TransfersDetailsModal` now restores `computeDestAmount(...)` and uses the scaled destination amount for the received-amount display instead of echoing the origin amount.

src/features/tokens/hooks.ts
```diff
@@ -15,6 +20,16 @@ import { useStore } from '../store';
 import { getTokenKey } from './utils';
 
 export function useWarpCore() {
+  const warpCore = useStore((s) => s.warpCore);
+  if (!warpCore) {
+    throw new Error('Warp context not ready');
+  }
+  return warpCore;
+}
```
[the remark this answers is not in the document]
paulbalaji: • Not fixing this.

  Reason:

  - the premise is stale on current code
  - WarpContextInitGate intentionally gates on metadata/provider readiness, not warpCore
  - runtime-required render paths already use /Users/pbio/work/hwui-1030-rebase/src/features/tokens/hooks.ts or
    ensureWarpRuntime()

  Current usage check:

  - /Users/pbio/work/hwui-1030-rebase/src/features/tokens/hooks.ts has no live callers
  - /Users/pbio/work/hwui-1030-rebase/src/features/tokens/hooks.ts also has no live callers
  - active callers use:
      - /Users/pbio/work/hwui-1030-rebase/src/features/tokens/hooks.ts in fee quotes / approvals / transfer form

  So:

  - no current render-time crash path
  - changing the gate to wait for warpCore would undo the metadata-first split
  - changing useWarpCore() to silently return undefined/[] would weaken invariants

  I’d reply that this is by design:

  - useWarpCore() is the strict runtime-required hook
  - useReadyWarpCore() is the render-safe observation hook
  - ensureWarpRuntime() is the imperative async boundary

src/features/analytics/utils.ts
```diff
@@ -85,8 +87,9 @@ export function trackTransactionFailedEvent(
   const token = overrideToken || warpCore.tokens.find((t) => getTokenKey(t) === originTokenKey);
   if (!token) return;
 
+  const multiProvider = warpCore.multiProvider as MultiProtocolProvider;
```
Xaroz: do we need need assertion here?

src/features/balances/hooks.ts
```diff
@@ -101,17 +144,128 @@ export function useEvmWalletBalance(
...
+      case ProtocolType.Starknet:
+        return getAddressForChain(starknetAddresses, protocol, chainName);
+      case ProtocolType.Radix:
+        return getAddressForChain(radixAddresses, protocol, chainName);
+      case ProtocolType.Aleo:
+        return getAddressForChain(aleoAddresses, protocol, chainName);
+      case ProtocolType.Tron:
+        return getAddressForChain(tronAddresses, protocol, chainName);
+      default:
+        return undefined;
```
Xaroz: I feel like this could live inside a function outside of the component as an util function or even better just a function that returns the address depending on the protocol and then you only need to call `getAddressForChain` once

src/features/balances/hooks.ts
```diff
@@ -101,17 +144,128 @@ export function useEvmWalletBalance(
...
+    },
+  });
+
+  return {
+    isEnabled,
+    error,
+    isError,
+    isLoading,
+    balance: data && token ? new TokenAmount(data.value, token) : undefined,
+  };
```
Xaroz, line 262: hmm maybe this should live inside the `balances/evm.ts` file, remind me why we need `useWagmiBalance` if we already have `fetchChainBalances` in evm.ts? or is this for other purpose

src/features/chains/ChainEditModal.tsx
Xaroz, line 43: interesting there seems to be a eslint warning here

src/features/chains/hooks.ts
```diff
@@ -1,20 +1,27 @@
...
 import { useStore } from '../store';
 import { getChainDisplayName } from './utils';
 
-export function useMultiProvider() {
-  return useStore((s) => s.multiProvider);
+export function useMultiProvider(): MultiProtocolProvider {
+  const multiProvider = useStore((s) => s.multiProvider);
+  if (!multiProvider) {
+    throw new Error('Warp context not ready');
+  }
```
Xaroz, line 15: wouldn't you use `useReadyMultiProvider` for cases like this?

src/features/chains/utils.ts
```diff
@@ -41,7 +40,7 @@ export function getNumRoutesWithSelectedChain(
   selectedChain: ChainName,
   isSelectedChainOrigin: boolean,
 ): ChainSearchMenuProps['customListItemField'] {
-  const multiProvider = warpCore.multiProvider;
+  const multiProvider = warpCore.multiProvider as MultiProtocolProvider;
```
Xaroz: ditto on the assertion here

src/features/chains/ChainEditModal.tsx
paulbalaji: ya the warning is pre existing but i've fixed it on troy's PR
https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/996/changes/303c9eefcb2256bf8d1a87239257678bfbc06754#diff-c7c0c5080d1bbca60edc9e06123164855ce810d8a94c8174a6cb6fc0ab9b5a46

src/features/analytics/utils.ts
paulbalaji: Agreed. The `as MultiProtocolProvider` cast was unnecessary here, so I removed it. `warpCore.multiProvider` already satisfies the helper path we use.

src/features/chains/hooks.ts
paulbalaji: Not here. `useMultiProvider()` is the strict hook for callsites that require the provider and should fail loudly if they render too early. `useReadyMultiProvider()` is the render-safe variant for paths that can tolerate `undefined`. This file intentionally exposes both.

src/features/chains/utils.ts
paulbalaji: Agreed. Removed the extra cast here too. `warpCore.multiProvider` already satisfies this path directly.

src/features/balances/hooks.ts
paulbalaji: Different purpose. `fetchChainBalances(...)` in `balances/evm.ts` is the batched multicall path for list reads across many tokens/chains. This hook is the single connected-wallet hot path on the form, so I kept it on `useWagmiBalance(...)` for the active EVM wallet/client instead of routing it through the broader batch path.

src/features/tokens/utils.ts
```diff
@@ -250,7 +259,10 @@ export function groupTokensByCollateral(tokens: Token[]): Map<string, Token[]> {
  *
  * Results are cached by token object reference for O(1) subsequent lookups.
  */
-export function getCollateralKey(token: IToken): string {
+export function getCollateralKey(
+  token: IToken,
+  resolvedMap: Map<string, string> = resolvedUnderlyingMap,
```
Xaroz: why do we need to include a params for resolvedMap?

paulbalaji: We still need it for the init/rebuild path. `storeInit` / `storeRuntime` compute a fresh `resolvedMap` before the module-global map is installed, and `buildTokensArray(...)` / `groupTokensByCollateral(...)` need to key against that fresh map immediately. If `getCollateralKey(...)` only read the module-global state, correctness would depend on `setResolvedUnderlyingMap(...)` having already run, which is the ordering dependency this change was avoiding. Normal callsites still just use the default global map.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -51,30 +50,40 @@ export function useTokenTransfer(onDone?: () => void) {
...
-        updateTransferStatus,
-        setIsLoading,
-        onDone,
-        routeOverrideToken,
-      }),
+    async (values: TransferFormValues, routeOverrideToken: Token | null) => {
+      try {
+        const warpCore = await ensureWarpRuntime();
+        if (!warpCore) throw new Error('Warp runtime not ready');
+        const transferIndex = useStore.getState().transfers.length;
```
Xaroz, line 57: why not just get `transfers` from the destructured `useStore` above like it was before?

src/features/warpCore/warpCoreConfig.ts
```diff
@@ -26,14 +22,25 @@ type NullableAddressWarpCoreToken = Omit<WarpCoreToken, 'addressOrDenom'> & {
...
+  return result.data;
+}
+
 export async function assembleWarpCoreConfig(
   storeOverrides: WarpCoreConfig[],
   registry: IRegistry,
 ): Promise<{ config: WarpCoreConfig; wireDecimalsMap: WireDecimalsMap }> {
   const yamlResult = WarpCoreConfigSchema.safeParse(yamlWarpRoutes);
-  const yamlConfig = validateZodResult(yamlResult, 'warp core yaml config');
+  const yamlConfig = validateWarpCoreConfigResult(yamlResult, 'warp core yaml config');
```
Xaroz, line 41: doesn't `validateZodResult` do the same thing as this?

src/features/wallet/context/CosmosWalletContext.tsx
```diff
@@ -24,7 +24,10 @@ const theme = extendTheme({
 export function CosmosWalletContext({ children }: PropsWithChildren<unknown>) {
   const chainMetadata = useMultiProvider().metadata;
   const { chains, assets } = useMemo(() => {
-    const multiProvider = new MultiProtocolProvider({ ...chainMetadata, cosmoshub });
+    const multiProvider: Parameters<typeof getCosmosKitChainConfigs>[0] = new MultiProviderAdapter({
```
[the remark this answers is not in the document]
paulbalaji: I checked the current render tree and this is not a live crash path. `WarpContextInitGate` wraps `WalletProviders` in `_app.tsx`, so `CosmosWalletContext` only renders after the metadata/provider bootstrap gate opens. So `useMultiProvider()` here is intentionally the strict hook. If someone later moves `WalletProviders` outside that gate then this concern becomes real, but that is not the current tree.

src/features/transfer/useTokenTransfer.ts
paulbalaji: Because the index now needs to be read **after** the async `ensureWarpRuntime()` boundary, not from render-time state. If we pull `transfers` into the destructured selector above, the callback closes over whatever length existed when the hook rendered, which is exactly the stale-index race this change was fixing. `useStore.getState().transfers.length` here is intentional because it reads the latest length at execution time, right before we hand off to `executeTransfer(...)`.

src/features/warpCore/warpCoreConfig.ts
paulbalaji: Effectively yes, but this file now just has the specialized local version. There is no current `validateZodResult(...)` helper in this repo path anymore. `validateWarpCoreConfigResult(...)` is doing the same safe-parse unwrap + loud failure, just with the `WarpCoreConfig`-specific type/result here.

Xaroz (commented):
a few comments, but my biggest complain is probably that with these changes the code is harder to reason about (the ensure runtime warp, the split of the store, etc...) and might be hard to get used to in terms of understanding and perhaps future development too 


Edit: and from what I can tell with the benchmarks the performance gains are mostly minimals (except for build and initial chunks vs server loading) for all this overhead

src/features/warpCore/warpCoreConfig.ts
paulbalaji: Yes, same basic behavior. The reason for the local helper is that we removed the broad `@hyperlane-xyz/sdk` root import in this file. I did not want to widen the import surface again just to reuse `validateZodResult(...)`, so this keeps the same safe-parse unwrap behavior locally while staying on the narrower SDK subpath imports used elsewhere in this PR.

src/features/tokens/utils.ts
Xaroz: see my comment about my complain 😄

paulbalaji: Fair. I kept the explicit map threading where it matters (`buildTokensArray(...)` / `groupTokensByCollateral(...)`), but hid it behind a private helper so the public `getCollateralKey(...)` API is simple again. The explicit map is still needed during init/rebuild before the module-global map is installed, but it no longer needs to leak through the exported helper signature.

paulbalaji:
fair. i cba to push this PR any further so i'll close it for anyone in future to resurrect if they want

at the least explorer is improved and we've upstreamed things to the monorepo that had a knock-on effect (which is why the diff to main is much smaller now than when the pr started).

paulbalaji:
On the broader complexity concern: I agree the local ergonomics cost is real, and that was the main trade in this PR. The reason I still think the direction is correct is that the old “pick the right runtime/provider for me” path was hiding broad imports on the hot path. This PR makes that cost explicit and keeps the heavier runtime upgrade behind `useReady*` / `ensure*` boundaries instead. The win is mostly in client/static output and bootstrap/build behavior, not server output. I tried to keep the complexity localized to the store/runtime boundary and upstream shared pieces where possible so app code does not keep re-growing local wrappers. If there are specific spots that still feel like unnecessary ceremony after this pass, I’d rather collapse those than defend them on principle.

PR #1041, opened 2026-04-07
feat: add balance fetching for Cosmos tokens

Xaroz:
## Summary
- Add balance fetching for Cosmos/CosmosNative tokens via `bank.allBalances()` — 1 RPC call per chain instead of N individual queries
- Classify cosmos token standards into bank-batchable (most) vs SDK fallback (`CwHypNative` only)
- Per-chain bech32 address lookup from connected cosmos wallet

## Token Classification
| Category | Standards | Denom Source |
|----------|-----------|-------------|
| Bank (addressOrDenom) | CosmosNative, CosmosIbc, CosmosIcs20, CWNative, CW20, CwHypSynthetic | `addressOrDenom` |
| Bank (collateral) | CwHypCollateral, CosmNativeHypCollateral | `collateralAddressOrDenom` |
| Bank (predictable) | CosmNativeHypSynthetic | `hyperlane/{addressOrDenom}` |
| SDK fallback | CwHypNative | Dynamic via contract query |

## Drive-by
- Add `resolve-pr-reviews` Claude Code skill for interactive PR review resolution


Xaroz:
@claude review

src/features/balances/cosmos.ts
```diff
@@ -0,0 +1,105 @@
...
+ */
+export async function fetchCosmosChainBalances(
+  group: CosmosChainGroup,
+  rpcUrl: string,
+  address: string,
+): Promise<Record<string, bigint>> {
+  if (group.bankTokens.length === 0) return {};
+
+  try {
+    const client = await StargateClient.connect(rpcUrl);
```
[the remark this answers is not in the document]
Xaroz: Fixed in c88e14f6.

src/features/balances/cosmos.ts
```diff
@@ -0,0 +1,105 @@
...
+    const allCoins = await client.getAllBalances(address);
+    const coinMap = new Map(allCoins.map((c) => [c.denom, BigInt(c.amount)]));
+
+    const out: Record<string, bigint> = {};
+    for (const { key, denom } of group.bankTokens) {
+      const balance = coinMap.get(denom);
```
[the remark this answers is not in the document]
Xaroz: Skipping — consistent with EVM and Sealevel behavior where missing key = no balance. No consumer relies on explicit 0n keys.

Xaroz:
@claude review the reply

troykessler (approved):

src/features/balances/cosmos.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/** Standards where the bank denom is `addressOrDenom`. */
+const BANK_DENOM_FROM_ADDRESS: TokenStandard[] = [
+  TokenStandard.CosmosNative,
+  TokenStandard.CosmosIbc,
+  TokenStandard.CosmosIcs20,
+  TokenStandard.CWNative,
+  TokenStandard.CW20,
+  TokenStandard.CwHypSynthetic,
```
paulbalaji: **Bug: `CwHypSynthetic` uses wrong denom source**

`BANK_DENOM_FROM_ADDRESS` causes the bank query to use `token.addressOrDenom`, which is the **warp router address** for this standard.

In SDK 29.0.0, `Token.getAdapter()` builds:
```js
new CwHypSyntheticAdapter(chainName, multiProvider, {
  warpRouter: addressOrDenom,       // ← not used for balance
  token: collateralAddressOrDenom,  // ← this is the bank denom
});
```

So `getBalance()` reads via `collateralAddressOrDenom`, not `addressOrDenom`. This fast path will miss `CwHypSynthetic` balances.

No tokens of this standard exist in the registry today, so not a live bug — but will produce incorrect results when deployed.

**Fix:** move `TokenStandard.CwHypSynthetic` to `BANK_DENOM_FROM_COLLATERAL`.

src/features/balances/cosmos.ts
```diff
@@ -0,0 +1,107 @@
...
+];
+
+/** Standards where the bank denom is `collateralAddressOrDenom`. */
+const BANK_DENOM_FROM_COLLATERAL: TokenStandard[] = [
+  TokenStandard.CwHypCollateral,
+  TokenStandard.CosmNativeHypCollateral,
+];
```
paulbalaji, line 29: **Subtle: `CosmNativeHypCollateral` diverges from SDK balance resolution**

This uses `collateralAddressOrDenom` as the bank denom directly. The SDK instead instantiates `CosmNativeHypCollateralAdapter({ token: addressOrDenom })` and resolves the live denom from the chain via `provider.getToken()`. If the config value ever drifts from the on-chain denom, this batch path returns a different result than the SDK fallback.

In practice this works when warp route configs are correct (and the code already falls back to `{ type: 'unknown' }` when `collateralAddressOrDenom` is absent), so this is low risk. Noting for awareness — if you see phantom zero balances on CosmNativeHypCollateral tokens in the future, this is the place to look.

src/features/balances/hooks.ts
```diff
@@ -169,13 +181,28 @@ export function useTokenBalances(tokens: Token[], scope: string, addressOverride
         }
       }
 
+      // Cosmos — per-chain bech32 address lookup
+      if (cosmosAddresses.length > 0) {
```
paulbalaji: **Gap: `addressOverride` is silently ignored for Cosmos**

When a user pastes a Cosmos recipient address, `getAddressProtocolType` detects it and stores it in `effectiveAddresses`. But this Cosmos block reads exclusively from `cosmosAddresses` (connected wallet). The pasted address is never used.

Result: destination-mode token lists show the connected wallet's Cosmos balances, not the pasted recipient's. EVM and Sealevel handle `addressOverride` correctly.

Cosmos bech32 addresses are per-chain, so supporting this properly is non-trivial. At minimum, a comment here documenting the intentional gap would prevent future confusion:
```ts
// Note: addressOverride (pasted recipient) is not applied to Cosmos lookups.
// Cosmos addresses are chain-specific bech32 — resolving chain from a pasted
// address requires prefix matching, which is not yet implemented.
```

paulbalaji (dismissed):
## Consolidated Review

Core Cosmos batch-balance pattern is clean and consistent with the EVM/Sealevel approach. Three issues below — one correctness bug, one subtle SDK divergence, and one behavioral gap. Inline comments on each.

### Resolved threads (agree with resolutions)
- StargateClient connection leak — fixed in c88e14f6
- Zero-balance tokens omitted — consistent with EVM/Sealevel; no consumer relies on explicit `0n` keys

### Not covered here
- CodeRabbit's two open threads on `SKILL.md` (unrelated drive-by file) are doc-only nits

src/features/balances/cosmos.ts
Xaroz: Fixed in aa129cfc — moved CwHypSynthetic to BANK_DENOM_FROM_COLLATERAL.

src/features/balances/cosmos.ts
Xaroz: Intentional tradeoff — using config values avoids extra RPC calls. If collateralAddressOrDenom is wrong in config, it would also break EVM/Sealevel balance fetching, not just Cosmos. Falls back to SDK when absent.

src/features/balances/hooks.ts
Xaroz: Fixed in aa129cfc — cosmos addressOverride now matches bech32 prefix against chain metadata, same pattern as WarpCore validation.

.claude/skills/resolve-pr-reviews/SKILL.md
```diff
@@ -0,0 +1,106 @@
...
+gh api repos/{owner}/{repo}/pulls/{pr}/comments --jq '.[] | select(.position != null or .line != null)'
+
+# Get issue-level comments (general PR comments from reviewers, not bots)
+gh api repos/{owner}/{repo}/issues/{pr}/comments
+```
+
+Filter out:
+- Deployment/CI bots (Vercel deploy previews, CI status checks)
+- Already-resolved threads
+- Your own previous replies
```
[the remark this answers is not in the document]
Xaroz: Skipping — the endpoint works, we've used it successfully in this PR to reply to other comments.

.claude/skills/resolve-pr-reviews/SKILL.md
```diff
@@ -0,0 +1,106 @@
...
+**IMPORTANT: You MUST ask the user this question BEFORE showing any analysis. Do NOT skip this step or present comments prematurely.**
+
+After fetching and analyzing comments internally, use the `AskUserQuestion` tool to present a selection prompt:
+
+```
+Question: "Found N review comments on PR #XXXX. How would you like to go through them?"
+Options:
+  - "One by one" — Present each comment individually, user decides fix/skip before next
+  - "All at once" — Present all comments together, user reviews full list
+```
```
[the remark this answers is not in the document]
Xaroz: Skipping — this is a skill instruction file (pseudocode), not linted.

PR #1045, opened 2026-04-13
feat: add USDS, WETH, USDT igra yield warp routes to Nexus whitelist

troykessler:
Adds the following warp routes to the Nexus UI whitelist:

| Field | Value |
| ----- | ----- |
| **Linear** | [AW-548](https://linear.app/hyperlane-xyz/issue/AW-548/igra-usds-owner-yield-susds-sky-protocol) · [AW-551](https://linear.app/hyperlane-xyz/issue/AW-551/igra-owner-yield-morpho-gauntlet-weth) · [AW-545](https://linear.app/hyperlane-xyz/issue/AW-545/igra-morpho-steakhouse-usdt-yield-route) |

- `USDS/ethereum-igra`
- `WETH/ethereum-igra`
- `USDT/ethereum-igra`

### DIFF
--- a/src/consts/warpRouteWhitelist.ts
+++ b/src/consts/warpRouteWhitelist.ts
@@ -47,10 +47,14 @@ export const warpRouteWhitelist: Array<string> | null = [
   'USDC/paradex',
   'USDC/igra',
 
+  // USDS routes
+  'USDS/ethereum-igra',
+
   // USDT routes
   'USDT/eclipsemainnet-ethereum-solanamainnet',
   'USDT/ethereum-hyperevm',
   'USDT/ethereum-inevm',
+  'USDT/ethereum-igra',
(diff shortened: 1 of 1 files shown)


src/consts/warpRouteWhitelist.ts
paulbalaji, line 1: ah the comment i left on the registry PRs was to make these be just `/igra` like the other routes

Image

oh well it's just a name, but would be nice to be consistent and stick to the new stable warp id convention and not the old chain1-chain2 convention

paulbalaji (approved):

src/consts/warpRouteWhitelist.ts
troykessler: I see, I first thought we only want stable route ids for multicollateral ones, should I change the warp ids of the three igra yield routes?

PR #1058, opened 2026-04-16
fix: only open tronlinks wallet

troykessler:
Fixed the cycling through different wallets and only open TronLink


Xaroz:
thanks! one small suggestion is that maybe we can add more providers and see if the user can actually choose between all of them

Xaroz (approved):

PR #1064, opened 2026-04-17
test: make e2e tests resilient across production branches

Xaroz:
## Summary

Several e2e tests hardcoded main-branch values (app name, default tokens, tip-card copy, specific chain names, aleo route IDs) and failed when run against production branches like `nexus`. This PR refactors them to derive values from app constants/registry metadata so they work across branches.

## Test changes

- **`tests/page-load/transfer-form-visible.spec.ts`** — uses `APP_NAME` for page title; uses `getOriginTokenButton`/`getDestinationTokenButton` testid helpers; resolves chain displayName via registry so `toContainText` matches what the UI actually renders (not the internal slug)
- **`tests/page-load/tip-card.spec.ts`** — skips when `!config.showTipBox`; asserts the tip-card container by testid + `Hide tip` button (heading copy is customized per branch)
- **`tests/embed/routes-param.spec.ts`** — new `resolveTestRoutes()` helper picks `warpRouteWhitelist[0]`/`[1]` on prod branches, falls back to `USDC/aleo`/`ETH/aleo` on main, skips on empty whitelist
- **`tests/chain-selection/filter-by-type.spec.ts`** — resolves mainnet chain displayName via registry; scopes assertions to `.token-picker-chain-row`; verifies filter behavior via the `Clear` button visibility (confirms the filter click actually took effect, not a silent no-op) + row-count delta
- **`tests/page-load/default-tokens.spec.ts`**, **`query-param-override.spec.ts`** — same slug→displayName fix applied (had the same latent bug)

## Helpers

- **`tests/helpers/constants.ts`** (new)
  - `resolveTestRoutes()` — maps `warpRouteWhitelist` to `(primary, secondary, skip)` for embed route tests
  - `resolveChainDisplayName(slug)` — mirrors `getChainDisplayName` logic but loads `chainMetadata` via dynamic import to sidestep Playwright's CJS runner choking on the registry barrel's transitive top-level await (from `@provablehq/wasm`)
- **`tests/helpers/locators.ts`** — adds `getTipCard(page)` alongside existing token-button helpers

## Component change

- **`src/components/tip/TipCard.tsx`** — adds `data-testid="tip-card"` so tests can target the tip card by testid instead of a class selector. Forked TipCards on prod branches (nexus etc.) will need the same one-line attribute on their next main merge.

## Verification

All 9 previously-failing chromium tests (+ the 4 newly-fixed default-tokens/query-param-override cases) pass on both `main` and `nexus`.


Xaroz:
@claude review my friend

tests/helpers/constants.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (warpRouteWhitelist === null) {
+    return { primary: 'USDC/aleo', secondary: 'ETH/aleo', skip: false };
+  }
+  if (warpRouteWhitelist.length === 0) {
+    return { primary: '', secondary: '', skip: true };
+  }
+  return {
+    primary: warpRouteWhitelist[0],
+    secondary: warpRouteWhitelist[1] ?? warpRouteWhitelist[0],
+    skip: false,
```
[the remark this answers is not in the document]
Xaroz: nah this is fine

tests/chain-selection/filter-by-type.spec.ts
```diff
@@ -1,15 +1,29 @@
 import { test, expect } from '@playwright/test';
+import { config } from '../../src/consts/config';
 import { getOriginTokenButton } from '../helpers/locators';
 
+// Default origin is assumed mainnet — used to verify the Testnet filter hides it.
+// Chain rows in the ChainFilterPanel use the .token-picker-chain-row class.
+const defaultOriginChain = config.defaultOriginToken?.split('-')[0];
```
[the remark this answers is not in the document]
Xaroz: chainame has no hyphen

tests/helpers/constants.ts
```diff
@@ -0,0 +1,17 @@
+import { warpRouteWhitelist } from '../../src/consts/warpRouteWhitelist';
+
+// Picks known-good warp route IDs for embed ?routes= tests.
+// Prefers whitelist entries on prod branches; falls back to registry routes on main.
+export function resolveTestRoutes(): { primary: string; secondary: string; skip: boolean } {
+  if (warpRouteWhitelist === null) {
+    return { primary: 'USDC/aleo', secondary: 'ETH/aleo', skip: false };
+  }
```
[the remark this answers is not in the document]
Xaroz: we are fine with this for now

tests/chain-selection/filter-by-type.spec.ts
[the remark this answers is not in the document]
Xaroz: Skipping — @Xaroz confirmed chain names have no hyphens.

tests/helpers/constants.ts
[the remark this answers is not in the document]
Xaroz: Skipping — acceptable duplicate-route behavior for the multi-route test on single-entry whitelists.

tests/helpers/constants.ts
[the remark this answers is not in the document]
Xaroz: Skipping for now — we'll revisit if the aleo routes ever leave the canonical registry.

tests/chain-selection/filter-by-type.spec.ts
```diff
@@ -1,15 +1,29 @@
...
     await page.goto('http://localhost:3000');
     await page.getByText('Send').first().waitFor({ state: 'visible' });
 
     // Open token selector
     await getOriginTokenButton(page).click();
     await expect(page.getByText('Select Token')).toBeVisible();
 
+    const defaultChainRow = page
+      .locator('.token-picker-chain-row')
+      .filter({ hasText: new RegExp(defaultOriginChain!, 'i') });
```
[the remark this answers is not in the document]
Xaroz: Fixed in 2843228f — escaped and anchored the regex to ```^${name}\b```.

tests/page-load/transfer-form-visible.spec.ts
```diff
@@ -16,7 +18,12 @@ test.describe('Page Load - Transfer Form', () => {
     await expect(page.getByRole('button', { name: 'Connect wallet' }).first()).toBeVisible();
 
     // Send section: default origin token
-    await expect(page.getByRole('button', { name: 'ethereum USDC Ethereum' })).toBeVisible();
+    if (config.defaultOriginToken) {
+      const [originChain, originSymbol] = config.defaultOriginToken.split('-');
+      const originButton = page.getByRole('button', { name: new RegExp(originSymbol, 'i') }).first();
+      await expect(originButton).toBeVisible();
+      await expect(originButton).toContainText(originChain, { ignoreCase: true });
+    }
```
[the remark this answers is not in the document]
Xaroz: Escape + word-boundary anchor applied in 2843228f. Dropped `test.skip` for undefined defaults — the app falls back to featuredTokens / first routable token, so the test still has coverage via the generic page assertions.

tests/page-load/transfer-form-visible.spec.ts
```diff
@@ -32,7 +39,12 @@ test.describe('Page Load - Transfer Form', () => {
     await expect(page.getByText('Balance: 0.00', { exact: true })).toBeVisible();
 
     // Receive section: default destination token
-    await expect(page.getByRole('button', { name: 'base USDC Base' })).toBeVisible();
+    if (config.defaultDestinationToken) {
+      const [destChain, destSymbol] = config.defaultDestinationToken.split('-');
+      const destButton = page.getByRole('button', { name: new RegExp(destSymbol, 'i') }).last();
+      await expect(destButton).toBeVisible();
+      await expect(destButton).toContainText(destChain, { ignoreCase: true });
+    }
```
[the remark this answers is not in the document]
Xaroz: Fixed in 2843228f — replaced `.first()`/`.last()` with the existing `getOriginTokenButton`/`getDestinationTokenButton` testid helpers.

tests/page-load/tip-card.spec.ts
```diff
@@ -1,19 +1,21 @@
...
     await page.goto('http://localhost:3000');
     await page.getByText('Send').first().waitFor({ state: 'visible' });
 
-    // Tip card visible
-    const heading = page.getByRole('heading', { name: 'Bridge Tokens with Hyperlane Warp Routes!' });
-    await expect(heading).toBeVisible();
-    await expect(page.getByRole('link', { name: 'More' })).toBeVisible();
+    // Tip card container visible — content/copy varies per branch, so assert structure only
+    const tipCard = page.locator('.tip-card').first();
+    await expect(tipCard).toBeVisible();
```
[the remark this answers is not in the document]
Xaroz: Fixed in 2843228f — added `data-testid="tip-card"` to the TipCard component and a `getTipCard` helper, so the locator no longer needs `.first()`. Note: prod branches with a forked TipCard (e.g. nexus) will need the same one-line attribute.

Xaroz:
@claude one more time

tests/chain-selection/filter-by-type.spec.ts
```diff
@@ -1,15 +1,35 @@
 import { test, expect } from '@playwright/test';
+import { config } from '../../src/consts/config';
 import { getOriginTokenButton } from '../helpers/locators';
 
+// Default origin is assumed mainnet — used to verify the Testnet filter hides it.
+// Chain rows in the ChainFilterPanel use the .token-picker-chain-row class.
+const defaultOriginChain = config.defaultOriginToken?.split('-')[0];
```
paulbalaji, line 7: **Blocker — slug vs displayName mismatch.** `defaultOriginToken.split('-')[0]` yields the internal chain *name* slug (e.g. `solanamainnet`, `eclipsemainnet`, `worldchain`, `bsc`), but chain rows are rendered with `chain.displayName` (`src/features/chains/ChainList.tsx:65` — "Solana", "Eclipse", "World Chain", "BNB Smart Chain"). The anchored regex `^<slug>\b` matches `ethereum`→"Ethereum" by coincidence, but on any prod branch whose default uses a divergent-displayName chain this test's baseline `expect(defaultChainRow.first()).toBeVisible()` (line 31) and the Mainnet-filter positive assertion (line 62–67) will false-fail on a perfectly healthy app.

Resolve the `displayName` via the registry/multiProvider at test time (similar to `getChainDisplayName` in `src/features/chains/utils.ts`), or match the token *symbol* within `.token-picker-token-row` instead of matching the chain row by slug.

tests/chain-selection/filter-by-type.spec.ts
```diff
@@ -21,14 +41,13 @@ test.describe('Chain Selection - Filter by Type', () => {
     // Click Testnet filter
     await page.getByRole('button', { name: 'Testnet', exact: true }).click();
 
-    // Chain list should only show testnet chains (e.g., Sepolia, Alfajores)
-    await expect(page.getByRole('button', { name: /Sepolia/i }).first()).toBeVisible();
-
-    // Mainnet chains should not be visible
-    await expect(page.getByRole('button', { name: 'ethereum Ethereum', exact: true })).not.toBeVisible();
+    // Mainnet chain (default origin) row should no longer be visible — filter works
+    await expect(defaultChainRow).toHaveCount(0);
```
paulbalaji: **Assertion is tautologically passable.** `toHaveCount(0)` on the default-origin row succeeds whenever the filter click renders *nothing* — e.g. a broken filter that empties the list, a handler that throws and unmounts the panel, or a stale locator. Consider adding a positive check that proves Testnet mode is actually active:

```ts
// either: assert the list still has rows (just not mainnet ones), e.g.
//   await expect(page.locator('.token-picker-chain-row')).not.toHaveCount(0);
// or: assert the filter chip's selected state
//   await expect(page.getByRole('button', { name: 'Testnet', exact: true })).toHaveAttribute('aria-pressed', 'true');
```

tests/chain-selection/filter-by-type.spec.ts
```diff
@@ -39,8 +58,12 @@ test.describe('Chain Selection - Filter by Type', () => {
...
-    // Should show mainnet chains
-    await expect(page.getByRole('button', { name: 'ethereum Ethereum', exact: true })).toBeVisible();
-    await expect(page.getByRole('button', { name: 'arbitrum Arbitrum', exact: true })).toBeVisible();
+    // Default origin chain row (mainnet) remains visible after Mainnet filter
+    await expect(
+      page
+        .locator('.token-picker-chain-row')
+        .filter({ hasText: chainRowRegex! })
+        .first(),
+    ).toBeVisible();
```
paulbalaji, line 81: **No before/after delta.** The default-mainnet row is already visible *before* the Mainnet filter is clicked (mainnet is the default view), so this passes even if the Mainnet click is a silent no-op. To prove the filter actually did something, capture the full chain-row count pre-click and assert it changes (or assert the filter chip's selected state, or that all visible rows are mainnet).

tests/page-load/transfer-form-visible.spec.ts
```diff
@@ -15,8 +18,15 @@ test.describe('Page Load - Transfer Form', () => {
...
-    // Send section: default origin token
-    await expect(page.getByRole('button', { name: 'ethereum USDC Ethereum' })).toBeVisible();
+    // Send section: default origin token (only assert when configured; otherwise the app
+    // falls back to featuredTokens / first routable token — covered elsewhere)
+    const originButton = getOriginTokenButton(page);
+    await expect(originButton).toBeVisible();
+    if (config.defaultOriginToken) {
+      const [originChain, originSymbol] = config.defaultOriginToken.split('-');
+      await expect(originButton).toContainText(originSymbol);
+      await expect(originButton).toContainText(originChain, { ignoreCase: true });
```
paulbalaji: **Blocker — same slug vs displayName issue here.** The origin button renders `{symbol}` + `getChainDisplayName(token.chainName)` (`src/features/tokens/TokenSelectField.tsx:187,202`), so `toContainText(originChain)` with `originChain = 'solanamainnet'` will not find `'solanamainnet'` in a button labeled `"USDC Solana"`. Works today by coincidence because `main`/`nexus` both use `ethereum-USDC` where slug == displayName.

This also applies to the destination block at line 48–50. Resolve the displayName at runtime, or drop the `chain` portion of the assertion and just assert the symbol — the chain is already verified by `query-param-override.spec.ts` / `default-tokens.spec.ts`.

tests/page-load/transfer-form-visible.spec.ts
```diff
@@ -1,12 +1,15 @@
...
+import { getDestinationTokenButton, getOriginTokenButton } from '../helpers/locators';
 
 test.describe('Page Load - Transfer Form', () => {
   test('should display the transfer form on page load', async ({ page }) => {
     await page.goto('http://localhost:3000');
     await page.getByText('Send').first().waitFor({ state: 'visible' });
 
     // Page title
-    await expect(page).toHaveTitle('Hyperlane Warp UI Template');
+    await expect(page).toHaveTitle(APP_NAME);
```
paulbalaji, line 12: **Tautological check.** `APP_NAME` is imported from `src/consts/app.ts` and is the exact value rendered into `<title>` in `src/components/layout/AppLayout.tsx` and `src/pages/embed.tsx`. Both sides of the assertion derive from the same constant — this can never fail, so it doesn't actually verify the `<title>` was rendered. Assert a stable substring (e.g. `/Hyperlane/i` or `/Warp/i`) instead, or simply assert the title is non-empty.

paulbalaji (commented):
## Review Summary

The follow-up commit `2843228f` addresses most of the prior reviews — regex escaping/anchoring, testid helpers, `data-testid="tip-card"`, feature-flag skip guards are all in good shape. Nice work.

Not approving yet: two findings below defeat the PR's own cross-branch-resilience claim on any prod branch whose default chain slug diverges from its `displayName` (e.g. `solanamainnet`→"Solana", `bsc`→"BNB Smart Chain", `eclipsemainnet`→"Eclipse", `worldchain`→"World Chain"). It passes on `main` and `nexus` by coincidence — both happen to use `ethereum-USDC` where slug == displayName.

### Highlights

1. **Blocker — slug vs displayName mismatch (×2)** `config.defaultOriginToken.split('-')[0]` yields the internal chain *name* (`solanamainnet`), but the rendered chain row uses `chain.displayName` (`src/features/chains/ChainList.tsx:65`) and the token button uses `getChainDisplayName(...)` (`src/features/tokens/TokenSelectField.tsx:187,202`). Inline comments on both files.
2. **Loose assertions that can't catch a broken filter** — Testnet filter only asserts the default row disappears (passes if the list renders empty for any reason); Mainnet filter has no before/after delta (passes if the click is a no-op). Inline comments below.
3. **Tautological title check** — `toHaveTitle(APP_NAME)` where both sides derive from the same exported constant. Inline comment below.

### Already addressed (not re-raising)

`escapeRegex` + `^…\b` anchoring, `.first()/.last()` → testid helpers, `data-testid="tip-card"`, feature-flag skip guards, USDC/aleo fallback (author confirmed acceptable), single-entry whitelist duplicate route (author confirmed acceptable), CodeRabbit's `\b`/hyphen nit (author confirmed chain names have no hyphens).

### Observations Outside This PR

- `tests/chain-selection/filter-by-protocol.spec.ts:23,57` and `tests/chain-selection/edit-chain.spec.ts:20,48` still hardcode `'ethereum Ethereum'` — the exact pattern this PR removes elsewhere. Either mark as out of scope in the PR description or extend the migration, otherwise the cross-branch goal remains partial.
- `tests/helpers/constants.ts` exports only `resolveTestRoutes()` — the filename is misleading; `routes.ts` or folding into `locators.ts`/`testRoutes.ts` would match the existing convention. Nit.
- Implicit cross-branch contract isn't documented anywhere. For these tests to survive on forks, a prod branch must preserve: `defaultOriginToken` format `chain-SYMBOL` with a mainnet chain, `showTipBox` flag, `warpRouteWhitelist` shape, `data-testid="tip-card"`, the `.token-picker-chain-row` class, `APP_NAME` as `<title>`, and the `"Hide tip"` button title. A short block comment at the top of `tests/helpers/constants.ts` or a `tests/README.md` would make this self-documenting.
- `.tip-card` and `.token-picker-chain-row` are CSS classes doing double duty as test hooks. A reskin/restyle breaks tests silently. Prefer `data-testid` consistently going forward.
- `tests/page-load/tip-card.spec.ts:17` still depends on the hardcoded `"Hide tip"` button title — a retranslated/rebranded close button breaks it on forks. Consider `data-testid="tip-card-close"` on the button in `src/components/tip/TipCard.tsx`.

tests/chain-selection/filter-by-type.spec.ts
```diff
@@ -1,15 +1,35 @@
...
 
+// Default origin is assumed mainnet — used to verify the Testnet filter hides it.
+// Chain rows in the ChainFilterPanel use the .token-picker-chain-row class.
+const defaultOriginChain = config.defaultOriginToken?.split('-')[0];
+
+const escapeRegex = (s: string) => s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+// Anchor to start and follow with word boundary so short names (e.g. "eth") don't match "ethereum"
+const chainRowRegex = defaultOriginChain
+  ? new RegExp(`^${escapeRegex(defaultOriginChain)}\\b`, 'i')
+  : null;
```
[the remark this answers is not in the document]
Xaroz: Skipping — the whole regex approach is being replaced (see paulbalaji's thread below) with displayName matching via a new `resolveChainDisplayName` helper, so the `\b` edge case is moot.

tests/chain-selection/filter-by-type.spec.ts
Xaroz: @paulbalaji fixed in ff4c8da4. Added a `resolveChainDisplayName` test helper that reads `chainMetadata` from the registry (via dynamic import to avoid Playwright's CJS-vs-ESM-TLA issue with the registry barrel) and mirrors `getChainDisplayName` logic. Chain rows now match by displayName instead of slug.

tests/chain-selection/filter-by-type.spec.ts
Xaroz: @paulbalaji fixed in ff4c8da4 — added `await expect(page.getByRole('button', { name: 'Clear' })).toBeVisible();` after the Testnet click. The Clear button only appears when a filter is active, so it confirms the click actually took effect before the `toHaveCount(0)` assertion.

tests/chain-selection/filter-by-type.spec.ts
Xaroz: @paulbalaji fixed in ff4c8da4 — same Clear-button positive check added after the Mainnet click to prove it wasn't a silent no-op.

tests/page-load/transfer-form-visible.spec.ts
Xaroz: @paulbalaji fixed in ff4c8da4 — uses `resolveChainDisplayName` for both origin and destination buttons. Also applied the same fix to `default-tokens.spec.ts` and `query-param-override.spec.ts` which had the same latent bug.

tests/page-load/transfer-form-visible.spec.ts
Xaroz: @paulbalaji skipping — the intent is to verify the `<title>` renders whatever `APP_NAME` resolves to, so self-referential is the desired behavior. Test still catches missing/broken `<title>` rendering; only blind spot is a bad `APP_NAME` value itself, which would surface in review. Any hardcoded substring (`/Hyperlane/i`) would re-introduce the cross-branch fragility this PR removes.

tests/chain-selection/filter-by-type.spec.ts
```diff
@@ -1,15 +1,31 @@
...
 import { getOriginTokenButton } from '../helpers/locators';
 
+// Default origin is assumed mainnet — used to verify the Testnet filter hides it.
+// Match chain rows by displayName (what the UI renders), not by the internal slug.
+const defaultOriginChain = config.defaultOriginToken?.split('-')[0];
+
 test.describe('Chain Selection - Filter by Type', () => {
   test('should filter chains by Testnet type', async ({ page }) => {
+    test.skip(!defaultOriginChain, 'No defaultOriginToken configured');
+    const defaultChainDisplayName = await resolveChainDisplayName(defaultOriginChain!);
```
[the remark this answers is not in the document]
Xaroz: Fixed in f0bbac91 — added `isTestnetChain` helper that reads `chainMetadata[slug].isTestnet` from the registry, and both Testnet/Mainnet tests now `test.skip` when the default origin is a testnet chain.

tests/helpers/constants.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+// Mirrors `getChainDisplayName` in src/features/chains/utils.ts. Can't import
+// that function directly because its file transitively pulls in
+// @hyperlane-xyz/registry, whose barrel includes aleo/@provablehq/wasm TLA
+// and breaks Playwright's CJS test loader.
+export async function resolveChainDisplayName(slug: string): Promise<string> {
+  const { chainMetadata } = await import('@hyperlane-xyz/registry');
+  const metadata = chainMetadata[slug];
+  return metadata?.displayName || toTitleCase(slug);
```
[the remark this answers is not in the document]
Xaroz: Fixed in f0bbac91 — fallback now `metadata?.displayName || metadata?.name || slug`, matching `chain.displayName || chain.name` from `src/features/chains/hooks.ts:52`.

tests/chain-selection/filter-by-type.spec.ts
```diff
@@ -1,15 +1,35 @@
...
     await page.goto('http://localhost:3000');
     await page.getByText('Send').first().waitFor({ state: 'visible' });
 
     // Open token selector
     await getOriginTokenButton(page).click();
     await expect(page.getByText('Select Token')).toBeVisible();
 
+    const defaultChainRow = page
+      .locator('.token-picker-chain-row')
+      .filter({ hasText: defaultChainDisplayName });
```
paulbalaji: **MEDIUM — substring collision in `hasText`.**

`.filter({ hasText: defaultChainDisplayName })` uses substring match. The registry exposes pairs like `Base`/`Base Sepolia`, `Arbitrum`/`Arbitrum Sepolia`, `Optimism`/`Optimism Sepolia`, `Polygon`/`Polygon Amoy`, `Solana`/`Solana Devnet`, `Mantle`/`Mantle Sepolia`. After the Testnet filter at line 42, the "Base Sepolia" row is still visible and still substring-matches "Base", so `toHaveCount(0)` at line 48 false-fails on any prod branch whose default origin is `base-*`/`arbitrum-*`/`optimism-*`/etc.

Fix: match on exact displayName via regex `^\s*<escaped>\s*$`, or anchor rows on the `chain.name` slug (e.g. add `data-chain={chain.name}` to `ChainList.tsx:102` and target that). Same issue applies at line 76.

tests/chain-selection/filter-by-type.spec.ts
```diff
@@ -39,8 +66,15 @@ test.describe('Chain Selection - Filter by Type', () => {
     // Click Mainnet filter
     await page.getByRole('button', { name: 'Mainnet', exact: true }).click();
 
-    // Should show mainnet chains
-    await expect(page.getByRole('button', { name: 'ethereum Ethereum', exact: true })).toBeVisible();
-    await expect(page.getByRole('button', { name: 'arbitrum Arbitrum', exact: true })).toBeVisible();
+    // Clear button confirms the filter is active (not a silent no-op)
+    await expect(page.getByRole('button', { name: 'Clear' })).toBeVisible();
```
paulbalaji, line 76: **NIT/MEDIUM — Mainnet-filter test has no negative assertion.**

The test only verifies the mainnet default stays visible + a Clear button appears. But the default is mainnet, so the row is visible with *no* filter too; and `Clear` shows for any active filter (per `ChainFilterPanel.tsx:154`, gated on `isFilterActive`). A regression where the Mainnet click silently applies Testnet semantics would still pass. Add a negative check: pick a known-testnet displayName via `Object.values(chainMetadata).find(m => m.isTestnet)` and assert `toHaveCount(0)` after the Mainnet click.

tests/helpers/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+
+// Mirrors ChainList's rendered label (`chain.displayName || chain.name`, see
+// src/features/chains/hooks.ts:52). Can't import the source `getChainDisplayName`
+// directly because its file transitively pulls in @hyperlane-xyz/registry,
+// whose barrel includes aleo/@provablehq/wasm TLA and breaks Playwright's CJS
+// test loader.
+export async function resolveChainDisplayName(slug: string): Promise<string> {
+  const { chainMetadata } = await import('@hyperlane-xyz/registry');
+  const metadata = chainMetadata[slug];
+  return metadata?.displayName || metadata?.name || slug;
```
paulbalaji: **MEDIUM — helper reads published registry only; UI renders merged metadata.**

`resolveChainDisplayName` reads `chainMetadata` from `@hyperlane-xyz/registry` directly, but the UI renders labels from the assembled metadata in `src/features/chains/metadata.ts:72` (merges `chains.ts` + `chains.yaml` on top of the registry) and `metadata.ts:99` (layers store/localStorage overrides on top of that). Concrete failure: `origin/nambrot/crossstableswap-demo` overrides `arbitrum.displayName` to "Arbitrum One" while the published registry returns "Arbitrum" — every test that asserts via this helper will assert the wrong string on that branch.

Also diverges from `src/features/chains/utils.ts:22` which falls back to `toTitleCase(metadata.name)` (this helper uses raw `metadata.name`). Any chain lacking `displayName` silently fails the `toContainText` assertion (case-sensitive).

Fix: either mirror the merge here (import `chains.ts`/`chains.yaml`, apply `mergeChainMetadataMap`) or resolve the display name inside `page.evaluate` against the hydrated store after init.

tests/helpers/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+    secondary: warpRouteWhitelist[1] ?? warpRouteWhitelist[0],
+    skip: false,
+  };
+}
+
+// Mirrors ChainList's rendered label (`chain.displayName || chain.name`, see
+// src/features/chains/hooks.ts:52). Can't import the source `getChainDisplayName`
+// directly because its file transitively pulls in @hyperlane-xyz/registry,
+// whose barrel includes aleo/@provablehq/wasm TLA and breaks Playwright's CJS
+// test loader.
```
paulbalaji: **NIT — comment is inaccurate and justifies machinery that isn't needed.**

The comment says `@hyperlane-xyz/registry`'s barrel includes aleo `@provablehq/wasm` TLA. It doesn't — `node_modules/@hyperlane-xyz/registry/dist/index.js` re-exports pure JSON (`chainMetadata`, `warpRouteConfigs`, `addresses`) with no TLA and no `@provablehq` anywhere. The actual TLA source is wallet/store modules reachable via `src/features/chains/utils.ts` → `multiProvider`.

Since the registry barrel is TLA-free, consider dropping the dynamic import + `async` entirely and using a plain static `import { chainMetadata } from '@hyperlane-xyz/registry'` in both helpers — simpler callers, no Promises. And rephrase the comment to point at the real TLA source.

paulbalaji (commented):
## Re-review summary

Thanks for the follow-up commits — `ff4c8da4` and `f0bbac91` cleanly resolved the earlier threads (slug→displayName, Clear-button positive assertion, testnet-default skip, `displayName || name || slug` fallback). All existing inline threads are accounted for.

**Not approving yet**: four net-new issues still break the PR's own cross-branch resilience claim on real prod branches. Two of them land outside the diff, so I'm calling them out here instead of inline.

## Observations outside this PR (but load-bearing for the cross-branch claim)

- `tests/page-load/query-param-override.spec.ts:12-14, 19-20, 22-23, 63` — **HIGH.** The first and third tests still hardcode `base ETH Base` / `ethereum ETH Ethereum`. Defeats the cross-branch claim on prod branches whose whitelist excludes `ETH/base-ethereum`. Concrete example: `origin/trump` whitelists only `TRUMP/arbitrum-avalanche-base-flowmainnet-form-optimism-solanamainnet-worldchain`, so the `ETH` token on `base` or `ethereum` doesn't exist there. Derive a valid pair from the effective whitelist/registry (similar to `resolveTestRoutes`) or `test.skip` when the targeted route isn't available.
- `tests/embed/routes-param.spec.ts:32-34` — **LOW.** `page.waitForTimeout(10000)` hard-wait burns 10s every run and doesn't assert the real failure signal. Replace with `await expect(page.getByText('Send').first()).toBeHidden({ timeout: 10000 })` or assert a known error element.

tests/chain-selection/filter-by-type.spec.ts
Xaroz: @paulbalaji fixed in 4d194898. Added `data-chain={chain.name}` + `data-is-testnet` attributes to `ChainButton` in `ChainList.tsx` and switched the locator to `[data-chain="<slug>"]` — no more displayName substring matching, so `Base` vs `Base Sepolia` collision is impossible.

tests/chain-selection/filter-by-type.spec.ts
Xaroz: @paulbalaji fixed in 4d194898 — added negative assertions in both tests: `.token-picker-chain-row[data-is-testnet="false"]` count is 0 after Testnet filter, and `[data-is-testnet="true"]` count is 0 after Mainnet filter. Catches silent no-op regressions on either side.

tests/helpers/constants.ts
Xaroz: @paulbalaji fixed in 4d194898 — `resolveChainDisplayName` deleted entirely. Tests now match chain rows by `data-chain={chain.name}` slug (rendered via `useChainInfos`, so it always reflects merged metadata from `chains.ts`/`chains.yaml`/store overrides). No more published-registry vs hydrated-metadata divergence.

tests/helpers/constants.ts
Xaroz: @paulbalaji fixed in 4d194898 — the whole dynamic-import machinery is gone along with the helper. Thanks for catching the incorrect TLA attribution.

Xaroz:
@paulbalaji re: the two out-of-scope observations in your re-review:

- **`routes-param.spec.ts:32-34` (LOW)** — fixed in 4d194898: replaced `page.waitForTimeout(10000)` with `await expect(page.getByText('Send').first()).toBeHidden({ timeout: 10000 })`. No more hard 10s wait, and the assertion is on the real failure signal.
- **`query-param-override.spec.ts` hardcoded `base ETH Base` / `ethereum ETH Ethereum` (HIGH)** — skipping this one for now. A robust fix would discover a valid (chain, symbol) pair dynamically from the app's own default state (swap origin↔destination via query params and assert the swap landed), but it's a larger rewrite of those tests. Tracked separately.

paulbalaji (approved):
## Re-review at `4d194898` — approving

The `data-chain` / `data-is-testnet` refactor in `4d194898` is a clean solution — anchoring tests on attributes rendered from the same `useChainInfos`/`multiProvider` path the UI uses eliminates the displayName/substring/merged-metadata class of issues entirely, and the helper deletion removes ~20 lines of machinery that was trying (and failing) to mirror UI state at test time. Nice shape.

All four inline findings from the prior pass are addressed on this head: substring collision → `[data-chain="<slug>"]` anchor; Mainnet-filter negative assertion → `[data-is-testnet="true"]` count == 0; registry/merged-metadata divergence → helper deleted (attribute sourced from the live UI); inaccurate TLA comment → gone. The `waitForTimeout(10000)` → `toBeHidden({ timeout: 10000 })` swap in `routes-param.spec.ts` is strictly stronger than the prior `.not.toBeVisible()`. All existing inline threads across all reviewers are resolved.

## Follow-up, non-blocking

One HIGH observation from the prior pass is still standing on `tests/page-load/query-param-override.spec.ts`:

- Tests 1 (lines 9-26) and 3 (lines 52-64) still hardcode `?origin=base&originToken=ETH&destination=ethereum&destinationToken=ETH` and assert `getByRole('button', { name: 'base ETH Base' })` / `'ethereum ETH Ethereum'`. On prod branches whose whitelist excludes the `ETH/base-ethereum` route — e.g. `origin/trump` (whitelist `[TRUMP/arbitrum-avalanche-base-flowmainnet-form-optimism-solanamainnet-worldchain]`) — these assertions fail on a healthy app. This contradicts the PR's stated cross-branch goal, but the fallback-defaults test (test 2) was cleaned up to the new pattern, so the fix pattern is already demonstrated: derive the URL params from `config.defaultOriginToken`/`config.defaultDestinationToken` (or from `resolveTestRoutes()`) and use `toHaveAttribute('data-chain', ...)` + `toContainText(symbol)` for the assertion.

Happy to leave this as a follow-up — the rest of the PR is in good shape and this class of failure is already well-understood on affected branches. File a follow-up issue if you'd like me to pick it up.

PR #1070, opened 2026-04-22
feat: blend fluent

troykessler:
Adds fluent warp routes to the Nexus UI whitelist:

`BLEND/fluent`


paulbalaji (commented):
the diff seems larger than expected again

paulbalaji:
perhaps targetting the wrong branch

paulbalaji (approved):

PR #1083, opened 2026-04-28
fix: show and log gas estimation error

troykessler:
Before the gas estimation failed silently, now we log the error and show "Estimation failed" instead of "-" in the ui


Xaroz:
@claude review

Xaroz (approved):

src/features/transfer/FeeSectionButton.tsx
```diff
@@ -42,7 +44,13 @@ export function FeeSectionButton({
...
   const hasFees = fees !== null;
   const isClickable = hasFees && !isLoading;
-  const feeText = isLoading ? loadingText : hasFees ? fees.totalFees : '-';
+  const feeText = isLoading
+    ? loadingText
+    : isError
+      ? 'Estimation failed'
+      : hasFees
+        ? fees.totalFees
+        : '-';
```
Xaroz: nit: can use `useMemo` to return this text now that it is deeply nested

paulbalaji (approved):

PR #1086, opened 2026-05-01
feat: store-resolved chainAddresses with addresses.yaml/ts overrides

Xaroz:
## Summary

Mirrors the existing `chainMetadata` flow for chain contract addresses, with two related upsides:

1. **Single source of truth** — `chainAddresses` is loaded once at warp context init and exposed via the Zustand store. Consumers read it synchronously instead of awaiting `registry.getAddresses()`.
2. **Filesystem override pattern** — adds `src/consts/chainAddresses.yaml` and `chainAddresses.ts` parallel to `chains.yaml` / `chains.ts`. Useful for local dev / forks that want to extend the registry without patching the store.

While restructuring, also lazy-loads the published `chainAddresses` and `chainMetadata` constants from `@hyperlane-xyz/registry` so they're shed from the initial bundle.

### Address resolution mirrors metadata
- `src/features/chains/addresses.ts` — new helper `assembleChainAddresses(chainsInTokens, registry)`. Spreads `{...chainAddressesYaml, ...chainAddressesTS}`, validates with `ChainAddressesSchema`, fetches from registry (or falls back to the published constant), filters to chains referenced by warp routes, and merges filesystem on top of registry per key.
- `src/features/store.ts` — adds `chainAddresses: ChainMap<ChainAddresses>` to `WarpContext` / `AppState`, populated alongside `chainMetadata` via `Promise.all` in `initWarpContext`.
- `src/consts/chainAddresses.yaml` + `src/consts/chainAddresses.ts` — empty templates (commented examples).

### Existing consumers migrated to the store
- `src/features/balances/evm.ts` — `fetchChainBalances` and `getBatchAddress` now accept `ChainMap<ChainAddresses>` as a parameter.
- `src/features/balances/hooks.ts` — reads `chainAddresses` from the store and passes to `fetchChainBalances`.

### Bundle: lazy-load the published registry constants
- The published `chainAddresses` (~420 KB raw / ~80 KB gz) and `chainMetadata` (~384 KB raw / ~80 KB gz) constants from `@hyperlane-xyz/registry` are no longer statically imported. They're resolved via dynamic `import('@hyperlane-xyz/registry')` only on:
  - registry fetch failure in `assembleChainMetadata` / `assembleChainAddresses`
  - `listRegistryContent` failure in `initWarpContext` (PartialRegistry fallback path)
  - the no-`NEXT_PUBLIC_REGISTRY_URL` branch in `assembleChainMetadata` (rare path used when no custom registry is configured)
- Webpack/Turbopack emits one shared async chunk for the registry package; both lazy imports dedupe.

## Side effects worth flagging

- **Cold-start regression for deployments without `NEXT_PUBLIC_REGISTRY_URL`**: bounded by one chunk fetch on first load (~100–300ms), then cached. Production deployments that set the env var only hit the lazy paths on registry-failure (already an error path).
- API of `assembleChainMetadata` / `getBatchAddress` / `fetchChainBalances` changed (additive param). Internal only.

## Verification

- `pnpm format`, `pnpm lint`, `pnpm typecheck` clean.
- `pnpm test` — 22 files, 231 tests pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,35 @@
 {
   "name": "@hyperlane-xyz/warp-ui-template",
-  "description": "A web app template for building Hyperlane Warp Route UIs",
   "version": "13.0.0",
+  "private": true,
+  "description": "A web app template for building Hyperlane Warp Route UIs",
+  "homepage": "https://www.hyperlane.xyz",
+  "license": "Apache-2.0",
   "author": "J M Rossy",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hyperlane-xyz/hyperlane-warp-ui-template"
+  },
+  "main": "dist/src/index.js",
+  "types": "dist/src/index.d.ts",
+  "scripts": {
+    "clean": "rm -rf dist cache .next",
+    "predev": "node scripts/fetch-fonts.mjs",
+    "dev": "next dev",
+    "prebuild": "node scripts/fetch-fonts.mjs",
+    "fetch-fonts": "node --env-file-if-exists=.env.local scripts/fetch-fonts.mjs",
+    "build": "next build",
+    "typecheck": "tsc",
+    "lint": "oxlint ./src",
+    "start": "next start",
+    "test": "vitest --watch false",
+    "test:e2e": "playwright test --project=chromium",
+    "test:e2e:wallet": "playwright test tests/e2e-wallet --project=chromium",
+    "test:e2e:wallet:smoke": "playwright test tests/e2e-wallet/smoke/gate.spec.ts tests/e2e-wallet/autoconnect/evm.spec.ts tests/e2e-wallet/balance-display/evm.spec.ts tests/e2e-wallet/autoconnect/solana.spec.ts --project=chromium",
+    "format": "oxfmt ./src",
+    "link:monorepo": "node scripts/link-monorepo.js",
+    "unlink:monorepo": "node scripts/unlink-monorepo.js"
+  },
   "dependencies": {
     "@chakra-ui/next-js": "^2.4.2",
     "@chakra-ui/provider": "^2.4.2",
@@ -21,10 +48,10 @@
     "@emotion/react": "^11.13.3",
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^2.2.0",
-    "@hyperlane-xyz/registry": "24.3.0",
-    "@hyperlane-xyz/sdk": "33.0.2",
-    "@hyperlane-xyz/utils": "33.0.2",
-    "@hyperlane-xyz/widgets": "33.0.2",
+    "@hyperlane-xyz/registry": "25.0.0",
+    "@hyperlane-xyz/sdk": "33.1.0",
+    "@hyperlane-xyz/utils": "33.1.0",
+    "@hyperlane-xyz/widgets": "33.1.0",
     "@interchain-ui/react": "^1.23.28",
     "@intercom/messenger-js-sdk": "^0.0.18",
     "@metamask/post-message-stream": "6.1.2",
@@ -99,34 +126,7 @@
   "engines": {
     "node": ">=24"
   },
-  "homepage": "https://www.hyperlane.xyz",
-  "license": "Apache-2.0",
-  "main": "dist/src/index.js",
   "packageManager": "pnpm@10.25.0",
-  "private": true,
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/hyperlane-xyz/hyperlane-warp-ui-template"
-  },
-  "scripts": {
-    "clean": "rm -rf dist cache .next",
-    "predev": "node scripts/fetch-fonts.mjs",
-    "dev": "next dev",
-    "prebuild": "node scripts/fetch-fonts.mjs",
-    "fetch-fonts": "node --env-file-if-exists=.env.local scripts/fetch-fonts.mjs",
-    "build": "next build",
-    "typecheck": "tsc",
-    "lint": "oxlint ./src",
-    "start": "next start",
-    "test": "vitest --watch false",
-    "test:e2e": "playwright test --project=chromium",
-    "test:e2e:wallet": "playwright test tests/e2e-wallet --project=chromium",
-    "test:e2e:wallet:smoke": "playwright test tests/e2e-wallet/smoke/gate.spec.ts tests/e2e-wallet/autoconnect/evm.spec.ts tests/e2e-wallet/balance-display/evm.spec.ts tests/e2e-wallet/autoconnect/solana.spec.ts --project=chromium",
-    "format": "oxfmt ./src",
-    "link:monorepo": "node scripts/link-monorepo.js",
-    "unlink:monorepo": "node scripts/unlink-monorepo.js"
-  },
-  "types": "dist/src/index.d.ts",
   "pnpm": {
     "overrides": {
       "tronweb>ethers": "^6.13.5",
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "@emotion/react": "^11.13.3",
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^2.2.0",
-    "@hyperlane-xyz/registry": "25.0.0",
+    "@hyperlane-xyz/registry": "24.3.0",
     "@hyperlane-xyz/sdk": "33.1.0",
     "@hyperlane-xyz/utils": "33.1.0",
     "@hyperlane-xyz/widgets": "33.1.0",
--- a/src/consts/chainAddresses.yaml
+++ b/src/consts/chainAddresses.yaml
@@ -3,6 +3,11 @@
 # override registry entries per key, mirroring how chains.yaml extends chains.
 # Schema: any contract addresses you'd find in a registry chain's addresses.yaml
 # (e.g. mailbox, quotedCalls, validatorAnnounce, ...)
+# Full set of valid keys: https://github.com/hyperlane-xyz/hyperlane-registry/blob/main/chains/ethereum/addresses.yaml
+# (or the ChainAddresses type from @hyperlane-xyz/registry)
+#
+# Note: schema validation runs at app init — a typo here will throw and
+# block the whole warp context init, not just balance fetching.
 ---
 # Example using local anvil chain:
 # anvil1:
--- a/src/consts/chainAddresses.yaml
+++ b/src/consts/chainAddresses.yaml
@@ -1,0 +1,13 @@
+# A map of chain name -> contract addresses
+# Merges with addresses from the configured registry. Filesystem entries
+# override registry entries per key, mirroring how chains.yaml extends chains.
+# Schema: any contract addresses you'd find in a registry chain's addresses.yaml
+# (e.g. mailbox, quotedCalls, validatorAnnounce, ...)
+---
+# Example using local anvil chain:
+# anvil1:
+#   mailbox: '0x610178dA211FEF7D417bC0e6FeD39F05609AD788'
+#   quotedCalls: '0x9A9f2CCfdE556A7E9Ff0848998Aa4a0CFD8863AE'
+# anvil2:
+#   mailbox: '0x610178dA211FEF7D417bC0e6FeD39F05609AD788'
+#   quotedCalls: '0x9A9f2CCfdE556A7E9Ff0848998Aa4a0CFD8863AE'
--- a/src/features/balances/evm.ts
+++ b/src/features/balances/evm.ts
@@ -1,5 +1,5 @@
-import { chainAddresses } from '@hyperlane-xyz/registry';
-import { MultiProtocolProvider, Token } from '@hyperlane-xyz/sdk';
+import { ChainAddresses } from '@hyperlane-xyz/registry';
+import { ChainMap, MultiProtocolProvider, Token } from '@hyperlane-xyz/sdk';
 import { ProtocolType, normalizeAddress } from '@hyperlane-xyz/utils';
 import {
   Hex,
@@ -62,8 +62,8 @@ function classifyToken(token: Token): { type: TokenClassification; erc20Address?
  * Prefer the registry's batchContractAddress when available — some chains
  * (e.g. ancient8, viction) have the standard multicall3 address compromised.
  */
-function getBatchAddress(chainName: string): Hex {
-  const addresses = (chainAddresses as Record<string, Record<string, string>>)[chainName];
+function getBatchAddress(chainName: string, chainAddresses: ChainMap<ChainAddresses>): Hex {
+  const addresses = chainAddresses[chainName];
   if (addresses?.batchContractAddress) {
     return addresses.batchContractAddress.toLowerCase() as Hex;
   }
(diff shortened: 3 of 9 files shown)


Xaroz:
@claude review

src/features/store.ts
```diff
@@ -316,11 +322,10 @@ async function initWarpContext({
...
     const chainsInTokens = Array.from(new Set(coreConfig.tokens.map((t) => t.chainName)));
-    const { chainMetadata, chainMetadataWithOverrides } = await assembleChainMetadata(
-      chainsInTokens,
-      currentRegistry,
-      chainMetadataOverrides,
-    );
+    const [{ chainMetadata, chainMetadataWithOverrides }, chainAddresses] = await Promise.all([
+      assembleChainMetadata(chainsInTokens, currentRegistry, chainMetadataOverrides),
+      assembleChainAddresses(chainsInTokens, currentRegistry),
+    ]);
```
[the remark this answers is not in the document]
Xaroz: Fixed in dada6eb8 — both setters now propagate chainAddresses and chainMetadata.

src/features/chains/addresses.ts
```diff
@@ -0,0 +1,49 @@
...
+    registryChainAddresses = (await import('@hyperlane-xyz/registry')).chainAddresses;
+  }
+
+  // Filter to only chains referenced by the configured warp routes
+  registryChainAddresses = objFilter(registryChainAddresses, (c, _a): _a is ChainAddresses =>
+    chainsInTokens.includes(c),
+  );
+
+  // Filesystem entries override registry entries per key
+  return objMerge<ChainMap<ChainAddresses>>(registryChainAddresses, filesystemAddresses);
```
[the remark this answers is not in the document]
Xaroz: Verified objMerge's behavior is correct here: default depth 10, recurses into nested objects when both sides have objects at the same key. So filesystem `{chain1: {mailbox: 0x1}}` preserves registry's other keys for chain1. Skipping the doc tweak.

src/features/chains/addresses.ts
```diff
@@ -0,0 +1,49 @@
...
+    } catch {
+      logger.debug(
+        'Failed fetching chain addresses from GH registry, using published addresses',
+        config.registryUrl,
+      );
+      registryChainAddresses = (await import('@hyperlane-xyz/registry')).chainAddresses;
+    }
+  } else {
+    logger.debug('Using default published registry for chain addresses');
+    registryChainAddresses = (await import('@hyperlane-xyz/registry')).chainAddresses;
```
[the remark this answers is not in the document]
Xaroz: Fixed in dada6eb8 — collapsed the registry-fail and no-URL branches in both addresses.ts and metadata.ts so the dynamic import is called once via a single `if (!resolved)` check. Skipped the cross-file helper since each call site needs a different field shape; keeping it inline reads cleaner.

src/features/chains/addresses.ts
```diff
@@ -0,0 +1,49 @@
...
+  registry: IRegistry,
+): Promise<ChainMap<ChainAddresses>> {
+  const result = z.record(ChainAddressesSchema).safeParse({
+    ...ChainAddressesYaml,
+    ...ChainAddressesTS,
+  });
+  if (!result.success) {
+    logger.warn('Invalid chain addresses', result.error);
+    throw new Error(`Invalid chain addresses: ${result.error.toString()}`);
+  }
```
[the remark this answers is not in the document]
Xaroz: Fixed in dada6eb8 — added an inline note in chainAddresses.yaml flagging that a typo here blocks warp context init.

src/features/balances/evm.ts
```diff
@@ -224,6 +224,7 @@ export async function fetchChainBalances(
   group: ChainGroup,
   multiProvider: MultiProtocolProvider,
   evmAddress: Hex,
+  chainAddresses: ChainMap<ChainAddresses>,
```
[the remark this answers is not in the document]
Xaroz: Noted — leaving as-is for this PR to keep scope focused on the chainAddresses store/lazy-load. Will refactor to an options object as a follow-up.

src/consts/chainAddresses.yaml
```diff
@@ -0,0 +1,13 @@
...
+# Schema: any contract addresses you'd find in a registry chain's addresses.yaml
+# (e.g. mailbox, quotedCalls, validatorAnnounce, ...)
+---
+# Example using local anvil chain:
+# anvil1:
+#   mailbox: '0x610178dA211FEF7D417bC0e6FeD39F05609AD788'
+#   quotedCalls: '0x9A9f2CCfdE556A7E9Ff0848998Aa4a0CFD8863AE'
+# anvil2:
+#   mailbox: '0x610178dA211FEF7D417bC0e6FeD39F05609AD788'
+#   quotedCalls: '0x9A9f2CCfdE556A7E9Ff0848998Aa4a0CFD8863AE'
```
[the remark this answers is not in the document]
Xaroz: Fixed in dada6eb8 — added a link to the registry's ethereum/addresses.yaml plus a pointer to the ChainAddresses type.

Xaroz:
@claude review agen

Xaroz:
@claude — fixed the doc nit in bbe4165a (`addresses.yaml` → `chainAddresses.yaml` in `chainAddresses.ts:9`). Thanks for the round-2 pass.

src/features/store.ts
```diff
@@ -153,6 +159,8 @@ export const useStore = create<AppState>()(
         });
         set({
           chainMetadataOverrides: filtered,
+          chainMetadata,
+          chainAddresses,
```
paulbalaji, line 165: **Blocker — setters don't propagate the resolved `registry`.** The destructure here (and in `setWarpCoreConfigOverrides` at ~line 193) pulls `chainMetadata, chainAddresses, multiProvider, warpCore, …` from `initWarpContext()` but omits `registry`, and the `set({...})` payload likewise omits it.

If the previous `initWarpContext` run swapped `GithubRegistry` → `PartialRegistry` in its catch block (line ~314, when `listRegistryContent` fails), every subsequent setter call reuses the original `GithubRegistry` from `get().registry` and re-attempts the GitHub fetch — paying the network round-trip on every chain-override change, and creating a brief window where the in-store `registry` and the `chainAddresses` derived from a different registry are inconsistent.

`setWarpContext` (the rehydration path) writes `registry` correctly; only the two setters drop it. Fix: include `registry` in the destructure and `set({...})` of both setters, mirroring `setWarpContext`.

src/features/balances/hooks.ts
```diff
@@ -160,7 +162,9 @@ export function useTokenBalances(tokens: Token[], scope: string, addressOverride
       if (evmAddr) {
         const { chainGroups, fallbackTokens } = groupEvmTokensByChain(tokens, multiProvider);
         for (const [chainId, group] of chainGroups) {
-          promises.push(fetchChainBalances(chainId, group, multiProvider, evmAddr as Hex));
+          promises.push(
+            fetchChainBalances(chainId, group, multiProvider, evmAddr as Hex, chainAddresses),
```
paulbalaji, line 184: `fetchChainBalances` now reads `chainAddresses` from the store, but the queryKey at line 156 doesn't include it. `WarpContextInitGate` blocks first paint until `useReadyMultiProvider()` is truthy, so this isn't an initial-load race — but a runtime reinit (`setChainMetadataOverrides` / `setWarpCoreConfigOverrides`) updates `chainAddresses` while react-query keeps serving cached balances for up to `staleTime` (30s).

For the chains the PR's own comment in `evm.ts` flags as having a compromised canonical multicall3 (ancient8, viction), that means up to 30s of balance fetches against the wrong batch contract after the override flips.

Suggest including a stable digest of `chainAddresses` in the queryKey — at minimum the keys for the chains being fetched.

src/features/chains/addresses.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (config.registryUrl) {
+    try {
+      logger.debug('Using custom registry chain addresses from:', config.registryUrl);
+      registryChainAddresses = await registry.getAddresses();
+    } catch {
+      logger.debug(
+        'Failed fetching chain addresses from GH registry, using published addresses',
+        config.registryUrl,
+      );
+    }
```
paulbalaji, line 36: GH fetch failure is logged at `logger.debug` and the actual `error` is dropped. Operators with a misconfigured `NEXT_PUBLIC_REGISTRY_URL` will silently fall through to the published constants merged with the filesystem map — no visible signal. `metadata.ts` has the same shape; worth raising both to `logger.warn(..., error)` so the cause surfaces.

src/features/chains/addresses.ts
```diff
@@ -0,0 +1,50 @@
...
+    registryChainAddresses = (await import('@hyperlane-xyz/registry')).chainAddresses;
+  }
+
+  // Filter to only chains referenced by the configured warp routes
+  registryChainAddresses = objFilter(registryChainAddresses, (c, _a): _a is ChainAddresses =>
+    chainsInTokens.includes(c),
+  );
+
+  // Filesystem entries override registry entries per key
+  return objMerge<ChainMap<ChainAddresses>>(registryChainAddresses, filesystemAddresses);
```
paulbalaji, line 50: Note: `objFilter` only filters `registryChainAddresses` against `chainsInTokens`; filesystem entries for chains not referenced by any warp route still merge through. Probably benign for current consumers (everyone keys by chain name), but worth flagging the asymmetric semantics. If intentional, ignore. If not, filter `filesystemAddresses` against `chainsInTokens` too before the merge.

package.json
```diff
@@ -21,10 +48,10 @@
     "@emotion/react": "^11.13.3",
     "@emotion/styled": "^11.13.0",
     "@headlessui/react": "^2.2.0",
```
paulbalaji, line 50: Process nit: this PR mixes a cosmetic key reorder + `@hyperlane-xyz/registry` 24.3.0 → **25.0.0** (major bump) + sdk/utils/widgets minor bumps into the chainAddresses feature. The major-version bump in particular deserves its own commit/PR with the release-notes context.

Also worth confirming the resolution: `@hyperlane-xyz/registry` is hoisted to 25.0.0 here, but `tron-sdk` still pulls 24.3.0 transitively through `pnpm-lock.yaml` — intentional?

paulbalaji (commented):
## Review Summary

Mirrors the existing `chainMetadata` flow for `chainAddresses`, lazy-loads the published registry constants, and adds yaml/ts override files. Verified the lazy-load claim is clean — both dynamic-`import('@hyperlane-xyz/registry')` specifiers in `addresses.ts`, `metadata.ts`, and the `initWarpContext` catch block string-match exactly, so webpack dedupes them into a single async chunk; the static `IRegistry`/`GithubRegistry`/`PartialRegistry`/`ChainAddresses` imports left in `store.ts` tree-shake away from the heavy aggregates per registry's `sideEffects: false` plus next.config.js `optimizePackageImports`. `pnpm typecheck` / `lint` / `test` (231 tests, 22 files) all pass.

Accounted for the 6 prior `claude[bot]` inline threads — all resolved by the author in `dada6eb8` (and the doc-nit followup in `bbe4165a`); not re-raising.

One real bug worth fixing before merge (finding #1 below — registry not propagated by the setters); a medium cache-staleness window (#2); and a few small items.

**Recommendation:** request the registry-propagation fix; the rest are non-blocking. Not approving in this run pending #1.

- `src/features/store.ts:377` — the outer-catch return uses `registry` (the parameter) rather than `currentRegistry`. If `listRegistryContent` already swapped to `PartialRegistry` and a downstream step then threw, the swap is forgotten. Cosmetic — prefer `registry: currentRegistry`. The line itself isn't in the diff so leaving it here.
- `src/features/warpCore/warpCoreConfig.ts:3` still statically imports `warpRouteConfigs` from `@hyperlane-xyz/registry` (~1.18 MB raw) — out of scope for this PR but a candidate for the same lazy-load treatment in a follow-up.
- No tests added for `assembleChainAddresses`, `getBatchAddress`, or the new `fetchChainBalances` parameter signature. Mirrors the existing gap for `assembleChainMetadata` — not a blocker.

src/features/store.ts
Xaroz: Fixed in 872af528 — both setters now destructure and propagate `registry` from `initWarpContext`, mirroring `setWarpContext`.

src/features/balances/hooks.ts
Xaroz: Fixed in 872af528 — added a narrow digest of `batchContractAddress` per chain to the queryKey. Kept it narrow (just the field `fetchChainBalances` actually reads) since at 50–100 chains a full `JSON.stringify` would be ~50–100 KB. Comment notes the coupling so the digest gets widened if `fetchChainBalances` reads more fields.

src/features/chains/addresses.ts
Xaroz: Fixed in 872af528 — bound the error and bumped to `logger.warn` in both `addresses.ts` and `metadata.ts`.

src/features/chains/addresses.ts
Xaroz: Intentional — leaving the asymmetry. The codebase-wide convention for `*.yaml` consts (chains.yaml, warpRoutes.yaml) is that filesystem entries flow through unfiltered; tightening just chainAddresses.yaml would make it the lone outlier with stricter semantics.

package.json
Xaroz: Reverted the `@hyperlane-xyz/registry` bump in 872af528 — back to 24.3.0. SDK 33.1.0 has no direct dep on registry (only `@chain-registry/types`, a different package), so we keep the SDK/utils/widgets bumps (needed for the SDK fix) without the major-version registry bump. Lockfile now resolves to a single `registry@24.3.0` everywhere; tron-sdk's transitive matches. The 25.0.0 bump can land in its own PR once tron-sdk catches up.

paulbalaji (approved):
## Re-review of 872af528

All five items from the prior round are addressed:

1. **Registry propagation** ✅ — Both `setChainMetadataOverrides` and `setWarpCoreConfigOverrides` now destructure `registry` from `initWarpContext()` and include it in their `set({...})` payload (`src/features/store.ts:147,163,179,195`), matching `setWarpContext`. No remaining drop sites in the 140–206 range.
2. **queryKey digest** ✅ — `chainAddressesKey` (`src/features/balances/hooks.ts:154-161`) is a narrow per-chain digest of `batchContractAddress`, `.sort()`-stable, and added to the queryKey at line 173. The inline comment ("only reads batchContractAddress per chain. If that expands, widen this digest accordingly") flags the coupling for future maintainers.
3. **logger.warn with bound error** ✅ — Both `addresses.ts` and `metadata.ts` now bind `error` in the catch and call `logger.warn(..., error)` instead of `logger.debug(...)` with a dropped error. Fallback flow is unchanged.
4. **Filesystem filter asymmetry** — Author kept as-is; reasonable to match the existing `chains.yaml` / `warpRoutes.yaml` convention rather than make `chainAddresses.yaml` the lone outlier. Acceptable.
5. **Registry version revert** ✅ — `package.json` pinned back to `@hyperlane-xyz/registry@24.3.0`; sdk/utils/widgets bumps preserved; `pnpm-lock.yaml` resolves to a single 24.3.0 (no 25.0.0 anywhere); tron-sdk transitive matches.

Build verification on the latest head: `pnpm typecheck` clean, `pnpm lint` clean (2 pre-existing console warnings, unrelated), `pnpm test` 22 files / 231 tests pass.

LGTM.

PR #1092, opened 2026-05-08
feat: relayer api

troykessler:
Implements the relay api and calls it only if the message originates from a cctp v2 route. The new env variable NEXT_PUBLIC_RELAY_API_URL has been already added to vercel

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,6 @@
 NEXT_PUBLIC_WALLET_CONNECT_ID=12345678901234567890123456789012
 NEXT_PUBLIC_RPC_OVERRIDES='{"chain1":{"http":"https://..."}}'
+NEXT_PUBLIC_RELAY_API_URL=http://localhost:8900
 
 # Predicate Attestation Support (server-side only, optional)
 # Required for routes with PredicateRouterWrapper
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,7 @@
 NEXT_PUBLIC_WALLET_CONNECT_ID=12345678901234567890123456789012
 NEXT_PUBLIC_RPC_OVERRIDES='{"chain1":{"http":"https://..."}}'
+
+# Localhost only; production deployments must use https://
 NEXT_PUBLIC_RELAY_API_URL=http://localhost:8900
 
 # Predicate Attestation Support (server-side only, optional)
--- a/src/features/transfer/relayApi.ts
+++ b/src/features/transfer/relayApi.ts
@@ -1,5 +1,20 @@
+import { ProviderType, TypedTransactionReceipt } from '@hyperlane-xyz/sdk';
+import { ProtocolType } from '@hyperlane-xyz/utils';
+
 import { config } from '../../consts/config';
 import { logger } from '../../utils/logger';
+
+// keccak256("MessageSent(bytes)") — emitted by MessageTransmitter V2 for both native USDC
+// transfers (depositForBurn) and GMP-only messages. Covers CCTP V2 GMP that lacks DepositForBurn.
+const MESSAGE_SENT_TOPIC = '0x8c5261668696ce22758910d05bab8f186d6eb247ceac2af2e82c7dc17669b036';
+
+// Circle's MessageTransmitter V2 addresses (lowercase for comparison).
+// Source: https://developers.circle.com/cctp/references/contract-addresses#messagetransmitterv2
+const MESSAGE_TRANSMITTER_V2_ADDRESSES = new Set([
+  '0x81d40f21f12a8f0e3252bccb954d722d4c464b64', // mainnet (all chains except EDGE)
+  '0x5b61381fc9e58e70efc13a4a97516997019198ee', // mainnet EDGE
+  '0xe737e5cebeeba77efe34d4aa090756590b1ce275', // testnet (all chains)
+]);
 
 interface RelayResponse {
   messages: Array<{
@@ -11,11 +26,27 @@ interface RelayResponse {
 }
 
 /**
- * Submits an origin transaction hash to the relayer API for fast processing.
+ * Submits an origin transaction hash to the relayer API for fast CCTP processing.
  * Fire-and-forget: errors are logged but never surfaced to the user.
  */
-export async function submitToRelayApi(originChain: string, txHash: string): Promise<void> {
+export async function submitToRelayApi(
+  originChain: string,
+  txHash: string,
+  originProtocol: ProtocolType,
+  txReceipt: TypedTransactionReceipt | null | undefined,
+): Promise<void> {
   if (!config.relayApiUrl) return;
+
+  const isCctp =
+    originProtocol === ProtocolType.Ethereum &&
+    txReceipt != null &&
+    (txReceipt.type === ProviderType.EthersV5 || txReceipt.type === ProviderType.Viem) &&
+    txReceipt.receipt.logs.some(
+      (log) =>
+        log.topics[0] === MESSAGE_SENT_TOPIC &&
+        MESSAGE_TRANSMITTER_V2_ADDRESSES.has(log.address.toLowerCase()),
+    );
+  if (!isCctp) return;
 
   try {
     const payload = { origin_chain: originChain, tx_hash: txHash };
--- a/src/features/transfer/relayApi.ts
+++ b/src/features/transfer/relayApi.ts
@@ -1,5 +1,5 @@
 import { ProviderType, TypedTransactionReceipt } from '@hyperlane-xyz/sdk';
-import { ProtocolType } from '@hyperlane-xyz/utils';
+import { ProtocolType, isNullish } from '@hyperlane-xyz/utils';
 
 import { config } from '../../consts/config';
 import { logger } from '../../utils/logger';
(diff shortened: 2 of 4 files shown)


src/features/transfer/relayApi.ts
```diff
@@ -0,0 +1,74 @@
...
+      (log) =>
+        log.topics[0] === MESSAGE_SENT_TOPIC &&
+        MESSAGE_TRANSMITTER_V2_ADDRESSES.has(log.address.toLowerCase()),
+    );
+  if (!isCctp) return;
+
+  try {
+    const payload = { origin_chain: originChain, tx_hash: txHash };
+    logger.debug('[RelayAPI] Requesting relay', { url: `${config.relayApiUrl}/relay`, payload });
+    const response = await fetch(`${config.relayApiUrl}/relay`, {
```
paulbalaji: **[major]** Browser-side call likely won't reach the deployed relay as currently configured. The corresponding monorepo PR (#8710) ships CORS allowing only `https://nexus.hyperlane.xyz` and `http://localhost:3000`, and the review on that PR required Cloudflare Access in front of public DNS. This app's Vercel prod + preview origins aren't in the allowlist, and if Access is enabled the browser can't safely hold service-token credentials.

Two options:
- **Public endpoint:** request that the relay-infra side add this app's prod + preview origins to CORS, and confirm per-tenant WAF rate limits are in place (sidecar already has 100/min app-level).
- **Server-side proxy:** add a Next.js API route (e.g. `pages/api/relay.ts`) that forwards to the relay using a server-only env var (drop the `NEXT_PUBLIC_` prefix). That keeps the relay URL off the client and avoids CORS entirely.

Worth deciding before merging since the contract affects the env-var name (`NEXT_PUBLIC_RELAY_API_URL` vs server-only `RELAY_API_URL`) and the helper's call shape.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -269,6 +270,8 @@ async function executeTransfer({
       msgId,
     });
 
+    if (originTxHash) submitToRelayApi(origin, originTxHash, originProtocol, txReceipt);
```
paulbalaji, line 273: **[major]** Fire-and-forget without `.catch`. The helper's internal `try/catch` only wraps the fetch block — the `isCctp` log scan (`log.address.toLowerCase()`, `log.topics[0]` access) runs before the try and would surface as an unhandled promise rejection if anything in a malformed receipt throws synchronously. Low real-world probability, but every other helper called from `executeTransfer` is awaited; this is the only fire-and-forget and it lacks the standard guard.

Either:

```ts
void submitToRelayApi(origin, originTxHash, originProtocol, txReceipt).catch((e) =>
  logger.warn('[RelayAPI] submission failed', e),
);
```

…or wrap the entire body of `submitToRelayApi` in a single `try/catch` instead of just the fetch.

src/features/transfer/relayApi.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (!config.relayApiUrl) return;
+
+  const isCctp =
+    originProtocol === ProtocolType.Ethereum &&
+    txReceipt != null &&
+    (txReceipt.type === ProviderType.EthersV5 || txReceipt.type === ProviderType.Viem) &&
+    txReceipt.receipt.logs.some(
+      (log) =>
+        log.topics[0] === MESSAGE_SENT_TOPIC &&
+        MESSAGE_TRANSMITTER_V2_ADDRESSES.has(log.address.toLowerCase()),
```
paulbalaji: **[minor]** PR description says "calls it only if the message originates from a cctp v2 route", but this gate triggers on any tx whose receipt contains a `MessageSent(bytes)` log from a V2 MessageTransmitter — regardless of warp-route type. Today every transfer goes through `WarpCore`, so it's fine in practice. Flagging because the comment above ("Covers CCTP V2 GMP that lacks DepositForBurn") explicitly anticipates non-DepositForBurn paths, where the topic-only match could fire on flows the relayer can't process. If you want a tighter gate, check the warp `Token` / route type in addition to log topology.

src/features/transfer/relayApi.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (!isCctp) return;
+
+  try {
+    const payload = { origin_chain: originChain, tx_hash: txHash };
+    logger.debug('[RelayAPI] Requesting relay', { url: `${config.relayApiUrl}/relay`, payload });
+    const response = await fetch(`${config.relayApiUrl}/relay`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(payload),
+    });
```
paulbalaji, line 59: **[nit]** `${config.relayApiUrl}/relay` produces `//relay` if an operator configures a trailing slash. One-liner:

```ts
const baseUrl = config.relayApiUrl.replace(/\/$/, '');
// ...
const response = await fetch(`${baseUrl}/relay`, { ... });
```

.env.example
```diff
@@ -1,5 +1,6 @@
 NEXT_PUBLIC_WALLET_CONNECT_ID=12345678901234567890123456789012
 NEXT_PUBLIC_RPC_OVERRIDES='{"chain1":{"http":"https://..."}}'
+NEXT_PUBLIC_RELAY_API_URL=http://localhost:8900
```
paulbalaji, line 5: **[minor]** `http://localhost:8900` is fine for dev, but `next.config.js` sets `block-all-mixed-content` + `upgrade-insecure-requests`, so operators copying this into prod with an `http://` URL will hit silent upgrades or fetch failures. A comment makes the constraint explicit:

```
# Localhost only; production deployments must use https://
NEXT_PUBLIC_RELAY_API_URL=http://localhost:8900
```

(May become moot if you switch to a server-side proxy per the comment in `relayApi.ts`.)

paulbalaji (commented):
## Review Summary

Direction looks right — fire-and-forget relay submission gated to CCTP V2 by receipt-log detection. A few things to resolve before merge; flagging two as blocking that need author input rather than a small fix.

### Blocking — needs author response

**1. Coordinate with the relay deployment (monorepo #8710).** The relay sidecar in #8710 ships CORS allowing only `https://nexus.hyperlane.xyz` and `http://localhost:3000`, and that PR's review explicitly required Cloudflare Access in front of public DNS. This app's Vercel prod + preview origins aren't on the allowlist, and if Access is enabled the browser can't hold service-token creds. Two paths:
  - **Public endpoint:** ask infra to add this app's origins to the CORS allowlist; the sidecar's existing 100/min app-level rate limit is probably enough, but confirm WAF/per-tenant limits are in place.
  - **Server-side proxy:** add a Next.js API route (e.g. `pages/api/relay.ts`) that forwards to the relay with a server-only `RELAY_API_URL` (drop the `NEXT_PUBLIC_` prefix). Sidesteps CORS and keeps the relay URL/credentials off the client.

  Worth deciding now since the env-var name and the call-site shape both depend on it.

**2. Fire-and-forget without `.catch`** at the call site (see inline). The helper's internal `try/catch` only wraps the fetch block; the `isCctp` log scan runs before it. One-line guard at the call site fixes it.

### Other items (inline)

- Detection breadth vs PR description (minor).
- Trailing-slash URL composition (nit).
- `.env.example` defaults to `http://` which conflicts with prod CSP `block-all-mixed-content` (minor).

### Skip

CodeRabbit's `AbortController` timeout suggestion — none of the existing fetches in this codebase (`graphqlClient`, `predicateClient`, `useTokenPrice`, `useIsAccountOfacSanctioned`) use timeouts. Skipping for consistency with local convention.

### Verification

`pnpm typecheck` passes. `pnpm lint` passes (2 pre-existing warnings in mock wallet files, unrelated).

Severity legend: **major** = blocking / needs author response; **minor** = should fix; **nit** = optional polish.

src/features/transfer/relayApi.ts
```diff
@@ -0,0 +1,74 @@
...
+  originChain: string,
+  txHash: string,
+  originProtocol: ProtocolType,
+  txReceipt: TypedTransactionReceipt | null | undefined,
+): Promise<void> {
+  if (!config.relayApiUrl) return;
+
+  const isCctp =
+    originProtocol === ProtocolType.Ethereum &&
+    txReceipt != null &&
```
Xaroz: prefer the use of `isNullish` utils here to keep consistency

src/features/transfer/useTokenTransfer.ts
Xaroz, line 273: need to use `await` here

troykessler: this is intended, we just want to send it out and dont care too much if it actually succeeds, worst case the tx is 5-10s slower by classical relayer indexing

src/features/transfer/relayApi.ts
troykessler: this is fine

src/features/transfer/relayApi.ts
troykessler: this is expected, we only want to allow nexus.hyperlane.xyz and nothing else, the rc relayers actually have localhost:3000 allowed so we can test everything locally

src/features/transfer/useTokenTransfer.ts
Xaroz: ah I guess this is just "fire-and-forget" so can ignore above

Xaroz (approved):

paulbalaji (approved):
Approving — both blockers from prior review are resolved in `4e47c37c`:

- **Unhandled rejection at the call site:** entire `submitToRelayApi` body is now wrapped in `try/catch`, so the `isCctp` log scan is covered. Fire-and-forget call site is fine as-is per author intent.
- **CORS / Cloudflare Access:** confirmed by @troykessler that direct frontend → relay is the intended deployment contract, with CORS scoped to `nexus.hyperlane.xyz` (+ `localhost:3000` on RC). No proxy needed for this template app.

The minor/nit items (trailing-slash normalization, `.env.example` https note, `isNullish`) all landed in the same commit.

One small follow-up worth considering, not blocking: the `.env.example` comment could note that the value is a base URL (helper appends `/relay`), so operators don't paste `https://relay-api.hyperlane.xyz/relay` and end up calling `/relay/relay`.

Out of scope but worth confirming with infra: WAF / per-tenant rate limits on the public endpoint, given CORS is the only gate.

`pnpm typecheck` passes; `pnpm lint` passes (2 pre-existing mock-wallet warnings).

PR #1098, opened 2026-05-14
feat(chains): add featured chain sort

paulbalaji:
## Summary
- Add a hard-coded `featuredChains` config list.
- Make the chain picker default sort show featured chains first, then alphabetical.
- Keep explicit name/chain id/protocol sort options available.

## Validation
- `pnpm vitest --watch false src/features/chains/chainFilterSort.test.ts`
- `pnpm typecheck`
- `pnpm lint` (3 existing no-console warnings)

### DIFF
--- a/src/features/chains/chainFilterSort.ts
+++ b/src/features/chains/chainFilterSort.ts
@@ -1,9 +1,11 @@
 import { ProtocolType } from '@hyperlane-xyz/utils';
 
-import { ChainInfo } from './hooks';
+import { config } from '../../consts/config';
+import type { ChainInfo } from './hooks';
 
 // ── Sort ────────────────────────────────────────────────────────────
 export enum ChainSortBy {
+  Featured = 'featured',
   Name = 'name',
   ChainId = 'chain id',
   Protocol = 'protocol',
@@ -20,7 +22,7 @@ export interface SortState {
 }
 
 export const defaultSortState: SortState = {
-  sortBy: ChainSortBy.Name,
+  sortBy: ChainSortBy.Featured,
   sortOrder: SortOrder.Asc,
 };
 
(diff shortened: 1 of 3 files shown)


src/features/chains/chainFilterSort.ts
```diff
@@ -86,6 +95,19 @@ export function chainSearch({
...
+          const c1Rank = featuredChainRank.get(c1.name);
+          const c2Rank = featuredChainRank.get(c2.name);
+          const c1Featured = c1Rank !== undefined;
+          const c2Featured = c2Rank !== undefined;
+
+          if (c1Featured && c2Featured && c1Rank !== c2Rank) {
+            return sort.sortOrder === SortOrder.Asc ? c1Rank - c2Rank : c2Rank - c1Rank;
+          }
+          if (c1Featured && !c2Featured) return -1;
+          if (!c1Featured && c2Featured) return 1;
```
Xaroz: it is not gonna happen that `c1Rank === c2Rank` so the check might not be necessary, but if we wanna keep it should be also make a return path for `c1Rank === c2rank`?

Xaroz (approved):

src/features/chains/chainFilterSort.ts
paulbalaji: Good catch. Updated this to return directly when both chains are featured, so an equal rank would return `0` while preserving the configured order otherwise.

src/features/chains/chainFilterSort.ts
```diff
@@ -44,7 +46,14 @@ export function isFilterActive(filter: ChainFilterState): boolean {
...
 
-export const sortOptions = [ChainSortBy.Name, ChainSortBy.ChainId, ChainSortBy.Protocol];
+export const sortOptions = [
+  ChainSortBy.Featured,
+  ChainSortBy.Name,
+  ChainSortBy.ChainId,
+  ChainSortBy.Protocol,
+];
+
+const featuredChainRank = new Map(config.featuredChains.map((chainName, i) => [chainName, i]));
```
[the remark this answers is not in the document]
paulbalaji: Still valid. Added explicit duplicate detection while building `featuredChainRank`; the error includes the duplicate chain name and both indices so invalid `config.featuredChains` fails loudly.

PR #1109, opened 2026-05-19
fix: fixed logic merging warp core config assuming eveything is an array

xeno097:
## Summary

  Fixes a bug in `reduceOptions` (`src/features/warpCore/warpCoreConfig.ts`) where every
  non-primitive value in `WarpCoreConfig.options` was assumed to be an array and merged via
  `.concat()`. If a config supplied an object-valued option (e.g. a forward-looking
  `sealevel.altAddresses`), `(acc[key] || []).concat(o[key])` produced an array wrapping the
  object instead of merging it, breaking downstream consumers.

  ## What changed

  Type-aware merge across the three value shapes that can appear under an options key:

  - **Array** → concatenate (preserves prior behavior for `interchainFeeConstants`,
  `localFeeConstants`, `routeBlacklist`)
  - **Object** → deep-merge last-wins via `objMerge` from `@hyperlane-xyz/utils` (already
  used elsewhere in the file)
  - **Scalar** → last-write-wins, skipping nullish incoming values via `isNullish` so a later
   `null` doesn't clobber an earlier real value

  Also guards `!Array.isArray(current)` on the object branch so a prior array isn't spread as
   `{0: ..., 1: ...}` if a later config writes an object under the same key.

  ## Tests

  Added table-style tests for `reduceOptions` covering:
  - array concat across configs
  - object deep-merge (nested keys preserved last-wins)
  - scalar overwrite
  - inconsistent-shape across configs (array then object)
  - nullish incoming scalar ignored
  - empty list and undefined entries

### DIFF
--- a/src/features/warpCore/warpCoreConfig.test.ts
+++ b/src/features/warpCore/warpCoreConfig.test.ts
@@ -1,7 +1,13 @@
-import { TokenStandard } from '@hyperlane-xyz/sdk';
+import { TokenStandard, WarpCoreConfig } from '@hyperlane-xyz/sdk';
 import { describe, expect, test } from 'vitest';
 
-import { dedupeTokens, NullableAddressWarpCoreToken } from './warpCoreConfig';
+import { dedupeTokens, NullableAddressWarpCoreToken, reduceOptions } from './warpCoreConfig';
+
+// The WarpCoreConfig['options'] schema currently only declares array-valued fields,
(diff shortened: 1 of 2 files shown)


paulbalaji:
@claude review please good sire

xeno097:
Superseded by #1170, which ports this `reduceOptions` fix (including the review-requested `objMerge` change and tests) alongside the SVM offchain fee-quoting flow, all targeting `main`. Closing in favor of that PR.

PR #1113, opened 2026-05-21
fix: prevent stale transfer data flash in details modal

troykessler:
## Problem

When submitting a second transfer while the details modal is open (or being re-opened), the popup briefly flashes the **previous** transfer's sender address, origin tx hash, and message ID before switching to the new transfer's spinner.

**Root cause:** `setTransferLoading(true)` is called in `TransferTokenForm` *before* `addTransfer` runs (which happens later inside the async `triggerTransactions` call). The `useEffect` in both `SideBarMenu` and `useAutoTransferModal` (embed page) triggered on `transferLoading` changing to `true` — at that moment `transfers[last]` is still the previous completed transfer, so the modal opened with stale data.

## Solution

Trigger modal open on `transfers.length` increasing (i.e., when the new transfer is actually in the array) rather than on `transferLoading`. A `prevTransfersLengthRef` tracks the previous count so the effect only fires on genuine additions, not on mount (which would otherwise re-open the modal for persisted transfers from prior sessions).

Applied to both:
- `src/features/wallet/SideBarMenu.tsx`
- `src/pages/embed.tsx` (`useAutoTransferModal`)

## Test plan

- [ ] Submit a transfer, let the modal open and show final data (sender, origin tx, message ID)
- [ ] Submit a second transfer — modal should immediately show the spinner with no flash of the first transfer's data
- [ ] Refresh the page (persisted transfers in sidebar) — modal should not auto-open on load
- [ ] Embed page (`/embed`): same two checks above

### DIFF
--- a/src/features/wallet/SideBarMenu.tsx
+++ b/src/features/wallet/SideBarMenu.tsx
@@ -40,7 +40,6 @@ export function SideBarMenu({
   isOpen: boolean;
   onClose: () => void;
 }) {
-  const prevTransfersLengthRef = useRef(transfers.length);
   const scrollContainerRef = useRef<HTMLDivElement>(null);
   const [isMenuOpen, setIsMenuOpen] = useState(false);
   const [isModalOpen, setIsModalOpen] = useState(false);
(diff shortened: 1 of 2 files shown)


paulbalaji (approved):
Reviewed the transfer modal lifecycle changes across the sidebar and embed paths. The length-based trigger matches the store lifecycle and avoids opening on the stale pre-addTransfer state, while status updates keep the same transfer length and do not reopen the modal. CI is green. No blocking issues found.

Xaroz:
@claude review

PR #1119, opened 2026-05-25
feat: same chain ccr swaps

troykessler:
## Summary

- **Same-chain CCR swap support**: detects when origin and destination are the same chain with CrossCollateralRouter tokens; marks the transfer immediately as `Delivered` (no relay step) since the swap is atomic in a single tx
- **Synthetic message IDs**: computes a deterministic msg ID for same-chain CCR swaps from the `ReceivedTransferRemote` log — `0x00000000 || keccak256("SameChainCCR" || txHash || logIndex)[0..28]` — matching what the scraper stores for explorer display
- **Immediate balance refresh**: refetches the destination token balance right after a same-chain CCR tx confirms so the balance watcher fires the delivery toast without waiting for the next 30s poll

## Test plan

- [x] Perform a same-chain CCR swap and confirm transfer status jumps directly to Delivered
- [x] Verify the message ID in the transfer history links to the correct explorer entry
- [x] Confirm destination balance updates immediately after swap (no 30s delay)
- [x] Perform a normal cross-chain transfer and confirm no regression in status flow or msg ID extraction


src/features/transfer/TransferTokenForm.tsx
```diff
@@ -352,7 +352,21 @@ function DestinationTokenCard({ isReview }: { isReview: boolean }) {
...
+
+  // For same-chain CCR swaps the delivery is atomic — refetch the balance immediately
+  // so the balance watcher detects the increase and fires the toast without waiting
+  // for the next 30-second poll.
+  useEffect(() => {
+    if (latestTxHash && latestTransfer?.origin === latestTransfer?.destination) {
+      refetchBalance();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [latestTxHash]);
```
Xaroz: think it would be better to just call `latestTransfer.originTxHash` so you can put the dependency as `latestTransfer`, and probably refetch might be needed here too, just not sure of the side effects

src/features/transfer/utils.ts
```diff
@@ -136,6 +137,66 @@ export async function tryGetMsgIdFromTransferReceipt(
...
+/**
+ * For a same-chain CCR swap, compute the synthetic message ID the scraper stores.
+ * Finds the ReceivedTransferRemote log emitted by destRouter and applies the formula:
+ *
+ *   msgId = 0x00000000 || keccak256("SameChainCCR" || txHash32 || logIndex8)[0..28]
+ *
+ * The 4-byte zero prefix makes synthetic IDs immediately distinguishable from real
+ * Hyperlane message IDs (which are uniform keccak256 outputs).
+ */
+export function tryGetSameChainCcrMsgId(
```
Xaroz, line 158: 1. are we able to check this message id in the explorer?
2. I am assuming you can't find the `msgId` in the logs of the `originTx` so it needs to be computed?

Xaroz (approved):
just a few comments mostly lgtm

src/features/transfer/utils.ts
```diff
@@ -136,6 +137,66 @@ export async function tryGetMsgIdFromTransferReceipt(
...
+      logs = r.logs ?? [];
+      txHash = r.transactionHash ?? r.hash;
+    } else {
+      return undefined;
+    }
+
+    if (!txHash) return undefined;
+
+    const destRouterLower = destRouter.toLowerCase();
+
```
paulbalaji, line 180: Loose log matching: filters only on `destRouter + topic[0]`, picks the **first** match, and ignores the indexed `origin` (topics[1]) and `sender` (topics[2]). Today every same-chain CCR tx emits exactly one such log, so this is fine — but if `destRouter` ever batches multiple legs into one tx, the synthetic msgId will point at the wrong leg in the explorer. Consider either:

- Filtering by `topics[1] === multiProvider.getDomainId(chain)` (use the already-passed `_chain` / `_multiProvider`), and/or matching `topics[2]` against the source router
- Returning `undefined` on ambiguity (more than one matching log)

The unused `_multiProvider`, `_chain`, `_sourceRouter` params become load-bearing if you do this — otherwise they're dead weight and should be dropped.

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -352,7 +352,21 @@ function DestinationTokenCard({ isReview }: { isReview: boolean }) {
...
+
+  const transfers = useStore((s) => s.transfers);
+  const latestTransfer = transfers[transfers.length - 1];
+  const latestTxHash = latestTransfer?.originTxHash;
+
+  // For same-chain CCR swaps the delivery is atomic — refetch the balance immediately
+  // so the balance watcher detects the increase and fires the toast without waiting
+  // for the next 30-second poll.
+  useEffect(() => {
+    if (latestTxHash && latestTransfer?.origin === latestTransfer?.destination) {
```
paulbalaji: Heuristic is broader than the comment implies. It refetches on **any** latest transfer with `origin === destination` — including non-CCR same-chain transfers (if any ever exist) or stale entries from previous sessions. Tightening suggestions:

- Match the final state: `status === TransferStatus.Delivered && destinationTxHash === originTxHash`
- Also check `latestTransfer`'s destination chain/token/recipient matches the current card, so an unrelated historical transfer doesn't kick the watcher when the user later switches destinations.

Worst case today is a spurious refetch + misleading delivery toast on an unrelated chain pair.

src/features/transfer/TransferTokenForm.tsx
paulbalaji, line 379: Non-blocking response to the earlier review: keeping `[latestTxHash]` is correct — don't switch to `[latestTransfer]`. The store mutates transfer objects in place on status transitions (Preparing → CreatingTxs → ...), so an object-reference dep would either miss the `originTxHash` update or fire on every status tick depending on how Zustand returns the slice. The hash string is the right signal. Worth a one-line comment above the `eslint-disable` explaining this so the next reader doesn't "fix" it.

paulbalaji (approved):
Approving — the same-chain CCR flow is sound and the existing review threads are addressed (relay-API guard added; synthetic msgId formula matches the Rust scraper fixture; `Delivered`-direct status is handled correctly downstream). Three non-blocking notes inline + one observation outside the diff.

- `src/features/transfer/TransfersDetailsModal.tsx:99` — `shouldTrackDelivery = isSent && !isFailed && !!msgId`. `Delivered` is in `SentTransferStatuses`, so once a same-chain CCR transfer lands the modal still polls `useMessageDeliveryStatus` every 10s while open. Cheap fix in a follow-up: also gate on `status !== TransferStatus.Delivered`.
- Worth adding a WUI-side unit test pinning the synthetic msgId formula against a known fixture, so a future scraper change can't drift silently.

src/features/transfer/utils.ts
troykessler: 1. yes the same chain swaps have been already indexed with this synthetic message id
2. yes exactly, for same chain swaps we have no dispatch and process, so we compute a synthetic msgId so we can still show them on the explorer without any big changes

PR #1121, opened 2026-05-27
fix: silence smart contract check RPC errors from user-facing toast

troykessler:
## Summary

- The `isSmartContract` check (`eth_getCode`) is a best-effort UX guard that warns users when sending to their own smart contract wallet on a chain where it doesn't exist as a contract
- When the underlying RPC call fails — likely rate limiting on public BSC RPCs, but also any transient network failure — it was surfacing as `toast.error`, showing confusing errors to users even though the transfer itself is completely unaffected
- Downgrade from `toast.error` to `logger.warn` so failures are still captured in logs but not shown to users

## Test plan

- [ ] Bridge between two EVM chains with a smart contract wallet — warning banner still appears when expected
- [ ] Confirm no error toast appears when BSC RPC is flaky


troykessler:
Closing in favor of #1122 — this branch was based on nexus and included too many unrelated commits.

PR #1127, opened 2026-05-28
feat(prices): shared CoinGecko cache + swap USD display

Xaroz:
## Summary

Unifies bridge and swap price fetching behind a single store-backed CoinGecko cache, and wires USD value into the swap form. Engine `/v1/tokens` now exposes `coinGeckoId` per token, so the swap side can finally key prices the same way the bridge does.

## Shared price cache

- New store field `tokenPrices: Record<string, { usd?, fetchedAt }>` + `mergeTokenPrices(requestedIds, fetched)`. Per-id `fetchedAt` so chain switches that grow the token catalogue only fetch the deltas; entries with `usd` absent are negative-cached so we don't retry IDs CoinGecko has no price for in a tight loop.
- New shared hook `useTokenPricesByIds(ids)` in `src/features/tokens/useTokenPrice.tsx` doing the delta-fetch into that cache. Batches at 250 IDs/request (CoinGecko free-tier rate-limits per-request, so bigger batches = fewer throttles). 1h cache. Bridge `useTokenPrices()` wraps it with the warpCore `coinGeckoIds`; swap `useTokenPrices()` wraps it with IDs derived + deduped from `knownTokens`. Overlapping IDs (ETH, USDC, ...) are fetched once total across both tabs.

## Failure handling

- `fetchPrices` throws on HTTP/network errors instead of swallowing into `{}` — lets the caller distinguish a real failure from a 200-OK-with-missing-IDs.
- `fetchPricesBatched` uses `Promise.allSettled` per-chunk so a 429 on one chunk doesn't poison the rest.
- queryFn: partial success merges what came back, next render's smaller `idsToFetch` fires a follow-up automatically. Total failure throws so TQ retries with default exponential backoff instead of caching an empty success.

## Stability

- Input `ids` is debounced (300ms) so rapid catalogue growth from the picker batches into one queryKey update — no overlapping in-flight fetches on the first page load. Debounced on input rather than the filter result so post-merge delta fetches fire without cascading 300ms waits.

## Swap UI

- New `useTokenUsdValue(token, amount)` convenience hook so origin + destination cards don't duplicate the price-lookup boilerplate. Coerces and strips grouping separators before `parseFloat` so a number/null/comma-formatted amount doesn't blow up.
- `SwapForm` shows `$1.23` next to the typed amount on origin and next to the expected output on destination. Destination row keeps the existing `Min: 0.99 ETH` indicator alongside USD.
- `getUsdValue` in `swap/balances/utils.ts` keys on `token.coinGeckoId` now that the engine provides it (was previously keying on lowercase symbol as a stopgap).

### DIFF
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -132,6 +132,13 @@ export interface AppState {
   knownTokens: Map<string, UiToken>;
   syncTokens: (tokens: UiToken[]) => void;
 
+  // Sticky USD price cache, keyed by coinGeckoId. Per-id `fetchedAt` so
+  // chain switches that grow the token catalogue only fetch the deltas.
+  // Entries with `usd` absent are negative-cached (CoinGecko had no
+  // price) so we don't retry in a tight loop.
+  tokenPrices: Record<string, { usd?: number; fetchedAt: number }>;
+  mergeTokenPrices: (requestedIds: string[], fetched: Record<string, number>) => void;
+
   // Shared component state
   transferLoading: boolean;
   setTransferLoading: (isLoading: boolean) => void;
@@ -327,6 +334,18 @@ export const useStore = create<AppState>()(
             }
           }
           return added > 0 ? { knownTokens: next } : state;
+        });
+      },
+
+      tokenPrices: {},
+      mergeTokenPrices: (requestedIds, fetched) => {
+        set((state) => {
+          const now = Date.now();
+          const next = { ...state.tokenPrices };
+          for (const id of requestedIds) {
+            next[id] = { usd: fetched[id], fetchedAt: now };
+          }
+          return { tokenPrices: next };
         });
       },
 
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -160,12 +160,14 @@ export interface AppState {
   // at WarpContext init). Consumed by the bridge `useTokenPrices` wrapper
   // which delegates to the shared `useTokenPricesByIds` cache below.
   coinGeckoIds: string[];
-  // Sticky USD price cache, keyed by coinGeckoId. Per-id `fetchedAt` so
-  // chain switches that grow the token catalogue only fetch the deltas.
-  // Entries with `usd` absent are negative-cached (CoinGecko had no
-  // price) so we don't retry in a tight loop.
-  tokenPrices: Record<string, { usd?: number; fetchedAt: number }>;
-  mergeTokenPrices: (requestedIds: string[], fetched: Record<string, number>) => void;
+  // Session-scoped USD price cache, keyed by coinGeckoId. `failedAt`
+  // backs off retries after rate-limit / network failures.
+  tokenPrices: Record<string, { usd?: number; fetchedAt?: number; failedAt?: number }>;
+  mergeTokenPrices: (
+    succeededIds: string[],
+    fetched: Record<string, number>,
+    failedIds: string[],
+  ) => void;
 }
 
 export const useStore = create<AppState>()(
@@ -339,12 +341,15 @@ export const useStore = create<AppState>()(
       },
 
       tokenPrices: {},
-      mergeTokenPrices: (requestedIds, fetched) => {
+      mergeTokenPrices: (succeededIds, fetched, failedIds) => {
         set((state) => {
           const now = Date.now();
           const next = { ...state.tokenPrices };
-          for (const id of requestedIds) {
+          for (const id of succeededIds) {
             next[id] = { usd: fetched[id], fetchedAt: now };
+          }
+          for (const id of failedIds) {
+            next[id] = { ...next[id], failedAt: now };
           }
           return { tokenPrices: next };
         });
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -132,13 +132,6 @@ export interface AppState {
   knownTokens: Map<string, UiToken>;
   syncTokens: (tokens: UiToken[]) => void;
 
-  // Sticky USD price cache, keyed by coinGeckoId. Per-id `fetchedAt` so
-  // chain switches that grow the token catalogue only fetch the deltas.
-  // Entries with `usd` absent are negative-cached (CoinGecko had no
-  // price) so we don't retry in a tight loop.
-  tokenPrices: Record<string, { usd?: number; fetchedAt: number }>;
-  mergeTokenPrices: (requestedIds: string[], fetched: Record<string, number>) => void;
-
   // Shared component state
   transferLoading: boolean;
   setTransferLoading: (isLoading: boolean) => void;
@@ -163,8 +156,16 @@ export interface AppState {
   // Set of router addresses per chain — used to prevent sending to warp route
   // addresses and to filter message API results
   routerAddressesByChainMap: Record<ChainName, Set<string>>;
-  // Deduplicated, sorted CoinGecko IDs for all tokens (used by useTokenPrices)
+  // Deduplicated, sorted CoinGecko IDs for the warpCore token set (built
+  // at WarpContext init). Consumed by the bridge `useTokenPrices` wrapper
+  // which delegates to the shared `useTokenPricesByIds` cache below.
   coinGeckoIds: string[];
+  // Sticky USD price cache, keyed by coinGeckoId. Per-id `fetchedAt` so
+  // chain switches that grow the token catalogue only fetch the deltas.
+  // Entries with `usd` absent are negative-cached (CoinGecko had no
+  // price) so we don't retry in a tight loop.
+  tokenPrices: Record<string, { usd?: number; fetchedAt: number }>;
+  mergeTokenPrices: (requestedIds: string[], fetched: Record<string, number>) => void;
 }
 
 export const useStore = create<AppState>()(
--- a/src/features/swap/SwapForm.tsx
+++ b/src/features/swap/SwapForm.tsx
@@ -24,7 +24,7 @@ import { RecipientConfirmationModal } from '../wallet/RecipientConfirmationModal
 import { WalletConnectionWarning } from '../wallet/WalletConnectionWarning';
 import { WalletDropdown } from '../wallet/WalletDropdown';
 import { useTokenBalance } from './balances/hooks';
-import { formatBalance, formatFeeAmount } from './balances/utils';
+import { formatBalance, formatFeeAmount, formatUsd } from './balances/utils';
 import { FeeSectionButton } from './FeeSectionButton';
 import { MaxButton } from './MaxButton';
 import {
@@ -38,6 +38,7 @@ import { TokenBalance } from './TokenBalance';
 import { getTokenByKeyFromMap, useTokenByKeyMap } from './tokens/hooks';
 import { TokenSelectField } from './tokens/TokenSelectField';
 import type { UiToken } from './tokens/types';
+import { useTokenUsdValue } from './tokens/useTokenPrice';
 import {
   FinalSwapStatuses,
   SwapStatus,
@@ -495,7 +496,9 @@ function OriginTokenCard({
   srcToken: UiToken | undefined;
   amountError: string | undefined;
 }) {
+  const { values } = useFormikContext<SwapFormValues>();
   const { data: balance, isLoading: isBalanceLoading } = useTokenBalance(srcToken);
+  const amountUsd = useTokenUsdValue(srcToken, values.amount);
 
   return (
     <div>
@@ -530,7 +533,7 @@ function OriginTokenCard({
           />
         </div>
         <div className="transfer-balance mt-1 flex items-center justify-between text-xs leading-[18px] text-gray-450 dark:text-foreground-secondary">
-          <span>$0.00</span>
+          <span>{formatUsd(amountUsd)}</span>
           <TokenBalance label="Balance" balance={balance ?? null} token={srcToken} />
         </div>
       </div>
@@ -572,14 +575,14 @@ function DestinationTokenCard({
     }
   }, [bestRoute, dstToken]);
   const minOutputDisplay = useMemo(() => {
-    if (!bestRoute || !dstToken) return '';
-    if (bestRoute.isBridgeOnly) return '';
+    if (!bestRoute || !dstToken || bestRoute.isBridgeOnly) return '';
     try {
       return formatUnits(BigInt(bestRoute.raw.outputMin), dstToken.decimals);
     } catch {
       return '';
     }
   }, [bestRoute, dstToken]);
+  const outputUsd = useTokenUsdValue(dstToken, outputDisplay);
 
   return (
     <div>
@@ -611,7 +614,12 @@ function DestinationTokenCard({
         </div>
         <div className="transfer-balance mt-1 flex items-center justify-between text-xs leading-[18px] text-gray-450 dark:text-foreground-secondary">
           <span>
-            {minOutputDisplay ? `Min: ${minOutputDisplay} ${dstToken?.symbol ?? ''}` : ''}
+            {formatUsd(outputUsd)}
+            {minOutputDisplay && (
+              <span className="ml-1.5">
+                · Min: {minOutputDisplay} {dstToken?.symbol ?? ''}
+              </span>
+            )}
           </span>
           <TokenBalance label="Remote Balance" balance={balance ?? null} token={dstToken} />
         </div>
--- a/src/features/swap/SwapForm.tsx
+++ b/src/features/swap/SwapForm.tsx
@@ -533,7 +533,7 @@ function OriginTokenCard({
           />
         </div>
         <div className="transfer-balance mt-1 flex items-center justify-between text-xs leading-[18px] text-gray-450 dark:text-foreground-secondary">
-          <span>{formatUsd(amountUsd)}</span>
+          <span>{amountUsd >= 0.01 ? formatUsd(amountUsd) : '$0.00'}</span>
           <TokenBalance label="Balance" balance={balance ?? null} token={srcToken} />
         </div>
       </div>
@@ -574,14 +574,6 @@ function DestinationTokenCard({
       return '';
     }
   }, [bestRoute, dstToken]);
-  const minOutputDisplay = useMemo(() => {
-    if (!bestRoute || !dstToken || bestRoute.isBridgeOnly) return '';
-    try {
-      return formatUnits(BigInt(bestRoute.raw.outputMin), dstToken.decimals);
-    } catch {
-      return '';
-    }
-  }, [bestRoute, dstToken]);
   const outputUsd = useTokenUsdValue(dstToken, outputDisplay);
 
   return (
@@ -613,14 +605,7 @@ function DestinationTokenCard({
           />
         </div>
         <div className="transfer-balance mt-1 flex items-center justify-between text-xs leading-[18px] text-gray-450 dark:text-foreground-secondary">
-          <span>
-            {formatUsd(outputUsd)}
-            {minOutputDisplay && (
-              <span className="ml-1.5">
-                · Min: {minOutputDisplay} {dstToken?.symbol ?? ''}
-              </span>
-            )}
-          </span>
+          <span>{outputUsd >= 0.01 ? formatUsd(outputUsd) : '$0.00'}</span>
           <TokenBalance label="Remote Balance" balance={balance ?? null} token={dstToken} />
         </div>
       </div>
--- a/src/features/swap/SwapForm.tsx
+++ b/src/features/swap/SwapForm.tsx
@@ -38,7 +38,7 @@ import { TokenBalance } from './TokenBalance';
 import { getTokenByKeyFromMap, useTokenByKeyMap } from './tokens/hooks';
 import { TokenSelectField } from './tokens/TokenSelectField';
 import type { UiToken } from './tokens/types';
-import { useTokenUsdValue } from './tokens/useTokenPrice';
+import { useTokenPrices, useTokenUsdValue } from './tokens/useTokenPrice';
 import {
   FinalSwapStatuses,
   SwapStatus,
@@ -69,6 +69,11 @@ export function SwapForm() {
 }
 
 function SwapFormContent() {
+  // Mounts the catalogue-wide price fetch once for the whole form. Cards
+  // and review modal read individual prices via `useTokenUsdValue` (pure
+  // store readers); without this top-level call, deep-linked URLs would
+  // render cards before the picker opens and see empty USD values.
+  useTokenPrices();
   const { values, errors, setErrors, setFieldValue, setValues } =
     useFormikContext<SwapFormValues>();
   const multiProvider = useMultiProvider();
@@ -533,7 +538,7 @@ function OriginTokenCard({
           />
         </div>
         <div className="transfer-balance mt-1 flex items-center justify-between text-xs leading-[18px] text-gray-450 dark:text-foreground-secondary">
-          <span>{amountUsd >= 0.01 ? formatUsd(amountUsd) : '$0.00'}</span>
+          <span>{!amountUsd ? '$0.00' : formatUsd(amountUsd)}</span>
           <TokenBalance label="Balance" balance={balance ?? null} token={srcToken} />
         </div>
       </div>
@@ -605,7 +610,7 @@ function DestinationTokenCard({
           />
         </div>
         <div className="transfer-balance mt-1 flex items-center justify-between text-xs leading-[18px] text-gray-450 dark:text-foreground-secondary">
-          <span>{outputUsd >= 0.01 ? formatUsd(outputUsd) : '$0.00'}</span>
+          <span>{!outputUsd ? '$0.00' : formatUsd(outputUsd)}</span>
           <TokenBalance label="Remote Balance" balance={balance ?? null} token={dstToken} />
         </div>
       </div>
--- a/src/features/swap/tokens/useTokenPrice.tsx
+++ b/src/features/swap/tokens/useTokenPrice.tsx
@@ -1,6 +1,36 @@
-// Engine doesn't expose token USD prices yet. Shape mirrors warp UI's
-// useTokenPrices so TokenList consumes it without changes; the price
-// map is empty for now.
+import { useMemo } from 'react';
+
+import { useStore } from '../../store';
+import { useTokenPricesByIds } from '../../tokens/useTokenPrice';
+import type { UiToken } from './types';
+
+// Swap-side prices: derive coinGeckoIds from the engine token catalogue
+// (knownTokens grows as the user browses chains/tokens) and delegate to
+// the shared store-backed cache. Overlapping IDs with the bridge tab
+// (ETH, USDC, etc.) are fetched once total.
 export function useTokenPrices(): { prices: Record<string, number>; isLoading: boolean } {
-  return { prices: {}, isLoading: false };
+  const knownTokens = useStore((s) => s.knownTokens);
+
+  const ids = useMemo(() => {
+    const set = new Set<string>();
+    for (const t of knownTokens.values()) {
+      if (t.coinGeckoId) set.add(t.coinGeckoId);
+    }
+    return Array.from(set).sort();
+  }, [knownTokens]);
+
+  return useTokenPricesByIds(ids);
+}
+
+// Convenience: USD value of a decimal-formatted token `amount` string.
+// Returns 0 when token has no coinGeckoId, no cached price, or the
+// amount fails to parse.
+export function useTokenUsdValue(token: UiToken | undefined, amount: string): number {
+  const { prices } = useTokenPrices();
+  const price = token?.coinGeckoId ? prices[token.coinGeckoId] : undefined;
+  return useMemo(() => {
+    const a = parseFloat(amount);
+    if (!price || isNaN(a)) return 0;
+    return a * price;
+  }, [amount, price]);
 }
--- a/src/features/swap/tokens/useTokenPrice.tsx
+++ b/src/features/swap/tokens/useTokenPrice.tsx
@@ -29,7 +29,9 @@ export function useTokenUsdValue(token: UiToken | undefined, amount: string): nu
   const { prices } = useTokenPrices();
   const price = token?.coinGeckoId ? prices[token.coinGeckoId] : undefined;
   return useMemo(() => {
-    const a = parseFloat(amount);
+    // Coerce + strip grouping separators — Formik may hand us a number/null
+    // and parseFloat("1,234.56") returns 1 without the strip.
+    const a = parseFloat(String(amount ?? '').replace(/,/g, ''));
     if (!price || isNaN(a)) return 0;
     return a * price;
   }, [amount, price]);
--- a/src/features/swap/tokens/useTokenPrice.tsx
+++ b/src/features/swap/tokens/useTokenPrice.tsx
@@ -22,19 +22,19 @@ export function useTokenPrices(): { prices: Record<string, number>; isLoading: b
   return useTokenPricesByIds(ids);
 }
 
-// USD value of a token `amount` string. Reads the single price via a
-// parameterized selector so consumers don't re-render on unrelated
-// tokenPrices mutations.
-export function useTokenUsdValue(token: UiToken | undefined, amount: string): number {
-  // Side-effect mount: keeps the catalogue-wide fetch hot for deep-linked
-  // URLs where cards render before the picker opens.
-  useTokenPrices();
+// USD value of a token `amount` string. Returns null when there's no info
+// (no coinGeckoId, no cached price, NaN/Infinity amount) so call sites can
+// distinguish "no data → $0.00" from "tiny positive → <$0.01" via formatUsd.
+// Pure read — relies on a parent mounting `useTokenPrices()` to keep the
+// store cache populated (see SwapFormContent). Per-id selector keeps
+// consumers from re-rendering on unrelated tokenPrices mutations.
+export function useTokenUsdValue(token: UiToken | undefined, amount: string): number | null {
   const id = token?.coinGeckoId;
   const price = useStore((s) => (id ? s.tokenPrices[id]?.usd : undefined));
   return useMemo(() => {
     // Strip grouping separators — parseFloat("1,234.56") returns 1 otherwise.
     const a = parseFloat(String(amount ?? '').replace(/,/g, ''));
-    if (!price || isNaN(a)) return 0;
+    if (!price || !Number.isFinite(a)) return null;
     return a * price;
   }, [amount, price]);
 }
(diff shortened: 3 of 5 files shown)


src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,130 @@
...
-    }
-    const data: CoinGeckoResponse = await res.json();
-    const result: Record<string, number> = {};
-    for (const [id, priceData] of Object.entries(data)) {
-      if (priceData?.usd != null) result[id] = priceData.usd;
+  const res = await fetch(
+    `https://api.coingecko.com/api/v3/simple/price?ids=${ids.join(',')}&vs_currencies=usd`,
+  );
+  if (!res.ok) {
+    throw new Error(`CoinGecko ${res.status} ${res.statusText}`);
```
xeno097: maybe stringify the body (if there is one) too sop we can have detailed msgs when the request fails

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,130 @@
+import { useDebounce } from '@hyperlane-xyz/widgets';
 import { useQuery } from '@tanstack/react-query';
+import { useMemo } from 'react';
 
 import { logger } from '../../utils/logger';
 import { useStore } from '../store';
 
-const PRICE_STALE_TIME = 300_000; // 5 min
+const PRICE_CACHE_MS = 60 * 60_000; // 1 hour
```
xeno097: 1h TTL is a big jump from the previous 5 min. ETH can move 3-5% in an hour, alts 10-20% — the swap `you receive ~$X` readout will be visibly stale. Consider 5-15 min.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,130 @@
...
-    if (!res.ok) {
-      logger.warn(`CoinGecko API error: ${res.status} ${res.statusText}`);
-      return {};
-    }
-    const data: CoinGeckoResponse = await res.json();
-    const result: Record<string, number> = {};
-    for (const [id, priceData] of Object.entries(data)) {
-      if (priceData?.usd != null) result[id] = priceData.usd;
+  const res = await fetch(
+    `https://api.coingecko.com/api/v3/simple/price?ids=${ids.join(',')}&vs_currencies=usd`,
```
xeno097: Two suggestions:

1. Add `NEXT_PUBLIC_COINGECKO_URL` (and optionally `NEXT_PUBLIC_COINGECKO_API_KEY` → `x-cg-pro-api-key` header). Every other external service in `consts/config.ts` (`NEXT_PUBLIC_REGISTRY_URL`, `NEXT_PUBLIC_FEE_QUOTING_URL`, …) has an override; this is the odd one out, and pro-tier removes the rate-limit class of problems.
2. As you flagged earlier — on `!res.ok`, include `await res.text().catch(() => '')` in the error message. Helps diagnose CDN interstitials / 4xx bodies where the status alone isn't enough.

Also: `ids.join(',')` should probably go through `encodeURIComponent` per id — slugs are safe today, but it's a cheap guarantee against future surprises.

xeno097 (commented):
## Brutally honest review (consolidated)

Good direction overall — moving from in-flight TQ dedup to a store-backed delta cache is the right call, and the `getUsdValue` switch from symbol → `coinGeckoId` eliminates a latent cross-chain collision. A few things I'd land before merging:

### Highest-priority
1. **`formatUsd(0)` renders `<$0.01`, not `$0.00`** — visible regression vs. bridge (`TransferTokenForm.tsx` explicitly gates with `>= 0.01` and falls back to `$0.00`). Both swap cards show `<$0.01` during price load / when token has no `coinGeckoId` / when amount is empty.
2. **Partial-failure refetch storm.** Successful chunks merge into the store → `idsToFetch` shrinks → new `queryKey` → next fetch fires immediately for the still-missing ids, **bypassing TQ's exponential backoff** (which only applies within a single key). With 429s on the failing chunks this hot-loops.
3. **No recovery from total failure.** With `refetchOnMount/Focus/Interval: false`, if all chunks fail 3× retries, `queryKey` never changes again → query stuck in error state until reload. The "queryKey shrinks on partial success" pattern only saves us if at least one chunk succeeds.
4. **Zero rate-limiting** on `Promise.allSettled(chunks.map(...))`. Free-tier CoinGecko is ~5–15 req/min. Once swap's `knownTokens` crosses 250 ids, chain browsing fires 2–3 parallel requests. Combined with the hot-loop in #2, deterministic 429 lockout in <1 min of normal use is plausible. Capping concurrency to 1–2 with a small inter-chunk delay defuses #2 and #4 together.

### Worth fixing
5. 1h TTL is aggressive for a financial UI (was 5 min). ETH routinely moves 3–5% in an hour. Consider 5–15 min.
6. `useStore((s) => s.tokenPrices)` whole-object subscription → every consumer (origin card, destination card, both `TokenList`s, bridge `useTokenPrices`) re-renders on every merge. Decompose to per-id selector or `useStoreWithEqualityFn` + shallow.
7. `staleTime: PRICE_CACHE_MS` on the `useQuery` is dead code — `queryKey` content changes on every successful merge, so `staleTime` doesn't gate anything. The real TTL lives in the `idsToFetch` filter.
8. No `NEXT_PUBLIC_COINGECKO_URL` / API-key env. Every other external service in `consts/config.ts` has overrides; this is inconsistent. Also addresses the earlier inline ask about logging response bodies on failure.
9. "Sticky USD price cache" comment in `store.ts` — it's not persisted (correctly excluded from `partialize`), so it's sticky only within the tab session. Worth clarifying the wording.

### Solid
- Per-id `fetchedAt` for delta-fetching across chain switches.
- `Promise.allSettled` per chunk so a 429 on one doesn't poison the rest.
- `throw` on total failure instead of caching `{}` (good reasoning in the comment).
- `tokenPrices` correctly excluded from `partialize`.
- `getUsdValue` keying by `coinGeckoId` rather than symbol.

- `src/utils/amount.ts:11`: `formatUsd(0)` returning `<$0.01` is the root of finding #1 — fix could live there (special-case `value === 0` → `'$0.00'`) rather than at every call site.
- `node_modules/@hyperlane-xyz/widgets` `useDebounce` uses referential `useEffect([value])`. Steady-state OK here because `syncTokens` ref-stabilizes `knownTokens`, but during initial bootstrap the rapid `knownTokens` mutations can keep resetting the 300ms timer → USD display stays empty until token loading quiets. Worth verifying with a throttled-network manual test.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,130 @@
...
+// rest. Only successful chunks contribute to `requestedIds` / `prices`;
+// failed chunks' IDs stay uncached and become eligible for refetch on
+// the next render.
+async function fetchPricesBatched(ids: string[]): Promise<BatchedFetchResult> {
+  if (ids.length === 0) return { requestedIds: [], prices: {} };
+  const chunks: string[][] = [];
+  for (let i = 0; i < ids.length; i += COINGECKO_BATCH) {
+    chunks.push(ids.slice(i, i + COINGECKO_BATCH));
+  }
+  const results = await Promise.allSettled(chunks.map(fetchPrices));
```
xeno097: All chunks fire simultaneously. CoinGecko free tier is ~5-15 req/min — once `knownTokens` crosses 250 ids, every catalogue-growth event fires 2-3 parallel requests, and the partial-failure flow below can trigger an immediate refetch without backoff. Realistic to get rate-limited out within a minute of normal browsing.

Capping concurrency to 1-2 (e.g. simple `for…of await` loop with a small inter-chunk delay) defuses this and the partial-failure refetch issue together.

Also: the comment on L9-11 says `250 matches /coins/markets per_page cap` — different endpoint, different constraint. `/simple/price` URL length at 250 ids is ~3KB+ (under typical 8KB limits, but Cloudflare-fronted endpoints have been known to 414 around 6-8KB). Worth measuring or dropping to ~100.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,130 @@
...
-  const coinGeckoIds = useStore((s) => s.coinGeckoIds);
+// Shared core: delta-fetch USD prices for `ids` into the store's
+// `tokenPrices` cache. Already-cached entries < 1h old are reused —
+// only the new IDs hit CoinGecko. Bridge + swap both wrap this with
+// their own ID sources (warp routes vs engine token catalogue).
+export function useTokenPricesByIds(ids: string[]): {
+  prices: Record<string, number>;
+  isLoading: boolean;
+} {
+  const tokenPrices = useStore((s) => s.tokenPrices);
```
xeno097, line 113: Subscribing to the whole `tokenPrices` object means every `mergeTokenPrices` call (which creates a fresh object reference) re-renders **every** consumer — both `TokenList`s, both swap cards via `useTokenUsdValue`, plus the bridge `useTokenPrices` — regardless of whether the id they care about changed.

For the swap cards specifically, where `useTokenUsdValue(token, amount)` only needs a single price, consider a parameterized selector: `useStore((s) => s.tokenPrices[id]?.usd)`. For broader consumers, `useStoreWithEqualityFn` + `shallow` over a flattened map keyed by `id`.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,130 @@
...
+  const mergeTokenPrices = useStore((s) => s.mergeTokenPrices);
+  // Debounce the input id set — rapid catalogue growth (e.g. picker
+  // browsing chains) batches into one queryKey update instead of firing
+  // redundant in-flight fetches that all overlap on the same IDs.
+  const debouncedIds = useDebounce(ids, 300);
```
xeno097, line 118: Minor / verify manually: `useDebounce` in widgets uses referential `useEffect([value])`. The steady state here is fine because `syncTokens` returns `state` when nothing was added, keeping `knownTokens` ref-stable.

But during initial token bootstrap (multiple `useTokens` pages → multiple `syncTokens` calls), `knownTokens` mutates rapidly → fresh `ids` arrays → 300ms timer keeps resetting → USD display stays empty until token loading quiets. On a fast network it's invisible; on a throttled connection it can be noticeable. Worth a manual test with `Slow 3G` in DevTools.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,130 @@
...
+      // to just the still-missing IDs → fresh queryKey fires a follow-up
+      // fetch for those automatically.
+      if (requestedIds.length > 0) {
+        mergeTokenPrices(requestedIds, prices);
+        return prices;
+      }
+      // Total failure: throw so TQ marks errored + retries (default 3x
+      // with exp backoff). Returning {} here would cache an empty
+      // "success" and lock the same queryKey out of refetching.
+      throw new Error('All CoinGecko price chunks failed');
```
xeno097: **Hot-loop risk on partial failure.** Walk through:
1. Chunk A succeeds, chunk B 429s.
2. `mergeTokenPrices` writes A's ids → `tokenPrices` mutates → `idsToFetch` recomputes to just B's ids → new `queryKey`.
3. TQ sees a new key, fires immediately. No backoff applied (TQ's 3x exp backoff is per-queryKey).
4. B fails again → throw → query errors → but the `idsToFetch` array is unchanged, so the same key stays.

With 429 specifically, this hammers CoinGecko at render-rate until something else clears the loop.

Fix shape: track `failedAt` per-id alongside `fetchedAt` in the store. `idsToFetch` filter additionally short-circuits ids with a recent `failedAt` (e.g., 30s window). That also resolves the "all chunks fail → query permanently errored" recovery case — stale `failedAt` becomes refetch-eligible naturally.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,130 @@
...
+  const { isLoading } = useQuery({
+    queryKey: ['tokenPrices', idsToFetch],
+    queryFn: async () => {
+      const { requestedIds, prices } = await fetchPricesBatched(idsToFetch);
+      // Partial: merge what we got. Next render's `idsToFetch` shrinks
+      // to just the still-missing IDs → fresh queryKey fires a follow-up
+      // fetch for those automatically.
+      if (requestedIds.length > 0) {
+        mergeTokenPrices(requestedIds, prices);
+        return prices;
```
xeno097: `return prices` from `queryFn` is never read — consumers get prices from the `tokenPrices` subscription (L76 + L115). The query is only being used as a debounced trigger. Either drop the return value, or actually use `data` in the hook instead of subscribing to the store. Mixed paths make the data flow harder to follow.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,130 @@
...
+      }
+      // Total failure: throw so TQ marks errored + retries (default 3x
+      // with exp backoff). Returning {} here would cache an empty
+      // "success" and lock the same queryKey out of refetching.
+      throw new Error('All CoinGecko price chunks failed');
+    },
+    enabled: idsToFetch.length > 0,
+    staleTime: PRICE_CACHE_MS,
```
xeno097: `staleTime: PRICE_CACHE_MS` doesn't do what it looks like it does. `queryKey` content changes whenever `idsToFetch` content changes (which is most successful merges), so TQ treats each as a separate query — `staleTime` only gates refetches against an existing queryKey that we rarely revisit. The real TTL lives in the `idsToFetch` filter at L83-89. Either drop the option or add a comment noting it's vestigial.

Related: with `refetchInterval: false, refetchOnMount: false, refetchOnWindowFocus: false`, after 3 retry exhaustion on a fully-failed fetch the query is stuck. See my comment on L98-105 — a `failedAt` short-circuit in the filter solves both.

src/features/swap/SwapForm.tsx
```diff
@@ -530,7 +533,7 @@ function OriginTokenCard({
           />
         </div>
         <div className="transfer-balance mt-1 flex items-center justify-between text-xs leading-[18px] text-gray-450 dark:text-foreground-secondary">
-          <span>$0.00</span>
+          <span>{formatUsd(amountUsd)}</span>
```
xeno097: `formatUsd(0)` returns `<$0.01` (see `src/utils/amount.ts:11`) — so the origin card renders `<$0.01` during price load, when `srcToken.coinGeckoId` is missing, or when `values.amount` is empty.

The bridge handles this explicitly (`TransferTokenForm.tsx:358-368` — `shouldShowPrice = totalTokenPrice >= 0.01`, falls back to literal `$0.00`). Two surfaces in the same product showing different placeholders is rough.

Either gate at the call site or teach `formatUsd` to special-case `value === 0` → `'$0.00'`. Also worth deciding: during an in-flight load, would a skeleton / `'—'` be clearer than either `$0.00` or `<$0.01`?

src/features/swap/SwapForm.tsx
```diff
@@ -611,7 +614,12 @@ function DestinationTokenCard({
         </div>
         <div className="transfer-balance mt-1 flex items-center justify-between text-xs leading-[18px] text-gray-450 dark:text-foreground-secondary">
           <span>
-            {minOutputDisplay ? `Min: ${minOutputDisplay} ${dstToken?.symbol ?? ''}` : ''}
+            {formatUsd(outputUsd)}
+            {minOutputDisplay && (
+              <span className="ml-1.5">
+                · Min: {minOutputDisplay} {dstToken?.symbol ?? ''}
+              </span>
+            )}
```
xeno097: Two concerns on this line:

1. Same `formatUsd(0)` → `<$0.01` issue as the origin card. With no `coinGeckoId` on `dstToken` or no cached price, the line reads `<$0.01 · Min: 0.0001 TOKEN` — looks like a real price exists when none does.
2. **The USD here reflects `outputDisplay` (expected/max output), but renders adjacent to a `Min:` label.** Easy to misread as "Min worth $X". Either also compute USD off `outputMin` (parse `bestRoute.raw.outputMin` the same way), or stack USD on its own row to break the visual association.

Minor: `·` is a one-off in this codebase. If kept, add a space and `aria-hidden="true"` so screen readers don't read "middle dot".

src/features/store.ts
```diff
@@ -156,8 +156,16 @@ export interface AppState {
...
-  // Deduplicated, sorted CoinGecko IDs for all tokens (used by useTokenPrices)
+  // Deduplicated, sorted CoinGecko IDs for the warpCore token set (built
+  // at WarpContext init). Consumed by the bridge `useTokenPrices` wrapper
+  // which delegates to the shared `useTokenPricesByIds` cache below.
   coinGeckoIds: string[];
+  // Sticky USD price cache, keyed by coinGeckoId. Per-id `fetchedAt` so
+  // chain switches that grow the token catalogue only fetch the deltas.
+  // Entries with `usd` absent are negative-cached (CoinGecko had no
+  // price) so we don't retry in a tight loop.
+  tokenPrices: Record<string, { usd?: number; fetchedAt: number }>;
```
xeno097: Comment calls this a "sticky USD price cache" but `tokenPrices` is correctly excluded from `partialize` (L388-392) — so it's sticky only within the tab session, lost on reload. Either reword ("session-scoped USD price cache") or persist with a TTL-prune on rehydrate. Current behavior is probably what you want; just align the doc.

src/features/tokens/useTokenPrice.tsx
Xaroz: Switched to 15min in 5c46def0.

src/features/tokens/useTokenPrice.tsx
Xaroz: Added response body to the thrown error and `encodeURIComponent` per id in 5c46def0. Holding off on the env override / pro-key support for now.

src/features/tokens/useTokenPrice.tsx
Xaroz: Done in 5c46def0 — the throw now includes `await res.text().catch(() => '')`.

src/features/tokens/useTokenPrice.tsx
Xaroz: Fixed in 5c46def0. Chunks now run sequentially with a 200ms inter-chunk delay, batch size dropped to 100 (you were right — the previous 250-comment referenced the wrong endpoint constraint; URL length on `/simple/price` is the real limit).

src/features/tokens/useTokenPrice.tsx
Xaroz: Tightened in 5c46def0 — `useTokenUsdValue` now reads via a parameterized selector `useStore((s) => s.tokenPrices[id]?.usd)`. The broader `useTokenPricesByIds` shape change is a larger refactor; deferred.

src/features/tokens/useTokenPrice.tsx
Xaroz: Added a TODO comment above the debounce in 5c46def0 flagging this for a manual throttled-network test + leading-edge debounce revisit if it surfaces.

src/features/tokens/useTokenPrice.tsx
Xaroz: Dropped in 5c46def0 — queryFn returns `null` now. Consumers read from the Zustand store; TQ is purely a debounced trigger + retry mechanism.

src/features/tokens/useTokenPrice.tsx
Xaroz: Good catch — fixed in 5c46def0. Added per-id `failedAt` to the cache entry, with a 30s short-circuit in the `idsToFetch` filter. Merge + fail-mark are now folded into a single `mergeTokenPrices(succeededIds, fetched, failedIds)` so each fetch cycle does one set() call.

src/features/tokens/useTokenPrice.tsx
Xaroz: Dropped `staleTime` + `refetchOn*` in 5c46def0 — you're right, they were dead options since the queryKey shifts on every successful merge. Added a comment noting Zustand is the cache of record. `failedAt` short-circuit (per the L105 thread) covers the retry-exhaustion stuck case.

src/features/swap/SwapForm.tsx
Xaroz: Fixed in 5c46def0 — gate at the call site (`amountUsd >= 0.01 ? formatUsd(amountUsd) : '$0.00'`), mirroring the bridge pattern.

src/features/swap/SwapForm.tsx
Xaroz: Fixed in 5c46def0. Same call-site gate. For the Min ambiguity I dropped the `Min: …` chip from this row entirely — slippage already shows in the SlippagePanel above and the review summary. Side benefit: no more `·` separator / a11y concern.

src/features/store.ts
Xaroz: Reworded to "session-scoped" in 5c46def0. Keeping the in-memory-only behavior as you suggested.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,125 @@
...
-    refetchInterval: PRICE_STALE_TIME,
-    refetchOnWindowFocus: false,
-    refetchOnMount: false,
+  const idsToFetch = useMemo(() => {
+    const now = Date.now();
+    return debouncedIds.filter((id) => {
+      const entry = tokenPrices[id];
+      if (!entry) return true;
+      if (entry.fetchedAt && now - entry.fetchedAt < PRICE_CACHE_MS) return false;
+      if (entry.failedAt && now - entry.failedAt < FAILED_BACKOFF_MS) return false;
```
paulbalaji: **Cache expiry isn't time-driven (likely-unintended regression).** `now` here only re-evaluates when `debouncedIds` or `tokenPrices` change. After the first successful fetch with a stable id set, `idsToFetch` settles to `[]`, the query disables (`enabled: idsToFetch.length > 0`), and nothing re-triggers — so the 15m `fetchedAt` TTL never actually drives a refresh in a long-open tab, and the 30s `failedAt` backoff only retries if some *unrelated* state happens to change. The old bridge hook had `refetchInterval: PRICE_STALE_TIME` (5m) for exactly this case, and it was dropped.

For a financial USD readout, prices going stale-forever is the most consequential item here. Consider a coarse `setInterval` tick folded into this memo's deps (or an interval that re-seeds `idsToFetch` from the full `ids`). Can you confirm whether dropping automatic refresh was intentional vs. just the TTL bump (5m→15m)?

src/features/swap/SwapForm.tsx
```diff
@@ -530,7 +533,7 @@ function OriginTokenCard({
           />
         </div>
         <div className="transfer-balance mt-1 flex items-center justify-between text-xs leading-[18px] text-gray-450 dark:text-foreground-secondary">
-          <span>$0.00</span>
+          <span>{amountUsd >= 0.01 ? formatUsd(amountUsd) : '$0.00'}</span>
```
paulbalaji: `amountUsd >= 0.01 ? formatUsd(amountUsd) : '$0.00'` collapses several distinct states — loading, missing `coinGeckoId`, empty amount, true-zero, **and a known positive sub-cent value** — all into `$0.00` (same on line 608). But `swap/tokens/TokenList.tsx:236` renders a known sub-cent value as `<$0.01` (via the same `formatUsd`, re-exported from `utils/amount.ts`), so within the swap tab the same token reads `$0.00` on the card and `<$0.01` in the list.

Suggest having `useTokenUsdValue` return a nullable USD and only rendering `$0.00` for exact zero, letting a known positive sub-cent flow through `formatUsd` → `<$0.01`. That also centralizes the gate instead of duplicating the `>= 0.01` literal across both cards.

src/features/swap/tokens/useTokenPrice.tsx
```diff
@@ -1,6 +1,40 @@
...
+  return useTokenPricesByIds(ids);
+}
+
+// USD value of a token `amount` string. Reads the single price via a
+// parameterized selector so consumers don't re-render on unrelated
+// tokenPrices mutations.
+export function useTokenUsdValue(token: UiToken | undefined, amount: string): number {
+  // Side-effect mount: keeps the catalogue-wide fetch hot for deep-linked
+  // URLs where cards render before the picker opens.
+  useTokenPrices();
```
paulbalaji: Calling `useTokenPrices()` purely for its mount side-effect (return discarded) defeats the per-id-selector intent and multiplies hook work: `OriginTokenCard`, `DestinationTokenCard`, and (when the picker is open) `TokenList` each mount a full `useTokenPricesByIds` → 3 concurrent `useDebounce` + filter + query instances. They dedupe network-wise (identical `queryKey`), so no duplicate fetches — but it's a "call a hook, throw away the result" smell plus 3× the debounce/memo work.

Cleaner: hoist one `useTokenPrices()` to the swap parent (as the bridge does in `TransferTokenForm`) and keep `useTokenUsdValue` a pure per-id reader.

src/features/swap/tokens/useTokenPrice.tsx
```diff
@@ -1,6 +1,40 @@
...
+  const id = token?.coinGeckoId;
+  const price = useStore((s) => (id ? s.tokenPrices[id]?.usd : undefined));
+  return useMemo(() => {
+    // Strip grouping separators — parseFloat("1,234.56") returns 1 otherwise.
+    const a = parseFloat(String(amount ?? '').replace(/,/g, ''));
+    if (!price || isNaN(a)) return 0;
+    return a * price;
```
paulbalaji, line 38: Hardening: `a * price` can overflow to `Infinity` for an extreme amount. `isNaN(a)` guards NaN but not `Infinity`, so the card renders `$∞` (`Infinity >= 0.01` → `formatUsd(Infinity)` → `(Infinity).toLocaleString()`). Cheap guard: `if (!price || !Number.isFinite(a)) return 0;` and/or guard the product. Requires absurd inputs — low severity.

paulbalaji (commented):
## Review Summary

Solid PR, and well-iterated through the first round. Unifying bridge + swap behind a single store-backed CoinGecko cache is the right move, and switching `getUsdValue` from lowercase-symbol to `coinGeckoId` removes a latent cross-chain collision.

**Prior review (xeno097 → addressed in `5c46def0`):** verified resolved — response-body in thrown error, `encodeURIComponent` per id, 15m TTL, sequential chunking + 200ms delay (batch 100), `failedAt` write, dropped dead `staleTime`/`refetchOn*`, `$0.00` gate, `Min:`/`·` ambiguity, store comment wording. The broad `tokenPrices` subscription on wider consumers and the bootstrap debounce-reset remain acknowledged-but-open (author added a TODO); not blocking.

**Cleared, no action:** CSP is fine — `next.config.js` uses `connect-src *` on this and the base branch, so the CoinGecko fetches are allowed (no regression). The `await res.text()` error body only reaches `logger.warn`/Sentry, never a user toast. `encodeURIComponent` is sufficient against URL injection.

**Leaving as a comment, not an approval** (this flow doesn't APPROVE). The offsets/logic are sound; the one item I'd want confirmed before merge is the no-time-driven-refresh regression below — the rest are quality.

- `src/features/balances/useFeePrices.ts:8`: comment `// 5 min, matches PRICE_STALE_TIME` is now stale — `PRICE_STALE_TIME` was deleted in this PR and the shared cache TTL is 15m. `useFeePrices` also still fetches outside the new shared cache, so the "single shared cache" goal isn't fully unified (fee prices skip the negative-cache/backoff). Minor; worth a follow-up.

src/features/tokens/useTokenPrice.tsx
Xaroz: Good catch — unintentional regression from dropping the TQ refetch options last round (they were vestigial against shifting queryKeys, but were also the only thing that drove time-based refresh in the stable-cache case).

Fixed in 18640905. Reshaped so TQ's `refetchInterval: PRICE_CACHE_MS / 2` owns the cadence: queryFn is extracted as a plain `fetchTokenPrices(ids, cache, merge)`, queryKey now keys on `debouncedIds` (shifts only on catalogue growth), and the staleness filter lives inside queryFn so each tick re-evaluates against the latest `Date.now()` + store snapshot. queryFn returns null when nothing is stale (no network). No useEffect, no `getState` — the queryFn closure carries the reactive `tokenPrices` from the hook's useStore selector.

src/features/swap/SwapForm.tsx
Xaroz: Fixed in 18640905. `useTokenUsdValue` now returns `number | null` (null = no info, 0 = legit zero, positive = real value including sub-cent), and the call sites use `!amountUsd ? '$0.00' : formatUsd(amountUsd)`. So a known positive sub-cent value flows through `formatUsd` → `<$0.01` and matches TokenList's rendering on the same token within the swap tab.

src/features/swap/tokens/useTokenPrice.tsx
Xaroz: Fixed in 18640905. Hoisted the `useTokenPrices()` mount to `SwapFormContent` so the catalogue-wide fetch fires once for the whole form. `useTokenUsdValue` is now a pure per-id reader (just the parameterized selector). Cards no longer spin up their own `useDebounce` + `useQuery` observers.

src/features/swap/tokens/useTokenPrice.tsx
Xaroz: Fixed in 18640905 — `useTokenUsdValue` now uses `Number.isFinite(a)` instead of `isNaN(a)`, which catches both NaN and Infinity. Folded into the same diff as the nullable-return rework.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -1,52 +1,137 @@
...
-    staleTime: PRICE_STALE_TIME,
-    refetchInterval: PRICE_STALE_TIME,
+  const { isLoading } = useQuery({
+    queryKey: ['tokenPrices', debouncedIds],
+    queryFn: () => fetchTokenPrices(debouncedIds, tokenPrices, mergeTokenPrices),
+    enabled: debouncedIds.length > 0,
+    refetchInterval: PRICE_CACHE_MS / 2,
     refetchOnWindowFocus: false,
-    refetchOnMount: false,
   });
```
paulbalaji, line 130: **Partial/total-failure recovery regressed to ~7.5min, and `FAILED_BACKOFF_MS` is now dead tuning.** (net-new from the round-2 reshape — flagging since it touches the same financial-staleness concern from my last review.)

The filter moved out of a reactive `useMemo`/`queryKey` and into the async `queryFn`, and `queryKey` is now `['tokenPrices', debouncedIds]` — which shifts **only on catalogue growth**, not on cache merges. So:

1. A fetch partially fails (some chunks 429): survivors merge, the rest get `failedAt` set. `requestedIds.length > 0` → `queryFn` returns `null` (success), **no throw**. The store mutates and re-renders, but `queryKey` is unchanged → **no follow-up fetch fires**.
2. Total failure: `throw` → TQ retries ~1s/2s/4s, but each retry re-runs `queryFn`, and the `failedAt < 30s` branch (line 85) filters every id out → `idsToFetch` empty → returns `null`. So the retries are absorbed as no-ops without re-hitting CoinGecko.
3. The only thing that re-invokes `queryFn` is `refetchInterval: PRICE_CACHE_MS / 2` = **7.5 min**.

Net: any id that hits a transient CoinGecko failure sits at `$0.00` for up to 7.5 min, and `FAILED_BACKOFF_MS = 30_000` never actually drives a retry (nothing re-evaluates the filter at 30s cadence). The PR body still advertises "next render's smaller `idsToFetch` fires a follow-up automatically" — that was true in round 1 but the queryKey decoupling broke it.

Not stuck/hot-looping (it does recover at the next tick), so not a hard blocker — but for a USD readout this undercuts the round-2 fix's intent. Options: invalidate/refetch this query after a partial-failure merge, or add a short `refetchInterval` (or a separate timer) that actually honors `FAILED_BACKOFF_MS`. Can you confirm the 7.5-min worst-case recovery is acceptable vs. intended?

paulbalaji (commented):
## Review Summary

Reviewed the current HEAD (`18640905`) with a focus on the round-2 reshape. The architecture is sound — single store-backed CoinGecko cache shared by bridge + swap, per-id delta fetching, `getUsdValue` keyed on `coinGeckoId`, sub-cent values flowing through `formatUsd` → `<$0.01` to match `TokenList`. Two review rounds of feedback are addressed.

**Round 1 (xeno097 → `5c46def0`):** previously verified resolved.

**Round 2 (`18640905`) — verified each author reply:**
- ✅ Nullable `useTokenUsdValue` return (`number | null`), call sites gate `!amountUsd ? '$0.00' : formatUsd(...)` — positive sub-cent now renders `<$0.01`, consistent with `TokenList`. (Minor: at the two call sites `null` and `0` render identically, so the nullable distinction isn't exercised there — fine.)
- ✅ `useTokenPrices()` hoisted to `SwapFormContent`; cards are per-id readers.
- ✅ `Number.isFinite(a)` now catches both NaN and Infinity.
- ⚠️ Time-driven refresh — the TTL-refresh case is fixed (`refetchInterval` now drives the 15m TTL), **but the reshape introduced a partial/total-failure recovery regression** — see inline comment.

**Acknowledged-deferred (confirmed still as-noted, not blocking):**
- Whole-object `useStore((s) => s.tokenPrices)` subscription in `useTokenPricesByIds` re-renders its host on every merge. The round-2 hoist added a second host at the top of `SwapFormContent`; since the cards aren't memoized, the per-id-selector win is partially defeated for the swap cards. Modest impact (merges are infrequent); matches the existing deferral.
- Bootstrap `useDebounce(ids, 300)` timer-reset on slow networks — TODO comment present at `useTokenPrice.tsx:108`.

**Not approving via this flow** (this review path only posts COMMENT). The one item I'd want confirmed before merge is the failure-recovery regression below; everything else is quality.

- `src/features/balances/useFeePrices.ts:8`: comment `// 5 min, matches PRICE_STALE_TIME` is still stale — `PRICE_STALE_TIME` was deleted in this PR. `useFeePrices` also still fetches outside the shared cache. Minor follow-up (carried over from prior review).

paulbalaji (approved):
Approving — two rounds of feedback addressed and verified at `18640905`. The one open item (partial/total-failure recovery now waits for the 7.5min `refetchInterval` tick rather than self-healing, leaving `FAILED_BACKOFF_MS` inert — see inline thread) is non-blocking: it auto-recovers and doesn't corrupt state. Worth a follow-up to restore the faster retry, but not a merge blocker. Nice work iterating on this.

src/features/tokens/useTokenPrice.tsx
Xaroz: Fixed in 57f6d34f. Restored self-healing via TQ-native retry: queryFn now throws on ANY failed chunk (was: only total failure), and useQuery gets `retry: 1, retryDelay: FAILED_BACKOFF_MS`. So a partial failure triggers a single retry after the rate-limit window passes, then falls back to the normal refetchInterval cadence.

Also bumped `FAILED_BACKOFF_MS` from 30s → **90s** to clear CoinGecko's rolling 60s rate-limit window with breathing room — 30s retries land mostly inside the same window and fail again. Same value drives both the filter (skip failed IDs within window) and TQ's retry delay, keeping semantics aligned.

Net behavior in a sustained outage: ~2 requests per 7.5min cycle (initial + one retry). Common transient case recovers in 90s instead of 7.5min.

PR #1128, opened 2026-06-01
feat(swap): classic ERC20 approvals + revoke handling

Xaroz:
## Summary

Replaces the swap form's Permit2 two-step approval (`ERC20.approve(Permit2)` + `Permit2.approve(UR)`) with a classic single-step `ERC20.approve(UR, amount)`. Mirrors the warp-bridge UX. Adds USDT-style revoke-before-approve handling.

## What changed

- **Renamed `permit2.ts` → `approval.ts`** with a slimmer surface: just `useApprovalStatus` + `ApprovalPhase` enum (`Idle / Native / NeedsRevoke / NeedsApprove / Ready`). The approval read stays reactive for the review-modal preview.
- **Writes moved inline into `useSwap.execute`** as plain async — no more separate `useApproveErc20` / `useRevokeErc20` hooks. `useSwap` now takes `spender / approvalAmount / isNative` and owns the full revoke → approve → swap sequence.
- **SDK's `EvmTokenAdapter`** does the heavy lifting — `isApproveRequired`, `isRevokeApprovalRequired`, `populateApproveTx`. Same code path for EVM and Tron via `MultiProtocolProvider` dispatch. The adapter handles base58 → hex internally for Tron addresses.
- **USDT revoke** mirrors `WarpCore.ts:548-573`: when prior allowance is non-zero and a new approval is needed, emit `approve(0)` before `approve(amount)`.
- **`SwapForm`** loses ~250 lines: no more wagmi `useReadContract` / `useWriteContract` boilerplate, no more two-phase approval branches. Just passes the new args to `swap.execute()`.

## Pairs with

Engine PR that defaults `usePermit2: false` for the WETH-unwrap site. The UR's `payOrPermit2Transfer` helper falls back from `ERC20.transferFrom` to Permit2 automatically for `V3_SWAP_EXACT_IN` / `BRIDGE_TOKEN` regardless of the flag, so no breakage during rollout — this UI works against both old and new engine builds.

## Tradeoffs

- Approval is **amount-based** (not max-uint), matching WarpCore. Each swap that exceeds remaining allowance re-prompts approve — same UX as the bridge.
- Permit2 isn't removed from the engine: clients can still opt in (`usePermit2: true`) for the gasless 2nd-tx ergonomics. This UI just doesn't pass it.


src/features/swap/useSwap.ts
```diff
@@ -61,6 +67,39 @@ export function useSwap() {
...
+              recipient: spender,
+            });
+            const { confirm } = await fns.sendTransaction({
+              tx: {
+                type: ProviderType.EthersV5,
+                transaction: { to: populated.to!, data: populated.data!, value: '0' },
+                category: 'transfer',
+              } as Parameters<typeof fns.sendTransaction>[0]['tx'],
+              chainName: srcChainName,
+            });
```
paulbalaji: **Blocker:** this hardcodes `ProviderType.EthersV5`, but the swap tx at line 110 branches `protocol === ProtocolType.Tron ? ProviderType.Tron : ProviderType.EthersV5`. The Tron wallet handler only accepts `ProviderType.Tron` and otherwise throws `Invalid Tron provider type EthersV5` — so non-native Tron swaps that need an approve/revoke fail here before reaching the swap. Tron is live (`/v1/chains`: `canSwap/canExecute=true`, with non-native swappable tokens).

Reuse the same branch as line 110:
```ts
const txType = protocol === ProtocolType.Tron ? ProviderType.Tron : ProviderType.EthersV5;
// ...
tx: { type: txType, transaction: { to: populated.to!, data: populated.data!, value: '0' }, category: 'transfer' }
```
(`protocol` is already in scope from line 53.)

src/features/swap/useSwap.ts
```diff
@@ -61,6 +67,39 @@ export function useSwap() {
...
+            updateSwapStatus(swapIndex, SwapStatus.ConfirmingApprove);
+            await confirm();
```
paulbalaji: **Medium:** `doApprove` awaits `confirm()` but never checks the receipt, unlike the swap tx (`isReverted(receipt)` at line 127). A reverted approval (e.g. a USDT-style approve-without-revoke edge) silently falls through into a doomed swap attempt with a less legible failure. Capture the receipt and run the same `isReverted` check before continuing.

src/features/swap/useSwap.ts
```diff
@@ -61,6 +67,39 @@ export function useSwap() {
           }
         }
 
+        // Approve / revoke before the swap tx. Mirrors WarpCore's pattern:
+        // bump non-zero existing allowance to zero first (USDT case), then
+        // approve the new amount.
+        if (args.spender && args.approvalAmount != null && !args.isNative) {
```
paulbalaji, line 78: **Low:** the old `permit2.ts` threw if `universalRouter === zeroAddress` before approving. This gate treats `0x0` as truthy, so a misconfigured/zero UR from the engine would pass and broadcast `approve(0x0, amount)`. Not attacker-exploitable (spender is a zod-validated address from the trusted engine response, and approving 0x0 is inert), but per CLAUDE.md's "fail loudly on broken invariants" it's worth re-adding an `isZeroishAddress(spender)` guard here.

src/features/swap/approval.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  const enabled = !!chainName && !!token && !!owner && !!spender && !isNative && amount != null;
+
+  const { data } = useQuery({
+    queryKey: [
+      'erc20-allowance',
+      chainName,
+      token?.toLowerCase(),
+      owner?.toLowerCase(),
+      spender?.toLowerCase(),
```
paulbalaji, line 58: Nit: `owner` is base58 for Tron (case-sensitive), so `.toLowerCase()` here is inconsistent with the repo's "only lowercase EVM addresses" rule and the sibling `tokens/approval.ts` hook (which keys on raw `owner`). Harmless in practice — it's only the cache key, and the raw `owner` is what's passed to the adapter on line 65 — but keying on the raw value avoids the inconsistency.

paulbalaji (commented):
## Review Summary

Clean, net-positive refactor — drops 365 lines of Permit2 dual-allowance logic for classic `ERC20.approve(UR, amount)` via the SDK's `EvmTokenAdapter`, and it's a genuine security improvement (exact-amount approvals instead of lingering infinite allowances, plus on-chain re-validation at execute time rather than trusting stale UI state). Revoke→approve sequencing, phase mapping, native short-circuit, and EVM tx path all verified correct.

**One blocker prevents approval** (Tron). The change was reviewed in two independent passes which agree on all findings below; `pnpm typecheck` and the vitest suite (252 tests) pass.

_No prior reviews, inline threads, or author replies existed on this PR — nothing to reconcile._

### Blocker
- **Tron approvals send the wrong tx type** (`useSwap.ts:90`). The approval/revoke tx hardcodes `ProviderType.EthersV5`, but the swap tx 20 lines below correctly branches to `ProviderType.Tron` for Tron. The Tron wallet handler rejects anything but `ProviderType.Tron` (throws `Invalid Tron provider type EthersV5`), so any **non-native Tron swap needing an approve/revoke fails before the swap**. Tron is live (`/v1/chains` has `tron canSwap/canExecute=true` with non-native swappable tokens). Note: the deleted `permit2.ts` had the same latent issue, so this is pre-existing — but this PR consolidates the write path and claims EVM/Tron parity, so it's the place to fix it.

### Medium
- **Approval receipt revert is not checked** (`useSwap.ts:97`) — see inline.

### Low
- **Lost zero-address guard on the spender/UR** (`useSwap.ts:73`) — see inline.

### Cleanup (outside the diff)
- `src/consts/config.ts:27,57,92` + `.env.example`: `permit2ExpirationSeconds` is now orphaned — its only consumers were in the deleted `permit2.ts`. Classic approve has no expiration concept; remove the const, the `Config` field, the frozen entry, and the env line.
- `src/features/api/types.ts:52-58`: the `permit2` field comment still documents the removed gating ("surfaces as 'Loading Permit2…'"). The field is no longer read by the UI — trim the comment to "returned by engine, currently unused" or drop the field.

### Verified correct (no action)
Revoke→approve mirrors WarpCore exactly; `EvmTokenAdapter` converts base58→hex and dispatches to the Tron provider internally (read path is fine — only the write path above is wrong); `confirm()` does await the receipt on EVM; spender provenance is a zod-validated address from the trusted engine `/v1/chains`; `value:'0'` carries no native-leak risk; no CSP/XSS/new-dep concerns.

Verdict: not approving while the Tron blocker stands. Everything else is medium/low/cleanup. Once the approval tx type is branched like the swap tx, this is good to merge after the two cleanup items.

troykessler (approved):

src/features/swap/useSwap.ts
Xaroz: Good catch — fixed in edcc3ff6. Hoisted `txType` out of the swap-tx block so both the approve and the swap use the same protocol-aware ternary. The old permit2.ts had this same EthersV5 hardcode but only ran for Permit2-only flows where it slipped through; under classic ERC20 approval it actually fires for non-native Tron, so this was untested-broken before.

src/features/swap/useSwap.ts
Xaroz: Fixed in edcc3ff6. `doApprove` now captures the receipt and runs `isReverted()`, throwing a clean error before proceeding. Mirrors the swap-tx handling so a reverted approval surfaces a legible message instead of cascading into a cryptic swap revert.

src/features/swap/useSwap.ts
Xaroz: Fixed in edcc3ff6. Added `isZeroishAddress(spender)` guard from `@hyperlane-xyz/utils` before the allowance read — throws with the chain name in the message. Restores the parity with the old permit2.ts behavior the rewrite accidentally dropped.

src/features/swap/approval.ts
Xaroz: Fixed in edcc3ff6 — dropped `.toLowerCase()` on `owner` in the queryKey. Keys the raw value now; only the EVM-shape fields (token, spender) get lowercased.

src/features/swap/useSwap.ts
```diff
@@ -62,14 +70,53 @@ export function useSwap() {
...
+              tx: {
+                type: txType,
+                transaction: { to: populated.to!, data: populated.data!, value: '0' },
+                category: 'transfer',
+              } as Parameters<typeof fns.sendTransaction>[0]['tx'],
+              chainName: srcChainName,
+            });
+            updateSwapStatus(swapIndex, SwapStatus.ConfirmingApprove);
+            const receipt = await confirm();
+            if (isReverted(receipt)) {
```
paulbalaji, line 106: Non-blocking follow-up: `isReverted` (useSwap.ts:189) returns `false` for any receipt that isn't `Viem`/`EthersV5`, so a **Tron** receipt never trips this guard — the reverted-approve protection added here (and the swap-path check at :136) is effectively EVM-only. Pre-existing for the swap path, but since this PR's stated goal is EVM/Tron parity on the write path, worth extending `isReverted` to handle `ProviderType.Tron` receipts (or explicitly noting the gap). On Tron you'd currently rely on `confirm()` itself throwing.

paulbalaji (approved):
## Review — Approve ✅

Reviewed via independent agent passes (correctness/sequencing, dead-code sweep, security + build), reconciled against the existing review history.

### Reconciling prior reviews
- **@paulbalaji**'s 4 inline findings — all **verified resolved** in `edcc3ff6`; author replies accurate:
  - Blocker (Tron tx type): `txType` hoisted and shared by both approve and swap paths — `useSwap.ts:62`.
  - Medium (approve revert): `doApprove` captures the receipt and runs `isReverted` before continuing — `useSwap.ts:103-107`.
  - Low (zero-address): `isZeroishAddress(spender)` guard restored — `useSwap.ts:78-80`.
  - Nit (queryKey): `owner` kept raw (case-sensitive Tron base58); only `token`/`spender` lowercased — `approval.ts`.
- **@troykessler** approved.

### Verified correct
- Revoke→approve→swap sequencing mirrors WarpCore exactly; USDT case handled (revoke to 0, then approve amount). Each step awaits its own `confirm()` — no race.
- UI phase mapping (`approval.ts`) and execute-time decision use identical inputs and agree; `execute` re-reads allowance live before broadcast, so a stale UI read can't skip/over-approve.
- Exact-amount approval (no lingering infinite allowance) — genuine security upside. `spender` is a zod-validated `Address` from the trusted engine `/v1/chains`; approve tx sends `value:'0'`, no native leak.
- `pnpm typecheck` passes; 252 tests pass. No new deps / CSP / external scripts.

### Still open — non-blocking, recommend before merge
Both are outside the diff (noted here rather than inline):
1. **Orphaned `permit2ExpirationSeconds`** — only consumers were in the deleted `permit2.ts`. Remove: `config.ts:27-29` (const), `:57` (Config field), `:92` (frozen entry), and `.env.example` (`NEXT_PUBLIC_PERMIT2_EXPIRATION_SECONDS`).
2. **Stale comment** at `api/types.ts:52-57` — still documents removed gating ("surfaces as 'Loading Permit2…'"). The field at `:58` stays (engine response); just trim the comment.

### New — low, follow-up
- See inline: Tron receipts bypass the new approve-revert guard.

Net: clean, well-scoped refactor with a real security improvement. Approving — the two cleanup items and the Tron `isReverted` gap are worth a quick follow-up but don't block.

src/features/swap/useSwap.ts
Xaroz: Fixed in f90516eb — added `ProviderType.Tron` to the early-return allowlists in both `isReverted` and `parseReceipt`. Tron's `TypedTransactionReceipt` wraps an ethers v5 `TransactionReceipt` (same `status` shape, same `logs` shape per the SDK's `ProviderType.ts:492-495`), so the existing status/log-parsing logic works as-is once the early return doesn't kick out Tron.

PR #1135, opened 2026-06-03
feat: show 3 message ids and show success toast

troykessler:
show the 3 message ids and a success toast just like the standard warp transfer

### DIFF
--- a/src/features/swap/SwapDetailsModal.tsx
+++ b/src/features/swap/SwapDetailsModal.tsx
@@ -10,6 +10,7 @@ import {
 } from '@hyperlane-xyz/widgets';
 import Image from 'next/image';
 import { useEffect, useMemo, useRef, useState } from 'react';
+import { toast } from 'react-toastify';
 
 import { ChainLogo } from '../../components/icons/ChainLogo';
 import { TokenChainIcon } from '../../components/icons/TokenChainIcon';
@@ -88,7 +89,7 @@ function SwapDetailsModalInner({
     recipient,
     originTxHash,
     destinationTxHash,
-    msgId,
+    msgIds,
     timestamp,
   } = swap;
 
@@ -103,13 +104,21 @@ function SwapDetailsModalInner({
   const isDelivered = status === SwapStatus.ConfirmedDestination;
   const isFinal = FinalSwapStatuses.includes(status);
 
-  // Delivery polling. Same-chain swaps have no msgId — skip polling.
-  const delivery = useMessageDeliveryStatus(msgId, isOpen && !isFinal, multiProvider);
+  // Poll the warp message for timeline/delivery status (fallback to first).
+  // Same-chain swaps have no msgIds — skip polling.
+  const pollingMsgId = (msgIds?.find((m) => m.label === 'warp') ?? msgIds?.[0])?.msgId;
+  const delivery = useMessageDeliveryStatus(pollingMsgId, isOpen && !isFinal, multiProvider);
 
   const hasUpdatedDelivery = useRef(false);
   useEffect(() => {
     hasUpdatedDelivery.current = false;
-  }, [msgId]);
+  }, [pollingMsgId]);
+
+  const LABEL_NAMES: Record<string, string> = {
+    warp: 'Warp Message ID',
+    commit: 'Commit Message ID',
+    reveal: 'Reveal Message ID',
+  };
 
   useEffect(() => {
     if (
@@ -122,6 +131,7 @@ function SwapDetailsModalInner({
       updateSwapStatus(swapIndex, SwapStatus.ConfirmedDestination, {
         destinationTxHash: delivery.destinationTxHash,
       });
+      toast.success('Swap complete! Funds have arrived.');
     }
   }, [delivery.isDelivered, delivery.destinationTxHash, swapIndex, status, updateSwapStatus]);
 
@@ -146,8 +156,8 @@ function SwapDetailsModalInner({
       ? MessageStatus.Failing
       : MessageStatus.Pending;
 
-  // Show timeline only for cross-chain swaps (msgId present).
-  const showTimeline = isSent && !isFailed && !!originTxHash && !!msgId;
+  // Show timeline only for cross-chain swaps (msgIds present).
+  const showTimeline = isSent && !isFailed && !!originTxHash && !!pollingMsgId;
 
   const date = useMemo(
     () => (timestamp ? formatTimestamp(timestamp) : formatTimestamp(Date.now())),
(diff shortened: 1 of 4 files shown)


src/features/swap/useSwap.ts
```diff
@@ -149,25 +150,82 @@ function isReverted(receipt: TypedTransactionReceipt): boolean {
...
-        logs?: Array<{ topics?: readonly string[] }>;
-        blockNumber?: bigint | number;
+  const rawReceipt = receipt.receipt as {
+    logs?: Array<{ address?: string; topics?: readonly string[] }>;
+    blockNumber?: bigint | number;
+  };
+  const logs = rawReceipt.logs ?? [];
+  const blockNumber = rawReceipt.blockNumber;
+
+  // The Mailbox emits Dispatch then DispatchId back-to-back for each dispatch() call.
```
paulbalaji: **[Medium — reuse]** This hand-rolled `Dispatch`/`DispatchId` pairing reinvents an SDK helper the repo already uses. `HyperlaneCore.getDispatchedMessages(rawReceipt)` parses each `Dispatch` log via the Mailbox interface and computes the id directly from the message payload (no `DispatchId` pairing/queue needed), returning `parsed.sender` and `parsed.destination`. The transfer flow already goes through this path (`src/features/transfer/utils.ts:126` → `core.extractMessageIds`).

Using the static `getDispatchedMessages(rawReceipt)` here (it works on raw `.logs` for both Viem and EthersV5, unlike `extractMessageIds` which throws on Viem) would drop the `DISPATCH_SIG`/`DISPATCH_ID_SIG` constants, the topic-slicing casts, and the queue (~50 lines), mapping `{ id, parsed }` → `{ msgId: id, sender: parsed.sender }`. Aligns with CLAUDE.md tips #2/#3 (reuse SDK + existing patterns). Not gating.

src/features/swap/useSwap.ts
```diff
@@ -149,25 +150,82 @@ function isReverted(receipt: TypedTransactionReceipt): boolean {
   return false;
 }
 
+interface ParsedMessage {
+  msgId: `0x${string}`;
+  sender: `0x${string}`;
+  destDomain: number;
```
paulbalaji: **[Low]** `destDomain` is parsed and threaded through `ParsedMessage` + the pending queue but never read by `labelMessages` or anything downstream (only `sender` is used). Drop it to reduce noise, or add a comment if it's intended for forward-compat.

src/features/swap/useSwap.ts
```diff
@@ -149,25 +150,82 @@ function isReverted(receipt: TypedTransactionReceipt): boolean {
...
+      .map((s) => s.router.toLowerCase()),
+  );
+
+  let nonWarpCount = 0;
+  return messages.map((msg) => {
+    if (bridgeRouters.has(msg.sender.toLowerCase())) {
+      return { msgId: msg.msgId, label: 'warp' as const };
+    }
+    // Commit precedes reveal in the CCS protocol.
+    const label = nonWarpCount === 0 ? ('commit' as const) : ('reveal' as const);
```
paulbalaji, line 199: **[Low — brittleness]** The commit/reveal split assumes exactly the current topology (≤1 non-warp pair, commit before reveal). If a future engine emits only a commit on origin, or a 3rd non-warp message, this silently mislabels (lone msg → `commit` with no `reveal`; 3rd → `reveal`). Consider a `logger.warn` when `nonWarpCount > 2` so a topology change surfaces loudly rather than mislabeling — consistent with the repo's "fail loudly on broken invariants" philosophy.

src/features/swap/SwapDetailsModal.tsx
```diff
@@ -122,6 +131,7 @@ function SwapDetailsModalInner({
       updateSwapStatus(swapIndex, SwapStatus.ConfirmedDestination, {
         destinationTxHash: delivery.destinationTxHash,
       });
+      toast.success('Swap complete! Funds have arrived.');
```
paulbalaji, line 141: **[Low — confirm intended]** Delivery (and thus this toast) is only detected while the modal is open: `useMessageDeliveryStatus(pollingMsgId, isOpen && !isFinal, ...)` stops polling on close. So if the user closes the modal mid-bridge, the swap stays non-final in history and the success toast never fires until they reopen it. The standard transfer detects delivery via a background balance watcher. This is a new feature (not a regression), but worth confirming the UX is intended.

src/features/swap/SwapDetailsModal.tsx
```diff
@@ -103,13 +104,21 @@ function SwapDetailsModalInner({
...
+  const pollingMsgId = (msgIds?.find((m) => m.label === 'warp') ?? msgIds?.[0])?.msgId;
+  const delivery = useMessageDeliveryStatus(pollingMsgId, isOpen && !isFinal, multiProvider);
 
   const hasUpdatedDelivery = useRef(false);
   useEffect(() => {
     hasUpdatedDelivery.current = false;
-  }, [msgId]);
+  }, [pollingMsgId]);
+
+  const LABEL_NAMES: Record<string, string> = {
```
paulbalaji: **[Nit]** `LABEL_NAMES` is a static constant recreated every render. Hoist to module scope alongside `STATUS_DESCRIPTION` / `DEFAULT_TIMINGS` where the other constant maps already live.

src/features/swap/SwapDetailsModal.tsx
```diff
@@ -190,7 +200,6 @@ function SwapDetailsModalInner({
     };
   }, [multiProvider, originChain, destChain, originTxHash, destinationTxHash, sender, recipient]);
 
-  const explorerLink = msgId ? getHypExplorerLink(multiProvider, originChain, msgId) : undefined;
 
   return (
```
paulbalaji, line 210: **[Nit]** Removing `explorerLink` left a double blank line above this `return` (lines 202–203). Prettier/lint will likely flag it.

paulbalaji (approved):
## Review Summary

**Approving with comments.** The feature is functionally sound and well-scoped. I traced the CCS flow end-to-end against the Mailbox / InterchainAccountRouter contracts and the route engine — the message parsing and labeling are **correct** for the current EVM CCS topology:

- `Mailbox.dispatch()` emits `Dispatch` immediately followed by `DispatchId`, so the per-mailbox FIFO pairing in `parseReceipt` holds (queue depth never exceeds 1 in practice).
- The warp leg's `Dispatch.sender` equals the bridge step's `router`, so the `warp` label is correct; commit/reveal are both dispatched by the ICA router **commit-first**, so the order-based fallback is correct.
- Toast double-fire is properly guarded (`hasUpdatedDelivery` ref + `status !== ConfirmedDestination` + polling disabled once final).
- No migration concern: swaps are in-memory only (`partialize` persists only `chainMetadataOverrides` + `transfers`), and `SwapHistoryItem` never had a released `msgId: string`. ✅

No blockers or high-severity issues. Inline comments are one medium (SDK reuse) plus low/nit hardening suggestions — none gating.

Typecheck/lint were not run (worktree had no `node_modules`) — worth a local confirm, mainly for the double blank line below.

PR #1138, opened 2026-06-05
feat(swap): UI polish — modal label, output USD + price impact, fee % + USD, ICA dest tx, default tokens

Xaroz:
## Summary

Round of UI polish on the swap form.

## Changes

- **Modal copy**: \`Bridging via Hyperlane…\` → \`Swapping via Hyperlane…\`. The whole operation is a swap from the user's POV; "bridging" sounds like a separate product.
- **Destination card**: drop the \`Min: X TOKEN\` line in favor of the output's USD value, matching the origin card's layout. Adds **price impact** next to the USD (e.g. \`\$1.20 (-2.4%)\`) with neutral / amber / red tinting at -1% / -3% thresholds — standard DEX-UI signal for how much value the swap is losing to pool slippage + fees combined. Renders only when both sides have a known \`coinGeckoId\` price.
- **Fee pill**: when every fee component has a USD price, show \`Fees: ≈\$0.04 (3.2%)\` mirroring the bridge form. Falls back to the raw per-token amounts (\`0.123 ETH, 0.456 USDC\`) if **any** component is unpriced (no partial sums — would mislead the % readout). Token breakdown stays accessible via the existing modal.
- **Fee USD fetcher**: the swap-side \`FeeSectionButton\` pumps fee components' coinGeckoIds into the shared \`useTokenPricesByIds\` cache so the fetcher pulls anything the user hasn't already triggered via the picker (e.g. dest-chain native for IGP). Shared hook handles dedup / debounce / backoff — no new RPC infrastructure.
- **New helper \`getTotalFeeUsd\`** in \`swap/balances/utils.ts\` — returns \`number | null\`. \`null\` propagates "unknown / partial" and triggers the token-amount fallback in the UI.
- **\`resolveCoinGeckoId(component, tokenMap)\` helper** — engine's \`/v1/tokens\` already includes native entries with \`coinGeckoId\` on every chain that supports swap source (\`ETH → ethereum\`, \`TRX → tron\`), so tokenMap lookup is sufficient. No special-case native branch needed.
- **Switch to \`fromWei\`** for amount-to-decimal conversion in \`getUsdValue\` / \`getTotalFeeUsd\` — proper BigNumber precision for large bigints, matches the rest of the warp-ui's amount math.
- **Destination tx is the ICA execution, not the bridge drop-off**: prefer the \`reveal\` message ID for delivery polling when present (its delivery tx IS the ICA execution on the destination chain — what actually completes the swap). Falls back to \`warp\` for pure-bridge routes. The "delivered" indicator and \`destinationTxHash\` now point at the ICA tx.
- **Default tokens on /swap cold load**: mirrors the bridge tab's \`defaultOriginToken\` / \`defaultDestinationToken\`. When the URL has no \`origin*\` / \`destination*\` params, prefill from new \`config.defaultSwapOriginToken\` / \`defaultSwapDestinationToken\` (format: \`chainName-address\`). Defaults ship as BNB on BSC → ETH on Base. Same \`useTokens({ ids })\` fetch path as the URL flow — defaults inherit matching, store-sync, USD readout, price-impact rendering, etc.

## Known follow-up (not in this PR)

For ICA flows we sometimes see two \`Reveal Message ID\` entries in the details modal — one is likely the bridge message getting mislabeled because \`route.raw.steps[*].router\` doesn't match the on-chain dispatcher for some bridge types (CCTP-v2-fast, XVELO etc). The labeler in \`useSwap.ts:labelMessages\` needs a wider sender check (or an engine-side \`dispatcher\` field). Tracking separately.

### DIFF
--- a/src/features/swap/FeeSectionButton.tsx
+++ b/src/features/swap/FeeSectionButton.tsx
@@ -1,8 +1,11 @@
 import { ChevronIcon, FuelPumpIcon, useModal } from '@hyperlane-xyz/widgets';
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 
+import { useChains } from '../api/hooks';
+import { getFeePercentage } from '../balances/feeUsdDisplay';
 import { useMultiProvider } from '../chains/hooks';
-import { formatFeeAmount } from './balances/utils';
+import { useStore } from '../store';
+import { formatFeeAmount, formatUsd, getTotalFeeUsd } from './balances/utils';
 import { FeeBreakdownModal } from './FeeBreakdownModal';
 import { getTokenByKeyFromMap, useTokenByKeyMap } from './tokens/hooks';
 import type { UiToken } from './tokens/types';
@@ -11,23 +14,44 @@ import type { FeeBreakdown, FeeComponent } from './types';
 interface Props {
   feeBreakdown: FeeBreakdown | undefined;
   isLoading: boolean;
+  /** USD value of the swap input, used to compute fee-as-% of transfer. */
+  inputUsd: number | null;
 }
 
-// Pill button — fuel-pump icon + "Fees: <total>" + chevron.
-// Fees grouped by fungibility (chainId + token address).
-export function FeeSectionButton({ feeBreakdown, isLoading }: Props) {
+// Pill button — fuel-pump icon + "Fees: <total> <USD> (<pct>)" + chevron.
+// Fees grouped by fungibility (chainId + token address). USD value +
+// percentage mirror the bridge form's FeeSectionButton presentation.
+export function FeeSectionButton({ feeBreakdown, isLoading, inputUsd }: Props) {
   const { isOpen, open, close } = useModal();
   const loadingText = useLoadingDots(isLoading);
   const tokenMap = useTokenByKeyMap();
   const multiProvider = useMultiProvider();
+  const { data: chainsResp } = useChains();
+  const tokenPrices = useStore((s) => s.tokenPrices);
 
-  const components = feeBreakdown?.components ?? [];
+  const components = useMemo(() => feeBreakdown?.components ?? [], [feeBreakdown?.components]);
   const isClickable = components.length > 0 && !isLoading;
 
   let feeText: string;
   if (isLoading) feeText = loadingText;
   else if (components.length === 0) feeText = '-';
   else feeText = formatTotalFee(components, tokenMap, multiProvider);
+
+  // Flat USD price map (coinGeckoId → number) for the fee summer.
+  const priceMap = useMemo(() => {
+    const out: Record<string, number> = {};
+    for (const [id, entry] of Object.entries(tokenPrices)) {
+      if (entry.usd != null) out[id] = entry.usd;
+    }
+    return out;
+  }, [tokenPrices]);
+
+  const feeUsd = useMemo(
+    () => getTotalFeeUsd(components, tokenMap, chainsResp?.chains, priceMap),
+    [components, tokenMap, chainsResp?.chains, priceMap],
+  );
+  const feeUsdText = feeUsd > 0 ? formatUsd(feeUsd, true) : null;
+  const pctText = inputUsd != null ? getFeePercentage(feeUsd, inputUsd) : null;
 
   return (
     <>
@@ -42,7 +66,7 @@ export function FeeSectionButton({ feeBreakdown, isLoading }: Props) {
         disabled={!isClickable}
       >
         <FuelPumpIcon width={14} height={14} className="mr-1" />
-        Fees: {feeText}
+        Fees: {feeUsdText ? `${feeUsdText}${pctText ? ` (${pctText})` : ''}` : feeText}
         {isClickable && <ChevronIcon direction="e" width="0.6rem" height="0.6rem" />}
       </button>
       {feeBreakdown && (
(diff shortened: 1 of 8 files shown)


src/features/swap/FeeSectionButton.tsx
```diff
@@ -11,24 +14,61 @@ import type { FeeBreakdown, FeeComponent } from './types';
...
+  useTokenPricesByIds(feeCoinGeckoIds);
+
+  // Flat USD price map (coinGeckoId → number) for the fee summer.
+  const priceMap = useMemo(() => {
+    const out: Record<string, number> = {};
+    for (const [id, entry] of Object.entries(tokenPrices)) {
+      if (entry.usd != null) out[id] = entry.usd;
+    }
+    return out;
+  }, [tokenPrices]);
```
paulbalaji: Non-blocking (medium): this discards the return of `useTokenPricesByIds(feeCoinGeckoIds)` and then rebuilds the same price map by hand from `store.tokenPrices`. The hook already returns `{ prices }` — the identical flattened `Record<string, number>` (`useTokenPrice.tsx:133-139`). Suggest:

```ts
const { prices: priceMap } = useTokenPricesByIds(feeCoinGeckoIds);
```

and drop the `tokenPrices` selector (line 29) plus this `priceMap` useMemo. Removes ~10 lines and a third copy of this flatten loop.

src/features/swap/SwapForm.tsx
```diff
@@ -618,7 +626,21 @@ function DestinationTokenCard({
...
+                className={`ml-1 ${
+                  priceImpactPct <= -3
+                    ? 'text-red-500'
+                    : priceImpactPct <= -1
+                      ? 'text-amber-600'
+                      : ''
+                }`}
+              >
+                ({priceImpactPct >= 0 ? '+' : ''}
+                {priceImpactPct.toLocaleString('en-US', { maximumFractionDigits: 2 })}%)
```
paulbalaji: Two minor nits:

1. **(low)** The price-impact tint thresholds `-3` / `-1` are inline magic numbers in the JSX. Extracting to named module constants (e.g. `PRICE_IMPACT_DANGER_PCT = -3`, `PRICE_IMPACT_WARN_PCT = -1`) makes them a single tuning point and self-documenting.
2. **(cosmetic)** `toLocaleString('en-US', { maximumFractionDigits: 2 })` sets only the max, so a whole-number impact renders `(+5%)` rather than `(+5.00%)`, inconsistent with the 2-decimal fee % style. Add `minimumFractionDigits: 2` if you want them to match.

src/features/swap/balances/utils.ts
```diff
@@ -21,5 +22,31 @@ export function getUsdValue(
...
+  tokenMap: Map<string, UiToken>,
+  prices: Record<string, number>,
+): number | null {
+  let total = 0;
+  for (const c of components) {
+    const { coinGeckoId, decimals } = resolveCoinGeckoId(c, tokenMap);
+    if (!coinGeckoId) return null;
+    const price = prices[coinGeckoId];
+    if (price == null) return null;
+    total += parseFloat(fromWei(c.amount.toString(), decimals)) * price;
```
paulbalaji, line 49: Non-blocking: the new `resolveCoinGeckoId` / `getTotalFeeUsd` helpers have no direct unit tests, unlike the reused `getFeePercentage` (covered in `feeUsdDisplay.test.ts`). A couple of cases would be cheap insurance — especially the all-or-nothing null behavior (any unpriced component → `null`) and the `fromWei` precision path for a large bigint amount.

paulbalaji (approved):
## Review Summary

Approve. Clean, well-scoped stacked PR (base `xaroz/swap-form`). Reviewed across correctness, code quality, and security — no blockers, only the non-blocking nits inline below. Validation passes: `pnpm typecheck` clean, `pnpm lint` 0 errors (3 pre-existing warnings outside the diff), full suite green (252 tests / 24 files).

### Verified strong points
- **`fromWei` switch** in `getUsdValue` / `getTotalFeeUsd` is a genuine precision improvement over `Number(bigint) / 10 ** decimals` (avoids float precision loss above 2^53).
- **Fails safe**: `getTotalFeeUsd` is all-or-nothing (returns `null` if any component is unpriced → UI falls back to raw token amounts rather than an understated sum), and price impact returns `null` when either side lacks a price (hidden, never a falsely reassuring number). `getFeePercentage` additionally guards `inputUsd <= 0`.
- **ICA reveal-vs-warp polling** (`SwapDetailsModal`) makes "Delivered" reflect the actual destination ICA execution — strictly *less* premature than the old warp-message polling. Falls back to `warp` for pure-bridge routes.
- **No XSS/injection surface**: URL-derived ids are lowercased and only matched against the engine token allowlist; form values come from matched `UiToken`s, never raw URL strings; no `dangerouslySetInnerHTML`. No new fetch endpoints / CSP changes.
- Default-token format divergence from the bridge tab (`chain-address` vs `chain-SYMBOL`) is intentional and documented.

### Note
- No direct unit tests for the new `getTotalFeeUsd` / `resolveCoinGeckoId` helpers (flagged inline) — the reused `getFeePercentage` is covered.

troykessler (approved):

src/features/swap/balances/utils.ts
Xaroz: Fixed in 137b73a2: added focused unit tests for fee CoinGecko resolution, all-or-nothing null behavior, and large bigint USD conversion.

src/features/swap/SwapForm.tsx
Xaroz: Fixed in 137b73a2: extracted named price-impact thresholds and formatted the displayed percentage with two decimal places.

src/features/swap/FeeSectionButton.tsx
Xaroz: Fixed in 137b73a2: reused the price map returned by useTokenPricesByIds and removed the duplicate store selector/flattening loop.

PR #1140, opened 2026-06-08
feat: unify transaction history

Xaroz:
## Summary
- unify bridge and swap history in a single typed transaction store
- open bridge or swap detail modals from the shared sidebar history
- hide GraphQL bridge legs that are already represented by local swap history

### DIFF
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -263,7 +263,7 @@ export const useStore = create<AppState>()(
       // User transaction history
       transactionHistory: [],
       addBridgeTransaction: (data) => {
-        const id = makeTransactionId(TransactionHistoryItemType.Bridge, data.timestamp);
+        const id = createTransactionId(TransactionHistoryItemType.Bridge, data.timestamp);
         set((state) => ({
           transactionHistory: [
             ...state.transactionHistory,
@@ -273,7 +273,7 @@ export const useStore = create<AppState>()(
         return id;
       },
       addSwapTransaction: (data) => {
-        const id = makeTransactionId(TransactionHistoryItemType.Swap, data.timestamp);
+        const id = createTransactionId(TransactionHistoryItemType.Swap, data.timestamp);
         set((state) => ({
           transactionHistory: [
             ...state.transactionHistory,
@@ -419,21 +419,23 @@ export const useStore = create<AppState>()(
           transfers?: TransferContext[];
           swaps?: SwapHistoryItem[];
         };
-        if (state.transactionHistory) {
+        if (Array.isArray(state.transactionHistory)) {
           return {
             chainMetadataOverrides: state.chainMetadataOverrides ?? {},
             transactionHistory: state.transactionHistory,
           };
         }
 
+        const transfers = Array.isArray(state.transfers) ? state.transfers : [];
+        const swaps = Array.isArray(state.swaps) ? state.swaps : [];
         const transactionHistory: TransactionHistoryItem[] = [
-          ...(state.transfers ?? []).map((data) => ({
-            id: makeTransactionId(TransactionHistoryItemType.Bridge, data.timestamp),
+          ...transfers.map((data) => ({
+            id: createTransactionId(TransactionHistoryItemType.Bridge, data.timestamp),
             type: TransactionHistoryItemType.Bridge,
             data,
           })),
-          ...(state.swaps ?? []).map((data) => ({
-            id: makeTransactionId(TransactionHistoryItemType.Swap, data.timestamp),
+          ...swaps.map((data) => ({
+            id: createTransactionId(TransactionHistoryItemType.Swap, data.timestamp),
             type: TransactionHistoryItemType.Swap,
             data,
           })),
@@ -462,8 +464,11 @@ export const useStore = create<AppState>()(
   ),
 );
 
-function makeTransactionId(type: TransactionHistoryItem['type'], timestamp: number): string {
-  return `${type}-${timestamp}-${Math.random().toString(16).slice(2)}`;
+function createTransactionId(type: TransactionHistoryItem['type'], timestamp: number): string {
+  const suffix =
+    crypto.randomUUID?.() || `${Date.now()}-${Math.random().toString(36).substring(2, 9)}`;
+
+  return `${type}-${timestamp}-${suffix}`;
 }
 
 async function initWarpContext({
--- a/src/features/store.ts
+++ b/src/features/store.ts
@@ -41,7 +41,16 @@ import {
 import { assembleWarpCoreConfig } from './warpCore/warpCoreConfig';
 
 // Increment this when persist state has breaking changes
-const PERSIST_STATE_VERSION = 2;
+const PERSIST_STATE_VERSION = 3;
+
+export const TransactionHistoryItemType = {
+  Bridge: 'bridge',
+  Swap: 'swap',
+} as const;
+
+export type TransactionHistoryItem =
+  | { id: string; type: typeof TransactionHistoryItemType.Bridge; data: TransferContext }
+  | { id: string; type: typeof TransactionHistoryItemType.Swap; data: SwapHistoryItem };
 
 interface WarpContext {
   registry: IRegistry;
@@ -91,12 +100,13 @@ export interface AppState {
   warpCore: WarpCore;
   setWarpContext: (context: WarpContext) => void;
 
-  // User history
-  transfers: TransferContext[];
-  addTransfer: (t: TransferContext) => void;
-  resetTransfers: () => void;
-  updateTransferStatus: (
-    i: number,
+  // User transaction history
+  transactionHistory: TransactionHistoryItem[];
+  addBridgeTransaction: (t: TransferContext) => string;
+  addSwapTransaction: (s: SwapHistoryItem) => string;
+  resetTransactionHistory: () => void;
+  updateBridgeTransactionStatus: (
+    id: string,
     s: TransferStatus,
     options?: {
       msgId?: string;
@@ -105,14 +115,8 @@ export interface AppState {
       destinationTxHash?: string;
     },
   ) => void;
-  failUnconfirmedTransfers: () => void;
-
-  // Swap history (in-memory only; not persisted). Mirrors `transfers` shape.
-  swaps: SwapHistoryItem[];
-  addSwap: (s: SwapHistoryItem) => void;
-  resetSwaps: () => void;
-  updateSwapStatus: (
-    i: number,
+  updateSwapTransactionStatus: (
+    id: string,
     status: SwapStatus,
     options?: {
       msgIds?: LabeledMsgId[];
@@ -121,11 +125,9 @@ export interface AppState {
       destinationTxHash?: string;
     },
   ) => void;
-  failUnconfirmedSwaps: () => void;
-  // Index into `swaps` of the swap currently being viewed in the
-  // SwapDetailsModal. Set by useSwap.execute when a swap starts.
-  activeSwapIndex: number | null;
-  setActiveSwapIndex: (i: number | null) => void;
+  failUnconfirmedTransactions: () => void;
+  selectedTransactionId: string | null;
+  setSelectedTransactionId: (id: string | null) => void;
   // Accumulated engine-token catalogue. Every useTokens() result funnels
   // through syncTokens so SwapForm / SwapDetailsModal lookups go through
   // one place. Keyed by getSwapTokenKey (chainId-address). Not persisted.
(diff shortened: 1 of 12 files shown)


Xaroz:
@claude review

Xaroz:
@claude addressed the review in e48961b2.

- Duplicate store import: merged the two `../store` imports in `TransferTokenForm.tsx` into one import.
- Stale Zustand comment: updated the destination-balance effect comment so it reflects the new immutable transaction updates and why the dependency is the origin tx hash.
- Sidebar merge memoization: memoized the local bridge transaction data before passing it into `useMergedTransferHistory`, avoiding a fresh `.map()` array on every render.
- Selected history vs active swap execution: added `activeSwapTransactionId` so the swap form gates only on the currently executing swap, while `selectedTransactionId` remains just the modal/history selection. Opening an old pending swap from history no longer disables the form.
- Duplicate swap literal: changed the sidebar `HistoryItemType.Swap` value to derive from `TransactionHistoryItemType.Swap`.
- Swap bridge-leg dedup: normalized local swap message ids and GraphQL message ids with `toLowerCase()` before comparing, so casing drift will not duplicate the bridge leg.
- Local bridge identity: attached `transactionId` to local bridge sidebar items when history items are built, so click handling no longer depends on `entry.data === item.data` reference equality.

Skipped the persistence note intentionally: persisted swap history is expected behavior for this PR, and the token metadata fields are there so restored swaps can still render before live token maps hydrate.

Ran `pnpm format`, `pnpm lint`, and `pnpm typecheck` after the fixes.

src/features/transfer/TransfersDetailsModal.tsx
```diff
@@ -80,14 +82,7 @@ export function TransfersDetailsModal({
...
   const warpCore = useWarpCore();
-  const transfers = useStore((s) => s.transfers);
-  const updateTransferStatus = useStore((s) => s.updateTransferStatus);
-
-  // Find the index of this transfer in the store (for updating status)
-  const transferIndex = useMemo(
-    () => transfers.findIndex((t) => t === transfer || (t.msgId && t.msgId === transfer?.msgId)),
-    [transfers, transfer],
-  );
+  const updateBridgeTransactionStatus = useStore((s) => s.updateBridgeTransactionStatus);
```
paulbalaji, line 92: **🔴 High — the bridge modal freezes at its opening status.**

This component destructures every field from the `transfer` prop (`} = transfer || {}` just above), but its callers pass a one-time snapshot captured into React state — `SideBarMenu` (`setSelectedTransfer({ data: latest.data })`, auto-open L224 / click L208) and `embed` (L72). Because the store now updates transfers **immutably** (`updateBridgeTransactionStatus` returns a new `data` object), that captured snapshot never mutates, so an auto-opened transfer is stuck at `Preparing`/`CreatingTxs` and never surfaces `originTxHash`/`msgId`/`Delivered` until the user reopens it. (On the auto-open path it's worse: the snapshot has no `msgId`, so `shouldTrackDelivery` is false and the live delivery query at L96 never fires.) The old in-place mutation masked this.

`SwapDetailsModal` already does the right thing — `transactionHistory.find((e) => e.id === renderedId)` live each render. `transactionId` is already passed here, so mirror that pattern and fall back to the prop only for API-sourced items (which have no id):

```ts
const liveTransfer = useStore((s) =>
  transactionId
    ? s.transactionHistory.find(
        (i) => i.id === transactionId && i.type === TransactionHistoryItemType.Bridge,
      )?.data
    : undefined,
);
const tx = liveTransfer ?? transfer;
// then destructure status/originTxHash/msgId/... from `tx`
```

paulbalaji (commented):
## Review — re-reviewed against current head (`e48961b2`)

Re-reviewed after the `e48961b2` fix commit. The seven items from the prior `@claude` reviews are all correctly addressed: dup import merged, the stale "mutates in place" comment corrected, `bridgeTransferData` memoized, local-bridge identity now threaded via `transactionId` (no more `entry.data === item.data` reference equality), `HistoryItemType.Swap` derived from the store enum, swap-execution vs. UI-selection split into `activeSwapTransactionId`, and msgId casing normalized. The index → stable-id refactor and immutable updates are a genuine improvement. 👍

Posting as a comment (not blocking), but **I would hold off approving until the 🔴 item below** — it breaks the primary "start a transfer and watch it progress" flow, and slipped past the prior LGTM.

### 🔴 High — bridge details modal freezes at its opening status
The store now updates transfers **immutably** (`updateBridgeTransactionStatus` builds a new `data` object), but `TransfersDetailsModal` still reads every field from a one-time `transfer` **prop snapshot** that its callers capture into React state (`SideBarMenu` auto-open/click, `embed` auto-open). The old store mutated `TransferContext` in place, so the snapshot updated under the modal "for free"; with immutable updates it never changes. Result: after you initiate a transfer, the auto-opened modal stays at `Preparing`/`CreatingTxs` and never shows the origin tx hash, msgId, or Delivered until you close & reopen. Worse for the auto-open path — the frozen snapshot has no `msgId`, so `shouldTrackDelivery` is false and the live delivery query never runs either. `SwapDetailsModal` already does this correctly (re-selects live via `transactionHistory.find(id)` each render), which is why swaps are unaffected. `transactionId` is already plumbed through both callers, so the fix is local to the modal. Inline detail + suggested patch below.

### 🟡 Medium — persisted in-flight swaps can read stale after reload
Swaps are now persisted (intended). Inline note on the `visibleMessages` suppression — the gap is reconciliation, not persistence; acknowledging the author's deliberate choice to keep persistence + token meta.

### Acknowledged / already addressed — not re-raising
- All 7 items from the prior `@claude` reviews ✅ (`e48961b2`).
- Author intentionally persists swap history + `src/dstTokenMeta` so restored swaps render before live token maps hydrate — agreed, reasonable.

**Independent validation:** `pnpm typecheck` ✅ · `pnpm lint` ✅ (3 pre-existing `no-console` warnings, unrelated files) · `vitest` 258 tests ✅. No test covers the v2→v3 `migrate` (low, inline).

src/features/wallet/SideBarMenu.tsx
```diff
@@ -134,14 +216,17 @@ export function SideBarMenu({
...
-    prevTransfersLengthRef.current = transfers.length;
-    if (transfers.length > prev) {
-      setSelectedTransfer(transfers[transfers.length - 1]);
-      setIsModalOpen(true);
+    const prev = prevBridgeTransactionsLengthRef.current;
+    prevBridgeTransactionsLengthRef.current = bridgeTransactions.length;
+    if (bridgeTransactions.length > prev) {
+      const latest = bridgeTransactions[bridgeTransactions.length - 1];
+      if (latest) {
+        setSelectedTransfer({ transactionId: latest.id, data: latest.data });
```
paulbalaji, line 228: Part of the 🔴 above: this auto-open captures `latest.data` as a frozen snapshot, which no longer advances under immutable store updates. No change needed here if `TransfersDetailsModal` is fixed to read live by the `transactionId` already passed at L346 — this line can keep passing the id.

src/pages/embed.tsx
```diff
@@ -64,7 +69,7 @@ function useAutoTransferModal() {
         );
         return;
       }
-      setSelectedTransfer(latestTransfer);
+      setSelectedTransfer({ id: latestTransfer.id, data: latestTransfer.data });
```
paulbalaji, line 74: Same frozen-snapshot issue as `SideBarMenu` (🔴 above): `selectedTransfer.data` won't update as the transfer progresses. Resolved centrally once `TransfersDetailsModal` reads live data by the `transactionId` already passed at L115 — no embed-specific change required beyond that.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -90,9 +127,23 @@ export function SideBarMenu({
...
+    const ids = new Set<string>();
+    for (const item of transactionHistory) {
+      if (item.type !== TransactionHistoryItemType.Swap) continue;
+      for (const msg of item.data.msgIds ?? []) ids.add(msg.msgId.toLowerCase());
+    }
+    return ids;
+  }, [transactionHistory]);
+
+  const visibleMessages = useMemo(
+    () => messages.filter((message) => !swapMessageIds.has(message.msgId.toLowerCase())),
```
paulbalaji, line 141: **🟡 Medium — combined with swap persistence, a delivered swap can look stuck after reload.**

Swaps are now persisted, but a swap's delivery only polls inside an open `SwapDetailsModal` (keyed on `selectedTransactionId`, which is correctly *not* persisted → `null` after reload). Meanwhile this filter hides the GraphQL bridge leg for the swap's `msgIds`. So after a reload, a swap that actually delivered shows locally as `Bridging` *and* its completing API row is suppressed — nothing reconciles it until the user manually opens the swap. Bridges don't have this problem because `useMergedTransferHistory` backfills their status from the indexer with no modal open.

Suggest either reconciling non-final persisted swaps from the API on load, or only suppressing the API leg once the local swap status is final. (Persisting swaps itself is intended — this is purely about reconciliation.)

src/features/store.ts
```diff
@@ -393,13 +411,43 @@ export const useStore = create<AppState>()(
       partialize: (state) => ({
         // fields to persist
         chainMetadataOverrides: state.chainMetadataOverrides,
-        transfers: state.transfers, // Keep for transfers through non-indexed routes
+        transactionHistory: state.transactionHistory,
       }),
       version: PERSIST_STATE_VERSION,
+      migrate: (persistedState) => {
```
paulbalaji, line 417: The v2→v3 `migrate` transforms existing users' persisted data but has no test coverage — the highest-risk untested logic in the PR. A small unit test would be cheap insurance: assert old `{ transfers: [...] }` maps to `transactionHistory` bridge items, and that the already-migrated (`state.transactionHistory` present) passthrough branch is a no-op.

src/features/swap/SwapDetailsModal.tsx
```diff
@@ -56,40 +56,42 @@ const STATUS_DESCRIPTION: Record<SwapStatus, string> = {
...
+  // Keep the last id alive after close so delivery polling continues in background.
+  const lastIdRef = useRef<string | null>(null);
+  if (selectedTransactionId != null) lastIdRef.current = selectedTransactionId;
 
-  const renderedIndex = lastIndexRef.current;
-  const swap = renderedIndex != null ? swaps[renderedIndex] : undefined;
+  const renderedId = lastIdRef.current;
+  const item =
+    renderedId != null ? transactionHistory.find((entry) => entry.id === renderedId) : undefined;
+  const swap = item?.type === 'swap' ? item.data : undefined;
```
paulbalaji: Nit: `=== 'swap'` magic string — the PR added `TransactionHistoryItemType.Swap` for exactly this. Same at `SwapForm.tsx:175` and `:329`. Minor consistency only.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -134,14 +216,17 @@ export function SideBarMenu({
...
+    const prev = prevBridgeTransactionsLengthRef.current;
+    prevBridgeTransactionsLengthRef.current = bridgeTransactions.length;
+    if (bridgeTransactions.length > prev) {
+      const latest = bridgeTransactions[bridgeTransactions.length - 1];
+      if (latest) {
+        setSelectedTransfer({ transactionId: latest.id, data: latest.data });
+        setIsModalOpen(true);
+      }
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps -- transfers.length increasing guarantees a new transfers ref; listing transfers would re-run on status updates
```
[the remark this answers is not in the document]
Xaroz: Fixed in 68bf8fc9. Updated the eslint-disable explanation to match `bridgeTransactions.length`.

src/features/wallet/SideBarMenu.tsx
```diff
@@ -134,14 +216,17 @@ export function SideBarMenu({
   // previous transfer, which would happen if we triggered on transferLoading
   // because addTransfer is called after setTransferLoading(true)).
```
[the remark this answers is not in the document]
Xaroz: Fixed in 68bf8fc9. Updated the comment to reference bridge transactions and `addBridgeTransaction`.

src/features/swap/SwapDetailsModal.tsx
```diff
@@ -396,20 +404,22 @@ function formatAmount(amount: string, decimals: number | undefined): string {
 //   120 s → auto-mark Failed (preserves originTxHash)
 function ConfirmingOriginHint({ swap }: { swap: SwapHistoryItem }) {
   const [showHint, setShowHint] = useState(false);
-  const updateSwapStatus = useStore((s) => s.updateSwapStatus);
-  const swapIndex = useStore((s) => s.activeSwapIndex);
+  const updateSwapTransactionStatus = useStore((s) => s.updateSwapTransactionStatus);
+  const transactionId = useStore((s) => s.selectedTransactionId);
```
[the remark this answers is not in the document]
Xaroz: Fixed in 68bf8fc9. `ConfirmingOriginHint` now receives the rendered transaction id directly, with named timeout constants.

src/pages/embed.tsx
```diff
@@ -47,8 +47,13 @@ function usePostMessageBridge() {
 
 /** Auto-opens TransfersDetailsModal when a new transfer starts. */
 function useAutoTransferModal() {
-  const transfers = useStore((s) => s.transfers);
-  const [selectedTransfer, setSelectedTransfer] = useState<TransferContext | null>(null);
+  const transfers = useStore((s) =>
+    s.transactionHistory.filter((item) => item.type === TransactionHistoryItemType.Bridge),
+  );
```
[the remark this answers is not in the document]
Xaroz: Fixed in 68bf8fc9. The embed hook now selects `transactionHistory` directly and filters bridge entries in `useMemo`.

src/pages/embed.tsx
Xaroz: Covered by 68bf8fc9. Embed already passes the id; `TransfersDetailsModal` now resolves the live transfer from the store.

src/features/wallet/SideBarMenu.tsx
Xaroz: Covered by 68bf8fc9. The sidebar can keep passing the id; the modal now reads live data by that `transactionId`.

src/features/transfer/TransfersDetailsModal.tsx
Xaroz: Fixed in 68bf8fc9. `TransfersDetailsModal` now re-selects live bridge data by `transactionId` and falls back to the prop for API-sourced rows.

src/features/store.ts
```diff
@@ -414,6 +462,10 @@ export const useStore = create<AppState>()(
   ),
 );
 
+function makeTransactionId(type: TransactionHistoryItem['type'], timestamp: number): string {
+  return `${type}-${timestamp}-${Math.random().toString(16).slice(2)}`;
+}
```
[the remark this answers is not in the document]
Xaroz: Fixed in 68bf8fc9. Renamed the helper to `createTransactionId` and use `crypto.randomUUID?.()` with the app-local fallback pattern.

src/features/store.ts
Xaroz: Hardened in 68bf8fc9. I did not add tests in this pass, but migration now guards `transactionHistory`, `transfers`, and `swaps` with `Array.isArray` so malformed v2 local storage degrades to empty history instead of throwing during hydration.

src/features/wallet/SideBarMenu.tsx
Xaroz: Fixed in 68bf8fc9. GraphQL bridge legs are only suppressed for final local swaps, so non-final persisted swaps can still show indexed progress after reload.

src/features/swap/SwapDetailsModal.tsx
Xaroz: Fixed in 68bf8fc9. Replaced transaction-history swap comparisons with TransactionHistoryItemType.Swap; route step checks stay as-is because those are engine route data.

src/features/transfer/TransfersDetailsModal.tsx
paulbalaji: Verified at `68bf8fc9` ✅ — `TransfersDetailsModal` now resolves `liveTransfer` from `transactionHistory` by `transactionId` (narrowed to Bridge) with `tx = liveTransfer ?? transfer`, and every read switched to `tx`. Confirmed the snapshot freeze is gone for both the auto-open and sidebar-click paths, and API-sourced rows (no id) correctly fall back to the prop. The selector only yields a new ref when that transfer's `data` changes, so it stays live without over-rendering. Thanks for the quick turnaround.

src/features/wallet/SideBarMenu.tsx
paulbalaji: Verified at `68bf8fc9` ✅ — gating the `swapMessageIds` set on `FinalSwapStatuses` means a non-final persisted swap no longer hides its indexed bridge leg, so it surfaces real progress after reload. One minor, non-blocking tradeoff to be aware of: while a bridging swap is still non-final, its indexed bridge leg can briefly show up as a second row until the swap reaches a final status and the leg gets suppressed. That's self-resolving and strictly better than the prior stuck-status behavior, so 👍 — just flagging in case you want to dedupe the transient row later.

src/features/store.ts
paulbalaji: Acknowledged — the `Array.isArray` guards on `transactionHistory`/`transfers`/`swaps` are a solid hardening, and I'm not holding approval on the test. Leaving a focused v2→v3 `migrate` unit test as a follow-up nice-to-have since this path transforms existing users' persisted data.

src/features/swap/SwapDetailsModal.tsx
paulbalaji: Verified ✅ — and agreed on leaving the engine route-step `'bridge'`/`'swap'` comparisons as-is; those are SDK route data, not transaction-history types, so `TransactionHistoryItemType` wouldn't apply there.

paulbalaji (approved):
## Re-review @ `68bf8fc9` — Approving ✅

Re-reviewed the follow-up commit against all open feedback. Both items from my prior review are resolved and independently verified, as are the five Copilot findings and the author's replies on each thread.

### Verified fixes
- 🔴 → ✅ **Bridge modal freeze** — `TransfersDetailsModal` now reads live bridge data from the store by `transactionId` (`const tx = liveTransfer ?? transfer`), falling back to the prop only for API-sourced rows that have no id. The selector returns a new ref only when that transfer's `data` changes, so it stays live without over-rendering. Freeze is gone for both auto-open and sidebar-click.
- 🟡 → ✅ **Persisted swap staleness** — bridge legs are suppressed only for `FinalSwapStatuses`, so a non-final persisted swap surfaces indexed progress after reload. (Minor, non-blocking tradeoff noted in-thread: a transient duplicate bridge-leg row can appear mid-flight until the swap is final — self-resolving and strictly better than the prior stuck status.)
- ✅ **Copilot items** — `createTransactionId` uses `crypto.randomUUID?.()` with a fallback; embed selects `transactionHistory` and filters in `useMemo`; `ConfirmingOriginHint` receives `transactionId` as a prop so the 120s auto-fail survives modal close; both stale comments corrected; `migrate` now `Array.isArray`-guards malformed storage.
- ✅ Remaining `=== 'swap'` literals replaced with `TransactionHistoryItemType.Swap`; engine route-step checks correctly left as-is.

### Independent validation @ `68bf8fc9`
`tsc --noEmit` ✅ · `oxlint ./src` ✅ (3 pre-existing `no-console` warnings, 0 errors) · `vitest run` ✅ **25 files / 258 tests passed**.

### Residual (non-blocking)
No focused unit test for the v2→v3 `migrate` — left as a follow-up; the path is now defensively hardened against malformed localStorage.

Nice work iterating through this — LGTM. 🚀

troykessler:
ran this locally and looks good to me, do we want to visually separate hyperswaps from normal bridge transfers? they look the same and you can only tell if it was a hyperswaps by looking at the assets above. Also something which would be cool is to fetch the swap info from the CCS again and to display the route info again the user took. This will be important for recovery if the user closes his window and comes back again seeing that his hyperswap failed

Xaroz:
@troykessler added a small visual separator in 1422e490: transaction history rows now show a compact type badge on the route line, with Swap for hyperswaps and Bridge for normal bridge/API rows.

I agree on fetching/reconstructing the route info from CCS too, especially for reload/recovery when a hyperswap fails. I kept that out of this PR because the current persisted swap history only has endpoint tokens, amounts, tx hashes, and msgIds; it does not yet store the selected route steps or a CCS/commitment lookup key. I think that should be a follow-up that either persists the selected route as a fallback, then enriches from CCS/GraphQL by originTxHash/msgId/commitment id, or adds an explicit CCS recovery endpoint so the details modal can restore the route the user actually took.

PR #1143, opened 2026-06-10
fix: use reveal message for swap destination tx

Xaroz:
## Summary
- Label swap Hyperlane messages from bridge router senders and CCS message bodies so warp, commit, and reveal IDs are shown correctly.
- Poll the reveal message for the final destination transaction hash.
- Enable Hyperlane Explorer message links by default and keep swap message links behind the shared explorer-link flag.

### DIFF
--- a/src/features/swap/SwapDetailsModal.tsx
+++ b/src/features/swap/SwapDetailsModal.tsx
@@ -19,7 +19,7 @@ import ArrowRightIcon from '../../images/icons/arrow-right.svg';
 import LinkIcon from '../../images/icons/external-link-icon.svg';
 import { Color } from '../../styles/Color';
 import { formatTimestamp } from '../../utils/date';
-import { getHypExplorerMessageLink } from '../../utils/links';
+import { getHypExplorerLink } from '../../utils/links';
 import { logger } from '../../utils/logger';
 import { useMultiProvider } from '../chains/hooks';
 import { getChainDisplayName } from '../chains/utils';
@@ -27,7 +27,7 @@ import { useMessageDeliveryStatus } from '../messages/useMessageDeliveryStatus';
 import { TransactionHistoryItemType, useStore } from '../store';
 import { formatBalance } from './balances/utils';
 import { getTokenByKeyFromMap, useTokenByKeyMap } from './tokens/hooks';
-import { FinalSwapStatuses, SwapStatus, type LabeledMsgId, type SwapHistoryItem } from './types';
+import { FinalSwapStatuses, SwapStatus, type SwapHistoryItem } from './types';
 
 const DEFAULT_TIMINGS: StageTimings = {
   [MessageStage.Finalized]: null,
@@ -108,10 +108,9 @@ function SwapDetailsModalInner({
     recipient,
     originTxHash,
     destinationTxHash,
-    msgIds: storedMsgIds,
+    msgIds,
     timestamp,
   } = swap;
-  const msgIds = normalizePersistedMsgLabels(storedMsgIds);
 
   const originChain = multiProvider.tryGetChainName(srcChain) ?? `chain ${srcChain}`;
   const destChain = multiProvider.tryGetChainName(dstChain) ?? `chain ${dstChain}`;
@@ -339,7 +338,7 @@ function SwapDetailsModalInner({
                 key={id}
                 name={LABEL_NAMES[label] ?? 'Message ID'}
                 value={id}
-                url={getHypExplorerMessageLink(multiProvider, originChain, id) ?? undefined}
+                url={getHypExplorerLink(multiProvider, originChain, id) ?? undefined}
               />
             ))}
           </div>
@@ -357,20 +356,6 @@ function SwapDetailsModalInner({
       </div>
     </Modal>
   );
-}
-
-function normalizePersistedMsgLabels(
-  msgIds: LabeledMsgId[] | undefined,
-): LabeledMsgId[] | undefined {
-  if (
-    msgIds?.length === 3 &&
-    msgIds[0]?.label === 'commit' &&
-    msgIds[1]?.label === 'reveal' &&
-    msgIds[2]?.label === 'reveal'
-  ) {
-    return [{ ...msgIds[0], label: 'warp' }, { ...msgIds[1], label: 'commit' }, msgIds[2]];
-  }
-  return msgIds;
 }
 
 function TransferProperty({ name, value, url }: { name: string; value: string; url?: string }) {
--- a/src/features/swap/SwapDetailsModal.tsx
+++ b/src/features/swap/SwapDetailsModal.tsx
@@ -19,7 +19,7 @@ import ArrowRightIcon from '../../images/icons/arrow-right.svg';
 import LinkIcon from '../../images/icons/external-link-icon.svg';
 import { Color } from '../../styles/Color';
 import { formatTimestamp } from '../../utils/date';
-import { getHypExplorerLink } from '../../utils/links';
+import { getHypExplorerMessageLink } from '../../utils/links';
 import { logger } from '../../utils/logger';
 import { useMultiProvider } from '../chains/hooks';
 import { getChainDisplayName } from '../chains/utils';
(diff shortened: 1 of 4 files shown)


src/features/swap/SwapDetailsModal.tsx
```diff
@@ -338,7 +339,7 @@ function SwapDetailsModalInner({
                 key={id}
                 name={LABEL_NAMES[label] ?? 'Message ID'}
                 value={id}
-                url={getHypExplorerLink(multiProvider, originChain, id) ?? undefined}
+                url={getHypExplorerMessageLink(multiProvider, originChain, id) ?? undefined}
```
paulbalaji: **Feature-flag bypass — please confirm intent.** This switched from the gated `getHypExplorerLink` to the new ungated `getHypExplorerMessageLink`, which does *not* check `config.enableExplorerLink`. That flag defaults to `false` (`src/consts/config.ts:66`) and the transfer modal still honors it (`TransfersDetailsModal.tsx:152`). Net effect: in the default config, swap message links render but transfer links don't — an inconsistency under a single global toggle whose documented purpose is "link to the hyperlane explorer in the transfer modal".

If swaps should always link, consider broadening/splitting the flag (e.g. a dedicated `enableSwapExplorerLink`) and noting it. If not, just keep using `getHypExplorerLink` here and the new function isn't needed.

paulbalaji (commented):
## Review Summary

Reviewed the 3-file diff. The **core fix is sound**: replacing the stateful positional `nonWarpCount` heuristic with CCS body-prefix labeling (`0x01`=commit, `0x02`=reveal) is a genuine correctness improvement and removes the commit/reveal swap. `pnpm typecheck` and `pnpm lint` both pass (the 3 lint warnings are pre-existing console statements elsewhere, not introduced here). Verified against the SDK: `m.parsed.body` is a lowercase `0x`-prefixed hex string, and `REVEAL = 0x02` is confirmed by the SDK's ICA reveal encoding + test fixture (`@hyperlane-xyz/sdk` `middleware/account/icaCalls`). For the current EVM/Tron swap scope, warp bodies start with `0x00` (left-padded recipient), so no collision with the CCS prefixes.

Not approving yet — two open questions below (inline) that I'd like the author to confirm before merge:

1. **Explorer-link feature-flag bypass** — the swap modal now uses the *ungated* `getHypExplorerMessageLink`, so swap message links render even when `config.enableExplorerLink` is `false` (the default), while the transfer modal still respects the flag. Intentional, or should swaps stay gated?
2. **Migration positional assumption** — `normalizePersistedMsgLabels` assumes old persisted order is `[warp, commit, reveal]`; the assumption that the genuine reveal is the last entry isn't verified anywhere. Low stakes (see note), but worth confirming.

Context that lowers the severity of #2: the swap CCS feature is **not yet on `main`** (this PR targets the `xaroz/swap-form` feature branch), so `normalizePersistedMsgLabels` only backfills local test history — no production data exists. Per CLAUDE.md, backward-compat is "preferred, not required" for unreleased code.

Everything else is minor/optional polish.

src/utils/links.ts
```diff
@@ -11,6 +11,15 @@ export function getHypExplorerLink(
   msgId?: string,
 ) {
   if (!config.enableExplorerLink || !chain || !msgId) return null;
+  return getHypExplorerMessageLink(multiProvider, chain, msgId);
+}
+
+export function getHypExplorerMessageLink(
```
paulbalaji: Minor (naming): both functions produce the same `/message/<id>` URL — the only difference is the `enableExplorerLink` gate, not the link *kind*. `getHypExplorerMessageLink` reads like it builds a different sort of link. If the two-function split is kept, a name that signals the gating (e.g. `buildHypExplorerMessageLink` for the ungated builder) would be clearer. Tied to the flag question on `SwapDetailsModal.tsx:342` — if swaps stay gated, this second function can be dropped entirely.

src/features/swap/SwapDetailsModal.tsx
```diff
@@ -358,6 +359,20 @@ function SwapDetailsModalInner({
...
+  if (
+    msgIds?.length === 3 &&
+    msgIds[0]?.label === 'commit' &&
+    msgIds[1]?.label === 'reveal' &&
+    msgIds[2]?.label === 'reveal'
+  ) {
+    return [{ ...msgIds[0], label: 'warp' }, { ...msgIds[1], label: 'commit' }, msgIds[2]];
+  }
+  return msgIds;
+}
```
paulbalaji: The remap `[commit, reveal, reveal] → [warp, commit, reveal]` assumes the old persisted dispatch order was `index0=warp, index1=commit, index2=reveal`, i.e. the genuine reveal is always the last entry. I couldn't find anything (code, test, or history) that pins the on-chain Dispatch order — and since CCS posts the commitment before broadcast, a commit-first ordering is at least plausible, in which case `index2` could be the warp message and polling would target the wrong message. Two small asks:

1. Confirm the actual dispatch order against a real swap receipt (or add a brief comment documenting the old vs. new label mapping and *why* the old shape was `commit,reveal,reveal`).
2. Low priority given the feature is unreleased (only backfills local test history) — but if you want to be safe, re-deriving from `msgId`/body rather than trusting stale positional labels would be more robust.

Also minor: this runs unmemoized on every render and allocates a new array when it triggers (call site line 114). A `useMemo` keyed on `storedMsgIds` would preserve referential stability. Negligible perf, optional.

src/features/swap/useSwap.ts
```diff
@@ -250,20 +252,24 @@ function labelMessages(messages: ParsedMessage[], route: AugmentedRoute): Labele
...
+
+    logger.warn('Unexpected swap message shape; labeling as warp', {
+      msgId: msg.msgId,
+      sender: msg.sender,
+    });
+    return { msgId: msg.msgId, label: 'warp' as const };
   });
 }
+
+function getCcsMessageLabel(body: string): LabeledMsgId['label'] | null {
```
paulbalaji, line 271: Nice simplification over the old `nonWarpCount` counter. Two minor notes:

- `0x02`=reveal is authoritatively confirmed in the SDK; `0x01`=commit is convention-based (no COMMIT body fixture/enum in the bundled SDK to verify). A source citation/comment would lock it down.
- The CCS-prefix check runs before the bridge-router-sender check in `labelMessages`. For the current EVM/Tron scope there's no collision (warp bodies start `0x00`), so this is fine today. If swap support ever extends to non-EVM recipients (Solana/Starknet/etc., whose 32-byte recipient first byte is arbitrary), a warp body could begin `0x01`/`0x02` and be mislabeled. Checking the bridge-router allowlist *first* would be the more defensive ordering and costs nothing. Non-blocking.

troykessler (approved):

src/features/swap/SwapDetailsModal.tsx
```diff
@@ -338,7 +339,7 @@ function SwapDetailsModalInner({
                 key={id}
                 name={LABEL_NAMES[label] ?? 'Message ID'}
                 value={id}
-                url={getHypExplorerLink(multiProvider, originChain, id) ?? undefined}
+                url={getHypExplorerMessageLink(multiProvider, originChain, id) ?? undefined}
               />
```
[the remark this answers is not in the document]
Xaroz: Fixed in 85b8ddc5. The swap message ID row now calls getHypExplorerLink, so it respects config.enableExplorerLink.

src/features/swap/SwapDetailsModal.tsx
Xaroz: Fixed in 85b8ddc5. Swaps now use the same gated helper as transfers, and the global explorer link flag is enabled by default.

src/features/swap/SwapDetailsModal.tsx
```diff
@@ -19,15 +19,15 @@ import ArrowRightIcon from '../../images/icons/arrow-right.svg';
 import LinkIcon from '../../images/icons/external-link-icon.svg';
 import { Color } from '../../styles/Color';
 import { formatTimestamp } from '../../utils/date';
-import { getHypExplorerLink } from '../../utils/links';
+import { getHypExplorerMessageLink } from '../../utils/links';
 import { logger } from '../../utils/logger';
```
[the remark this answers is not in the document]
Xaroz: Fixed in 85b8ddc5. Swap message links now use the gated getHypExplorerLink helper, and explorer message links are enabled by default.

src/features/swap/SwapDetailsModal.tsx
Xaroz: Fixed in 85b8ddc5. Removed the old persisted-label normalization since this swap history shape is still new, so we no longer rely on positional remapping.

src/utils/links.ts
Xaroz: Fixed in 85b8ddc5. Removed the extra exported helper and restored getHypExplorerLink as the single message explorer link helper.

src/features/swap/useSwap.ts
Xaroz: Fixed in 85b8ddc5. Message labeling now checks known bridge routers before CCS body prefixes, with a short note on the CCS first-byte message type.

PR #1149, opened 2026-06-16
feat: unify transfer and swap flow

Xaroz:
## Summary

- Replaces the separate bridge/swap surfaces with a single unified token flow on `/` and embed.
- Combines bridge WarpCore tokens with router-engine swap tokens in one picker, with bridge routes preferred when both route types are available.
- Adds unified bridge exact-input behavior, review details, approval/fee handling, recipient warnings, add-token CTA, quotedCalls support, and bridge transaction submission.
- Adds unified swap quote, approval, slippage, route selection, fee summary, swap details/history, recovery status checks, and submission wiring.
- Adds featured-token filtering, same-collateral bridge route-member lookup, balance-aware picker rows/sorting with progressive hydration, and no Bridge/Swap tags in token picker rows.
- Removes legacy bridge-only form/picker/max/fee code after the unified form became the live surface.
- Hardens swap status/details edge cases: no frontend auto-fail timer for slow origin confirmations, case-sensitive token refs are preserved, and native-output recovery limitations are documented.
- Adds detailed implementation notes in `UNIFIED_TRANSFER_SWAP_PR_NOTES.md`.

## What is left

- Manual QA for real router-engine swap submit/recovery behavior.
- Native-output swap confirmation is still heuristic-only on the frontend; a stronger proof needs an engine/contract-provided destination execution signal.

See `UNIFIED_TRANSFER_SWAP_PR_NOTES.md` for the detailed breakdown and remaining checklist.


src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+            bridgeTransfer,
+            getQuotedCallsParams: quotedCalls.getQuotedCallsParams,
+          });
+        } finally {
+          setTransferLoading(false);
+        }
+        return;
+      }
+
+      if (routeMode === UnifiedRouteMode.Swap) {
```
paulbalaji, line 448: **Sanctions: swap submit path is unscreened.** The submit guard at `:356` only blocks bridge (`routeMode === Bridge && (isSanctioned || !addressConfirmed)`), and `isSanctioned` is itself disabled for swaps (`useIsAccountSanctioned(routeMode === Bridge)` at `:192`). This `submitSwap` branch performs no sanctions check. Corroborates the existing unresolved `github-actions` thread at `:192`. Cross-chain swap routes use the Hyperlane mailbox (same AML surface as bridge), so the gap matters. Suggest screening in both modes (`useIsAccountSanctioned(!!routeMode)`) and lifting the `if (isSanctioned) return;` guard above the route-mode branch.

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+    recipient: effectiveRecipient,
+  });
+
+  const swapValues = useMemo(
+    () => ({
+      srcChain: originToken?.swapToken?.chainId ?? null,
+      dstChain: destinationToken?.swapToken?.chainId ?? null,
+      srcToken: originToken?.swapToken?.address ?? '',
+      dstToken: destinationToken?.swapToken?.address ?? '',
+      amount: values.amount,
```
paulbalaji: **Swap quote not debounced.** `amount: values.amount` is passed raw into `useQuote`, so every keystroke triggers an engine quote POST. The legacy `SwapForm` debounced this (~750ms). Consider `useDebounce(values.amount, ...)` before feeding `swapValues`, matching the bridge path.

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+  const fees = getBridgeFeeEstimate(quote, quotedCalls);
+  const feePrices = useFeePrices(fees, warpCore.tokens, prices);
+  const transferUsd = getBridgeTransferUsd(quote, prices);
+
+  if (!fees && !quotedCalls.isLoading) return null;
+
+  return (
+    <FeeSectionButton
+      isLoading={quotedCalls.isLoading}
+      isError={false}
```
paulbalaji: `isError={false}` is hardcoded, so a failed fee quote can never surface to the user — they see `-`/stale fees instead of an error state. Wire this to the actual quote error (`quotedCalls.isError` / the exact-input quote error).

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+}
+
+function getBridgeFeeEstimate(
+  quote: ExactInputBridgeTransferQuote | undefined,
+  quotedCalls: ReturnType<typeof useQuotedCallsFeeQuotes>,
+): (WarpCoreFeeEstimate & { totalFees: string }) | null {
+  const interchainQuote = quotedCalls.fees?.interchainQuote ?? quote?.interchainQuote;
+  if (!interchainQuote) return null;
+
+  const localQuote = quotedCalls.fees?.localQuote ?? new TokenAmount(0n, interchainQuote.token);
```
paulbalaji: Local gas falls back to `new TokenAmount(0n, ...)` whenever `quotedCalls.fees?.localQuote` is absent (e.g. non-QuotedCalls paths, or `feeQuotingUrl` unset). The review/summary UI then shows local gas as 0 even though the actual submit (WarpCore) charges it. Either omit the row when unknown or label it as unestimated rather than implying zero.

paulbalaji (commented):
## Review: unify transfer + swap flow

Reviewed via a multi-agent pass. **No approval/request-changes — comment only.** Scope verified against the **current base `xaroz/swap-form`** (the base was changed from `main` mid-review; the standalone swap form + legacy bridge form already exist in the base, so this PR is largely a *consolidation* — real diff ≈ +4.8k/−5.8k, net deletion). Findings below are scoped to what this PR actually introduces.

The architecture is clean: unified form correctly wired into `index.tsx` / `embed.tsx` / `HomeView`, legacy `/bridge` + `/swap` routes and surfaces fully removed with no dead imports, bridge-first route detection + exact-input BigInt semantics are sound and well-tested, store persist migration (v2→v3) is backwards-compatible, and quote gating prevents inappropriate fetches. CSP needs no change (`connect-src *` pre-existing; only benign `img-src` logo hosts added) and no XSS found (logos gated by `isHttpsUrl`, token text auto-escaped).

### Existing review thread
- **Corroborated:** the unresolved `github-actions` thread at `UnifiedTokenCard.tsx:192` (swap route bypasses OFAC/Chainalysis screening) is **valid**. Note: the standalone `SwapForm` on the base branch also never screened swaps, so this is a *pre-existing gap consolidated here* rather than purely new — but cross-chain swap routes ride the Hyperlane mailbox (same AML surface as bridge), so it's worth resolving in this PR while both paths live in one component. See my note on the swap submit path below. No author response found.

### Highest-value items
1. Swap submit has no sanctions guard (`:414`) — see inline + the existing `:192` thread.
2. Silent failure paths: `submitSwap` swallows errors (`:1428`), `submitBridge` re-quote has no catch (`:1187`), and bridge `Max` lost its error toast (`:978`). Per the repo's “fail loudly” convention these should surface to the user.
3. Swap quote is no longer debounced (`:222`) — per-keystroke engine POSTs.
4. Fee/UI accuracy: `BridgeFeeSummary` can't show quote errors (`:698`) and local gas falls back to `0n` (`:836`).

### Known / author-tracked (not re-litigating, flagged inline briefly)
- Same-collateral route-family dedup (`catalog.ts`/`identity.ts`) and featured-token picker semantics (`list.ts`) are both listed under “Remaining Work” in `UNIFIED_TRANSFER_SWAP_PR_NOTES.md`. The dedup item's e2e guard was weakened (see `cosmos.spec.ts` inline).

### Confirm-intentional (behavior changes vs base)
- `routerApiUrl` now defaults to `undefined` → swap disabled unless `NEXT_PUBLIC_ROUTER_API_URL` set (`config.ts:19`).
- Symbol URL params (`?originToken=USDC`) no longer resolve; only config defaults accept symbols (`initial.ts`). README/notes updated to match — flagging as a breaking change for existing embed integrations using symbol params.
- Swap no longer has a Continue→review step before send (bridge still does). Confirm the lost approval/min-output review rows are acceptable.

Gates reported green by sub-agents: `lint` (pre-existing warnings only), `typecheck`, `vitest` (all pass), `build`.

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+        accounts,
+      );
+      if (!address) return;
+      const { address: connectedDestAddress } = getAccountAddressAndPubKey(
+        multiProvider,
+        destinationToken.bridgeToken.chainName,
+        accounts,
+      );
+      const recipient = values.recipient || connectedDestAddress || address;
+      const maxAmount = await bridgeMax.mutateAsync({
```
paulbalaji: **Bridge `Max` failure is silent.** `bridgeMax` (`useMutation`, `:943`) has no `onError`, and `setMax` awaits `mutateAsync` with no try/catch. `getExactInputBridgeMaxAmount` throws on the common insufficient-native-gas path, so this becomes an unhandled rejection with no user feedback. The deleted `maxAmount.ts` caught this and toasted `"Cannot simulate transfer, <chain> native balance may be insufficient."` — restore that. (Separately, the swap `Max` at `:961` sets full balance with no gas reservation; validation rejects it but the button still offers a broken max.)

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+  destinationToken: UnifiedToken | undefined;
+  recipient: string;
+  sender: string | undefined;
+  bridgeTransfer: ReturnType<typeof useTokenTransfer>;
+  getQuotedCallsParams: ReturnType<typeof useQuotedCallsFeeQuotes>['getQuotedCallsParams'];
+}) {
+  if (!originToken?.bridgeToken || !destinationToken?.bridgeToken || !recipient) return;
+  const warpCore = useStore.getState().warpCore;
+  const inputAmount = BigInt(toWei(values.amount, originToken.bridgeToken.decimals));
+  const quote = await getExactInputBridgeTransferQuote({
```
paulbalaji, line 1288: `submitBridge` re-quotes here, but the caller `onSubmit` (`:394`) wraps only in `try/finally` (no catch). An RPC failure between validation and submit becomes a silent unhandled rejection — the button resets but the user gets no error. Add explicit error handling/toast. (Also note this is a second quote after validation already produced one — minor RPC waste plus quote-drift risk between review and signing.)

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+      dstChainId: values.dstChain,
+      srcToken: srcToken.address,
+      dstToken: dstToken.address,
+      sender,
+      recipient,
+      spender: universalRouter,
+      approvalAmount: amountAtomic,
+      isNative: srcToken.isNative,
+    });
+  } catch {
```
paulbalaji: `catch {}` swallows the swap execution error entirely — only sets history status to `Failed`, with no `logger.error` and no user-facing toast/form error. Per the repo's "fail loudly" convention, log and surface this.

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+  quotedCalls: ReturnType<typeof useQuotedCallsFeeQuotes>;
+  isLoading: boolean;
+}) {
+  const fees = getBridgeFeeItems(quote, quotedCalls);
+  const remoteToken = quote?.connectedDestinationToken;
+  const transferToken = quote?.routeToken;
+  const sameTokenFeeAmount =
+    transferToken &&
+    quotedCalls.quotedCallsParams &&
+    quotedCalls.fees?.tokenFeeQuote?.token.equals(transferToken)
```
paulbalaji: Same-asset fee detection uses `tokenFeeQuote.token.equals(transferToken)`. If the token fee is paid in a *fungible variant* of the transfer token (synthetic/collateral), `.equals` returns false and `sameTokenFeeAmount` stays `0n`, so the approval amount (`:727`) under-includes the fee. Consider `isFungibleWith(transferToken)` to match the SDK's fungibility grouping. (Lower confidence — confirm against how the engine denominates token fees.)

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+  engineEnabled: boolean;
+  isInitialLookupLoading: boolean;
+}) {
+  const { values, errors, setValues, setErrors } = useFormikContext<UnifiedFormValues>();
+  const multiProvider = useMultiProvider();
+  const warpCore = useWarpCore();
+  const collateralGroups = useCollateralGroups();
+  const bridgeTokenMap = useTokenByKeyMap();
+  const { accounts } = useAccounts(multiProvider, config.addressBlacklist);
+  const { routerAddressesByChainMap, setOriginChainName, setTransferLoading } = useStore((s) => ({
```
paulbalaji: `useStore((s) => ({ ... }))` returns a fresh object every render without a `shallow` equality fn, so this component re-renders on every store change regardless of whether these three fields changed. Use the `shallow` comparator (or select fields individually).

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+  useEffect(() => {
+    if (Object.keys(errors).length) setErrors({});
+    setIsBridgeReview(false);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [
+    values.originTokenKey,
+    values.destinationTokenKey,
+    values.amount,
+    values.recipient,
+    routeMode,
```
paulbalaji, line 497: This error-reset effect's dependency list omits `values.slippageBps`, so changing slippage doesn't clear stale validation errors (e.g. a prior min-output/slippage error persists until another field changes). Add `values.slippageBps` to the deps.

src/features/unified/bridgeExactInput.ts
```diff
@@ -0,0 +1,171 @@
...
+      destinationToken,
+    });
+    interchainQuote = quote.igpQuote;
+    tokenFeeQuote = quote.tokenFeeQuote;
+
+    const sameTokenFees =
+      (originToken.isFungibleWith(interchainQuote.token) ? interchainQuote.amount : 0n) +
+      (originToken.isFungibleWith(tokenFeeQuote?.token) ? (tokenFeeQuote?.amount ?? 0n) : 0n);
+    const nextTransferAmount = inputAmount > sameTokenFees ? inputAmount - sameTokenFees : 0n;
+    if (nextTransferAmount === transferAmount) break;
```
paulbalaji: The fixed-point fee loop is capped at 4 iterations. On convergence (`nextTransferAmount === transferAmount`) the returned `transferAmount` and the fee quotes are consistent. But if it exhausts 4 iterations without converging, it returns `transferAmount` (updated at `:159`) paired with fee quotes computed on a *different* amount — an internally inconsistent tuple that then drives `submitBridge`. Result is a conservative over-estimate (transfers slightly less), not fund loss, but worth fixing: re-quote once on the final amount, or only break-and-return on convergence. No test covers the non-convergence path.

src/features/unified/tokens/catalog.ts
```diff
@@ -0,0 +1,115 @@
...
+function mergeToken(existing: UnifiedToken, token: UiToken): UnifiedToken {
+  return {
+    ...existing,
+    chainId: existing.chainId ?? token.chainId,
+    logoURI: existing.logoURI ?? token.logoURI,
+    coinGeckoId: existing.coinGeckoId ?? token.coinGeckoId,
+    swapToken: token,
+    capabilities: {
+      bridge: existing.capabilities.bridge,
+      swap: token.canSwap,
```
paulbalaji, line 79: Same-collateral dedup is last-wins: a later bridge token with an identical `getBridgeTokenIdentity` overwrites the earlier `byIdentity` entry, so only one router survives per collateral identity. This is the same-collateral route-family issue the author already lists under "Remaining Work" — flagging the exact site. The proposed direction (keep route-member metadata behind each unified row so search/URL-hydration/submit can resolve any same-collateral router while the row stays deduped) is the right fix.

tests/e2e-wallet/same-symbol-dedup/cosmos.spec.ts
```diff
@@ -1,57 +1,28 @@
...
 
-    // Same-symbol correctness: the destination flipped because the underlying
-    // warp route differs — NOT because the chain label happened to re-render.
-    // If dedup were broken and stride TIA fell back to the celestia route,
-    // both invocations would leave the destination button identical.
-    expect(strideDestText).not.toBe(celestiaDestText);
+    // The unified form no longer auto-picks the first destination after a new
+    // origin selection; same-symbol correctness here is that the origin label
+    // flips to the selected Cosmos route without falling back to Celestia.
+    expect(strideDestText).toBe(celestiaDestText);
```
paulbalaji: This assertion was inverted from `.not.toBe(celestiaDestText)` to `.toBe(celestiaDestText)` — and since the destination now clears to "Select token" on origin change, both sides are equal trivially. The test no longer guards same-symbol route *distinctness* (the very behavior tracked as the unfixed same-collateral bug). Consider re-asserting distinctness by explicitly selecting a stride-only destination and verifying the route, rather than comparing two empty placeholders.

src/consts/config.ts
```diff
@@ -16,8 +16,7 @@ const explorerApiUrl =
   process.env.NEXT_PUBLIC_EXPLORER_API_URL || 'https://explorer4.hasura.app/v1/graphql';
 const feeQuotingUrl = process.env.NEXT_PUBLIC_FEE_QUOTING_URL || undefined;
 const relayApiUrl = process.env.NEXT_PUBLIC_RELAY_API_URL || undefined;
-const routerApiUrl =
-  process.env.NEXT_PUBLIC_ROUTER_API_URL || 'https://router.services.hyperlane.xyz';
+const routerApiUrl = process.env.NEXT_PUBLIC_ROUTER_API_URL || undefined;
```
paulbalaji, line 19: `routerApiUrl` default changed from `https://router.services.hyperlane.xyz` to `undefined`, which disables the swap surface unless `NEXT_PUBLIC_ROUTER_API_URL` is set. This looks like the intended "engine-absence degrades to bridge-only" behavior — just confirming it's deliberate for default deployments, since it's a user-visible change from the base branch.

src/features/unified/tokens/initial.ts
```diff
@@ -0,0 +1,118 @@
...
+function findTokenFromParams(
+  tokens: UnifiedToken[],
+  params: URLSearchParams | undefined,
+  chainParam: WARP_QUERY_PARAMS,
+  tokenParam: WARP_QUERY_PARAMS,
+): UnifiedToken | undefined {
+  const chainName = params?.get(chainParam);
+  const tokenRef = params?.get(tokenParam);
+  if (!chainName || !tokenRef) return undefined;
+  return findUnifiedTokenByQueryRef(tokens, chainName, tokenRef);
```
paulbalaji, line 63: URL token params now resolve only via `findUnifiedTokenByQueryRef` (address/denom/zeroish), while symbol refs are accepted only for config defaults (`findUnifiedTokenByConfigRef`). So existing deep links like `?originToken=USDC` (symbol) silently stop matching. README/PR-notes were updated to use addresses, so this appears intentional — flagging as a breaking change for any existing embed integrations relying on symbol params.

src/features/unified/tokens/list.ts
```diff
@@ -0,0 +1,96 @@
...
+    return mode;
+  };
+
+  const routableTokens = routeCounterpartToken
+    ? allTokens.filter((token) => !!getRouteMode(token))
+    : allTokens;
+
+  const defaultTokens =
+    !hasFilter && config.featuredTokens.length > 0
+      ? routableTokens.filter((token) => getFeaturedTokenIndex(token) >= 0 || !!getRouteMode(token))
```
paulbalaji: Featured-token default semantics: with `featuredTokens` populated, the `|| !!getRouteMode(token)` clause keeps essentially every routable token, so the default picker is *not* featured-only as described in `UNIFIED_TRANSFER_SWAP_PR_NOTES.md`. The 50-cap (`:93`) correctly applies only when `featuredTokens` is empty. This matches the author's "final pass needed" note — flagging the exact spot.

src/features/unified/UnifiedTokenCard.tsx
```diff
@@ -0,0 +1,1442 @@
...
+    setValues((current) => hydrateInitialUnifiedTokenKeys(current, initialTokenKeys));
+  }, [tokens, collateralGroups, engineEnabled, isInitialLookupLoading, setValues]);
+
+  const routeMode = getUnifiedRouteMode({
+    originToken,
+    destinationToken,
+    collateralGroups,
+    engineEnabled,
+  });
+  const isSanctioned = useIsAccountSanctioned(routeMode === UnifiedRouteMode.Bridge);
```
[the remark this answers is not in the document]
Xaroz: Fixed in deb395eb. Sanctions are now evaluated for any active route mode, and swap submission is blocked when the account is sanctioned.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. Same change as the security thread: sanctions now apply to both bridge and swap submit paths.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. Swap quote amount is debounced before calling the engine, with a freshness check so submit cannot use a stale quote after the input changes.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. Bridge fee summary now receives the actual bridge/quoted-calls error state instead of hardcoding `false`.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. The summary no longer fabricates a zero local gas quote when local gas is unknown.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. Bridge Max now catches quote/simulation failures, logs them, and shows the existing insufficient-native-balance style toast.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. Bridge submit re-quote failures are now caught, logged, and surfaced via toast.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. Swap execution errors are now logged, shown to the user, and still mark the local history item failed.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. Same-token fee detection now uses `isFungibleWith` for the quoted token fee.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. Split the Zustand selection into individual selectors so this component does not subscribe through a fresh object each render.

src/features/unified/UnifiedTokenCard.tsx
Xaroz: Fixed in deb395eb. Added `values.slippageBps` to the error-reset dependencies.

src/features/unified/bridgeExactInput.ts
Xaroz: Fixed in deb395eb. If the fixed-point loop does not converge within the cap, it now re-quotes the final transfer amount so the returned amount and fees are internally consistent.

src/features/unified/tokens/catalog.ts
Xaroz: Fixed in 350b4540. Deduped unified rows now retain all same-collateral bridge route members for routing and URL lookup.

tests/e2e-wallet/same-symbol-dedup/cosmos.spec.ts
Xaroz: Fixed in deb395eb. The Cosmos e2e now selects explicit Celestia and Stride destinations and asserts the route/destination actually differ instead of comparing empty placeholders.

src/consts/config.ts
Xaroz: Confirmed intentional. Default deployments should degrade to bridge-only unless `NEXT_PUBLIC_ROUTER_API_URL` is configured.

src/features/unified/tokens/initial.ts
Xaroz: Confirmed intentional for URL params. Unified deep links use address/denom/zeroish refs so engine lookup can work reliably; legacy symbol refs are still accepted for config defaults only.

src/features/unified/tokens/list.ts
Xaroz: Fixed in deb395eb. The default all-chains picker is now featured-only when `featuredTokens` is configured, and the no-featured path still caps at 50.

PR #1152, opened 2026-06-22
feat: unify transfer form around Universal Router API

Xaroz:
## Summary

- Replace split swap/bridge UI flow with the engine-owned transfer form.
- Delegate token discovery, route availability, quotes, approvals, and executable tx payloads to Universal Router API.
- Remove WarpCore/local warp route UI dependencies and unify duplicated swap/transfer logic under transfer.
- Restore transfer safeguards for smart-contract recipients and zero native destination balance confirmation.
- Update unit/e2e coverage around engine schemas, balances, route filtering, delivery tracking, approvals, and transaction execution.

## Security Notes

- Checked PR diff for accidental secrets, private keys, mnemonics, unsafe HTML rendering, client-side secret storage, and CSP expansion.
- No runtime secret leakage found. Diff hits were test-only generated Solana key material and stubbed test API keys.
- Wallet approval path rejects zero-address spenders before sending approvals.
- `pnpm audit --prod --audit-level high` currently fails on transitive dependency advisories already in the wallet/Cosmos/Radix/Solana dependency tree; not fixed in this PR.


Xaroz:
@claude in depth review

Xaroz:
@claude addressed the actionable review feedback in `ac56090e`.

Fixed:
- removed Solana `extraSigners` handling from the UI after engine stopped exposing private key material
- refreshed known token metadata instead of insert-only cache
- cleared final transfer route cache entries
- made quote expiry UI update via widgets `useTimeout`
- clarified available-routes validation
- made token map selector pure and bootstrapped embed explicitly
- documented the transfer failure safety-net read
- documented native-output destination outcome limitation
- hardened CCS path schema
- added quote expiry and delivery target prioritization tests

Skipped as already covered/not worth destabilizing here:
- destination tx hash backfill was already fixed in `3af71432`
- `TokenAddress` zero-address concern is safe on approval paths; Starknet approvals come through SDK txs, not `route.approval`
- CCS ordering test needs a larger `useTransfer` extraction to do cleanly.

troykessler:
Alright @claude do a detailed review of this PR:

Check if there are any regressions.
Check if there are any improvements on code duplication and missed helper functions.
Check if the new code is well unit tested.
Check if there are any blockers before merging.

troykessler:
so its working very well so good job. One thing maybe we can improve is the token selection on the destination side, there is no info about if the token is bridgeable, swappable or not supported at all, i only get that info once i select it. In my previous poc i solved this with badges, so tokens which were bridgeable showed the bridge badge, then those which where swappable had a swap badge and those tokens which are not supported just had no badge. Maybe this is something for todays demo day we could discuss

troykessler:
also not sure if this is intended, for routes where there is definitely a standard bridge connection like eth ethereum to eth arbitrum it still shows me swap preview interacting with the UR, i would expect that we in this case get the tx information to do a remote transfer on the hyp tokens directly?

troykessler:
also i miss the feature when selecting a bridge token on the origin field that the destination field gets populated automatically with a destination token, else it is very difficult now to find out to what tokens the origin is bridgeable

troykessler:
dont know which other chains are missing but can't find the kyve chain or tokens right now on the new ui

Xaroz:
@claude verified the prior review items remain addressed after the latest route-hint changes. The pushed commit `1b47f075` keeps the existing Solana signer removal, known-token metadata refresh, final route cache cleanup, quote expiry timing, available-routes validation shape, pure token-map selector/bootstrap, failure safety-net docs, native-output limitation docs, CCS schema hardening, and related coverage intact.

Additional verification run for this update: `pnpm format`, `pnpm typecheck`, `pnpm lint` (existing no-console warnings only), and focused vitest for token/API helpers.

Xaroz:
@troykessler fixed the two picker UX points in `1b47f075`.

- Destination route badges: the destination token modal now calls cached `/available-routes` for the selected origin token. Direct results show `Bridge`; bridge wins over `Swap`; `Swap` only shows when both selected origin and destination token have `canSwap`; unsupported rows have no badge. Badges are destination-modal only.
- Origin prefill: selecting an origin token now uses the same cached `/available-routes` result and prefills the destination when the current destination is missing or no longer directly bridgeable. Opening the destination picker reuses the cached result instead of refetching while fresh.
- KYVE: currently investigating separately. The registry/engine config appears to include KYVE route data, but the current API data being consumed by the UI is not surfacing KYVE yet.

Xaroz:
@claude please do another review based on the comment above https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/1152#issuecomment-4926583693

Xaroz:
@claude please review the latest cleanup in `45a5390e`.

Addressed the non-blocking clarity items from the route-hint review:
- renamed `prefillDestinationForBridgeToken` to `prefillBestDestinationToken`, since the prefill now runs for any origin token that can query `/available-routes`
- documented why `AVAILABLE_ROUTES_STALE_TIME` is 10 minutes: avoid repeat picker/prefill calls while quote/submit validation remains the source of truth
- documented that origin changes intentionally prefer a direct bridge destination over preserving a swap-only destination
- updated the available-routes query test name to match the widened semantics
- added an explicit test that a swap-only current destination is replaced by the first direct bridge token

Verification run: `pnpm format`, `pnpm typecheck`, `pnpm lint` (existing warnings only), and focused vitest for token/API helpers.

Xaroz:
@claude please review the latest changes in b6a298a4.

Updated bridge-only route validation so the UI matches the Universal Router engine token discovery semantics instead of assuming every non-native route should validate against collateralAddressOrDenom. This fixes XERC20 routes like EZETH being filtered as unsafe when the API-selected token is addressOrDenom, while preserving collateral-style behavior for M0, Starknet collateral, Sealevel cross-collateral, etc., and native-sentinel handling for protocol-native collateral routes.

Added route-security regression coverage across the affected standards: non-lockbox XERC20, XERC20 lockbox, VS lockbox, M0 Portal, M0 Portal Lite, Sealevel synthetic/cross-collateral, Cw synthetic, CosmosNative synthetic/collateral, Starknet collateral, and Radix collateral. Also checked the local engine on localhost:3001 with representative token ids/search results to confirm the API-returned token addresses match the validator buckets.

Verification run:
- pnpm format
- pnpm lint (existing unrelated no-console warnings only)
- pnpm typecheck
- pnpm vitest src/features/balances src/features/transfer/engine src/features/routeSecurity --watch false --reporter=dot
- pnpm playwright test tests/e2e-wallet/balance-display/evm.spec.ts tests/e2e-wallet/tx-payload/evm.spec.ts tests/e2e-wallet/approval/evm.spec.ts --project=chromium --workers=1

Xaroz:
@claude addressed the follow-up from https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/1152#issuecomment-4929630873 in 521de55d.

Changes:
- Added comments explaining the protocol-native collateral derivation and the CosmosNative SDK enum key/value mismatch.
- Collapsed the duplicate non-EVM route test helpers into one helper.
- Added explicit coverage for EvmHypSyntheticRebase and AleoHypSynthetic route validation.

Verification run:
- pnpm format
- pnpm lint (existing unrelated no-console warnings only)
- pnpm typecheck
- pnpm vitest src/features/balances src/features/transfer/engine src/features/routeSecurity --watch false --reporter=dot
- pnpm playwright test tests/e2e-wallet/balance-display/evm.spec.ts tests/e2e-wallet/tx-payload/evm.spec.ts tests/e2e-wallet/approval/evm.spec.ts --project=chromium --workers=1

paulbalaji:
Bundle-size analysis for this PR:

- Main-route first-load JS: **34.79 MiB → 24.29 MiB gzip (-30.2%)**
- Main-route raw JS: **197.34 MiB → 129.24 MiB (-34.5%)**
- All emitted client JS: **54.10 MiB → 43.66 MiB gzip (-19.3%)**
- ~98% of the first-load gzip reduction is one fewer compiled copy of `@hyperlane-xyz/starknet-core` after removing the local WarpCore path and consolidating the bridge/swap forms.
- This is functional offloading to URE, not dependency pruning: WUI still ships wallets, balances, approvals, broadcasting, delivery tracking, and one full Starknet artifact payload.
- The remaining main-route first load is still **24.29 MiB gzip**; biggest contributors are Starknet core (~10.27 MiB), Radix engine toolkit (~5.74 MiB), and Hyperlane core (~5.00 MiB).
- The corresponding URE functionality currently lives on the `codex/direct-warp-bridge-quotes` stack, so this has a runtime API/deployment dependency on that stack.

Full exact-SHA methodology, attribution, residual costs, and URE ownership breakdown: https://gist.github.com/paulbalaji/71492fbf057df43ed06c71cfe798bec6

PR #1154, opened 2026-06-24
feat: combined bridge swap form

troykessler:


troykessler:
choosing different implementation (engine api) not required anymore

PR #1157, opened 2026-07-13
feat: Aleo token v2 support

troykessler:
## Summary
- Add Aleo (Shield wallet) warp route support: `aleotestnet` chain config, warp route entry, and wallet adapter integration (`AleoWalletContext`) via `@provablehq/aleo-wallet-adaptor-*` and `@hyperlane-xyz/aleo-sdk`
- Make the Aleo network the Shield wallet connects to configurable via `NEXT_PUBLIC_ALEO_NETWORK` (defaults to `mainnet`), instead of it being hardcoded — set to `testnet` for this app's `aleotestnet` warp route
- Upgrade `@hyperlane-xyz/*` deps from locally-linked monorepo tarballs to the latest published npm releases (`widgets`/`sdk`/`utils`/`aleo-sdk`/`cosmos-sdk`/`radix-sdk`/`sealevel-sdk` → `36.0.0`, `deploy-sdk`/`provider-sdk` → `7.0.0`, `starknet-sdk` → `28.1.2`, `tron-sdk` → `23.1.2`), removing the `.monorepo-tarballs` / `pnpm.overrides` linking setup entirely

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (no new errors, pre-existing warnings only)
- [x] `pnpm test` passes (252/252)
- [x] Manually verify Shield wallet connect flow against `aleotestnet` with `NEXT_PUBLIC_ALEO_NETWORK=testnet`


src/features/wallet/context/AleoWalletContext.tsx
```diff
@@ -1,8 +1,13 @@
+import { AleoNetwork, setAleoNetwork } from '@hyperlane-xyz/widgets/walletIntegrations/aleo';
 import { AleoPopupProvider } from '@hyperlane-xyz/widgets/walletIntegrations/aleo/AleoProviders';
 import { AleoWalletProvider } from '@provablehq/aleo-wallet-adaptor-react';
 import { ShieldWalletAdapter } from '@provablehq/aleo-wallet-adaptor-shield';
 import { PropsWithChildren } from 'react';
 
+import { config } from '../../../consts/config';
+
+setAleoNetwork(config.aleoNetwork === 'mainnet' ? AleoNetwork.MAINNET : AleoNetwork.TESTNET);
```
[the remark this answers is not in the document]
troykessler: your wrong, it exists

pnpm-workspace.yaml
```diff
@@ -20,13 +21,13 @@ overrides:
   elliptic: 6.6.1
   pbkdf2: 3.1.3
   form-data: 4.0.4
-  '@ledgerhq/errors': 6.31.0
+  "@ledgerhq/errors": 6.31.0
```
paulbalaji: should stick with '

package.json
```diff
@@ -102,7 +102,16 @@
...
+    "zustand": "^4.4.7",
+    "@hyperlane-xyz/aleo-sdk": "36.0.0",
+    "@hyperlane-xyz/cosmos-sdk": "36.0.0",
+    "@hyperlane-xyz/deploy-sdk": "7.0.0",
+    "@hyperlane-xyz/provider-sdk": "7.0.0",
+    "@hyperlane-xyz/radix-sdk": "36.0.0",
+    "@hyperlane-xyz/starknet-sdk": "28.1.2",
+    "@hyperlane-xyz/sealevel-sdk": "36.0.0",
+    "@hyperlane-xyz/tron-sdk": "23.1.2",
+    "@hyperlane-xyz/core": "11.3.1"
```
paulbalaji: rip why are all these required

src/features/wallet/context/AleoWalletContext.tsx
Xaroz, line 9: shouldn't this depend on the Aleo chain and not a config flag??

package.json
Xaroz: do we need these dependencies?

src/features/transfer/TransferTokenForm.tsx
```diff
@@ -1084,24 +1084,29 @@ async function validateForm(
...
+        const isPredicateRoute = await warpCore.isPredicateSupported(transferToken, destination);
+        if (!isPredicateRoute) throw error;
+        // Only swallow EVM execution reverts (predicate wrapper rejecting without attestation).
+        // Rethrow provider/RPC/network errors so they surface rather than silently
+        // appearing as "validation passed" and failing at submit-time.
+        const causeCode = (error as any)?.cause?.code;
+        if (causeCode !== 'CALL_EXCEPTION' && causeCode !== 'UNPREDICTABLE_GAS_LIMIT') throw error;
+        result = null;
+      }
     }
```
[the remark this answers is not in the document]
troykessler: not needed anymore

package.json
troykessler: these were the remains of the local link monorepo script, remove them again

troykessler: these were the remains of the local link monorepo script, remove them again

src/consts/warpRoutes.yaml
```diff
@@ -2,5 +2,183 @@
...
+      - token: ethereum|ethereum|0x78Ac7FECD1857f5BEEe98AB39096c9781F976D97
+      - token: ethereum|optimism|0x1FdA66FA15A261F01F1E09228D41bD0A806d7529
+      - token: ethereum|polygon|0x1FdA66FA15A261F01F1E09228D41bD0A806d7529
+    decimals: 6
+    logoURI: /deployments/warp_routes/USDC/logo.svg
+    name: USD Coin
+    scale: 1000000000000
+    standard: SealevelHypCollateral
+    symbol: USDC
+    tokenType: collateral
```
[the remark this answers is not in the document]
troykessler: removed

src/consts/warpRouteWhitelist.ts
```diff
@@ -2,7 +2,7 @@
 // Warp Route IDs use format `SYMBOL/chainname1-chainname2...` where chains are ordered alphabetically
 // If left null, all warp routes in the configured registry will be included
 // If set to a list (including an empty list), only the specified routes will be included
-export const warpRouteWhitelist: Array<string> | null = null;
+export const warpRouteWhitelist: Array<string> | null = [];
```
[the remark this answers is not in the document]
troykessler: reverted again

src/features/wallet/context/AleoWalletContext.tsx
troykessler: aleo is different here that we cant dynamically switch the networks, we need to specify the network during build time, before we simply hardcoded mainnet in the codebase itself, now with this we can at least deploy demo urls by setting the env variable to aleotestnet, by default its of course just mainnet

paulbalaji, line 9: [P1] Filter Aleo routes to the configured wallet network

The build-time setting makes sense given that Shield cannot switch dynamically, but it does not constrain the chains the UI exposes. With `warpRouteWhitelist = null`, registry 25.3.0 loads both `.../aleo` and `.../aleotestnet` routes. Widgets 36.0.0 reports no active Aleo chain, and its sender ignores the requested `chainName` while executing through the singleton connected via this setting. Selecting the opposite Aleo chain therefore prepares against one provider and submits on the other network. Please filter Aleo routes/tokens to `config.aleoNetwork`, or reject mismatched Aleo origins before connect/send, so every selectable origin matches Shield's configured network.

scripts/link-monorepo.js
```diff
@@ -172,36 +172,51 @@ if (packedPackages.length === 0) {
...
+    workspaceContent = workspaceContent.replace(/^(overrides:\n)/m, `$1${newLine}\n`);
+  }
+
+  console.log(`   ${name} -> file:${tarballPath}`);
 });
 
-// Write updated package.json
 fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2) + '\n');
-console.log('\n✅ Updated package.json\n');
+fs.writeFileSync(workspacePath, workspaceContent);
```
paulbalaji: [P1] Remove these workspace overrides during unlink

`unlink-monorepo.js` only removes `package.json` overrides. It then deletes `.monorepo-tarballs` and runs `pnpm install`, while the new `pnpm-workspace.yaml` overrides written here still point at those deleted files. Thus the documented `link:monorepo` -> `unlink:monorepo` round trip cannot restore a usable install. Please update the unlink side in the same change to remove matching workspace overrides before deleting the tarballs.

scripts/link-monorepo.js
```diff
@@ -172,36 +172,51 @@ if (packedPackages.length === 0) {
...
 
 packedPackages.forEach(({ name, tarballPath }) => {
-  // Update dependencies
-  if (packageJson.dependencies && packageJson.dependencies[name]) {
-    packageJson.dependencies[name] = `file:${tarballPath}`;
-    console.log(`   ${name} -> file:${tarballPath}`);
-  }
+  // Direct dep — the only 100% reliable anchor for file: resolution
+  if (!packageJson.dependencies) packageJson.dependencies = {};
+  packageJson.dependencies[name] = `file:${tarballPath}`;
```
paulbalaji: [P2] Preserve whether a package was originally a direct dependency

This changed from replacing only existing dependencies to adding every packed default package (plus the hardcoded core package) as a direct runtime dependency. `unlink-monorepo.js` later converts every file dependency to its latest npm version rather than removing packages that were absent originally, so a link/unlink round trip permanently pollutes `package.json`. This exact residue appeared and was manually removed in this PR's later commits. Please retain original membership/version, or keep packages that were not direct dependencies override-only.

scripts/link-monorepo.js
```diff
@@ -172,36 +172,51 @@ if (packedPackages.length === 0) {
...
+
+  // pnpm-workspace.yaml overrides — this is what actually forces transitive resolution
+  const escapedName = name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  const linePattern = new RegExp(`^  '${escapedName}':.+$`, 'm');
+  const newLine = `  '${name}': 'file:${tarballPath}'`;
+  if (linePattern.test(workspaceContent)) {
+    workspaceContent = workspaceContent.replace(linePattern, newLine);
+  } else {
+    // Insert after the 'overrides:' line
+    workspaceContent = workspaceContent.replace(/^(overrides:\n)/m, `$1${newLine}\n`);
```
paulbalaji: [P2] Make the workspace YAML upsert handle valid formatting variants

The lookup only recognizes two-space, single-quoted keys, and this insertion only recognizes an LF-only bare `overrides:` line. Replaying the logic with valid double-quoted/unquoted keys produces duplicate YAML map entries; CRLF or a missing header silently leaves the file unchanged while the script still reports success. The repo already has a YAML dependency, so a structured edit would be safest; at minimum handle quote/newline variants and fail when insertion did not occur.

paulbalaji (commented):
Rereviewed the current head (`cebe233b2325`) against `main`, including the latest fixes, live review threads/author replies, exact installed dependency implementations, and CI. The Provable 0.11.3 patches and dependency integrity look sound; frozen install/policy validation, typecheck, lint (0 errors), and all 252 unit tests pass. I found four current issues worth fixing, documented inline.

src/features/warpCore/warpCoreConfig.ts
```diff
@@ -131,7 +131,7 @@ export async function assembleWarpCoreConfig(
     ...yamlConfig.tokens,
     ...storeOverrideTokens,
   ] as NullableAddressWarpCoreToken[];
-  const tokens = filterUnconnectedToken(dedupeTokens(combinedTokens));
+  const tokens = filterUnconnectedToken(dedupeTokens(filterAleoTokensByNetwork(combinedTokens)));
```
paulbalaji, line 135: [P1] Prune connections when excluding the opposite Aleo network

This removes the opposite-network token objects, but retained peer tokens still have serialized `connections` pointing to them. `WarpCore.FromConfig` asserts every connection target exists, so the published registry now fails application initialization in both modes: exact registry 25.3.0 replay finds 7 dangling `aleotestnet` connections in mainnet mode and 23 dangling `aleo` connections in testnet mode, then throws `Connected token not found`. This matches current e2e smoke: 4/5 tests fail across every retry because the token picker, Send button, and wallet UI never initialize. Please prune connections targeting the excluded chain (or filter complete route configs before flattening), then drop newly unconnected tokens, and cover both network settings with a `WarpCore.FromConfig` integrity test.

scripts/unlink-monorepo.js
```diff
@@ -55,6 +59,30 @@ try {
...
+        .map((item) => item.key.value)
+        .forEach((name) => {
+          if (isPackedTarballRef(workspaceDoc.getIn(['overrides', name]))) {
+            workspaceDoc.deleteIn(['overrides', name]);
+            removedWorkspaceCount++;
+          }
+        });
+
+      if (removedWorkspaceCount > 0) {
+        fs.writeFileSync(workspacePath, String(workspaceDoc));
```
paulbalaji: [P2] Defer the workspace write until dependency restoration succeeds

This write happens before the `npm view` restoration loop and `failedToRestore` check. If any lookup fails, the script exits with `package.json` still pointing at the linked tarballs but `pnpm-workspace.yaml` already stripped of its local override-only entries, so the printed option to keep using the linked versions no longer holds. An isolated failure-path harness reproduces that partial state. Please retain the updated YAML document in memory and write it alongside `package.json` only after the restoration failure check passes.

paulbalaji (commented):
Rereviewed the latest head (`505d48313267`) against the prior reviewed head and `main`, including all resolved threads, the exact registry/SDK connection path, link/unlink failure behavior, and current CI. Three prior tooling findings are fixed. Two current issues remain inline. Local frozen install, typecheck, lint (0 errors), and all 252 unit tests pass; current `e2e-smoke` fails deterministically because WarpCore initialization returns an empty transfer state.

src/features/warpCore/warpCoreConfig.ts
```diff
@@ -176,6 +176,20 @@ function fillMissingCoinGeckoIds(
...
+const ALEO_MAINNET_CHAIN = 'aleo';
+const ALEO_TESTNET_CHAIN = 'aleotestnet';
+
+function filterAleoTokensByNetwork(
+  tokens: NullableAddressWarpCoreToken[],
+): NullableAddressWarpCoreToken[] {
+  const excludedAleoChain =
+    config.aleoNetwork === 'mainnet' ? ALEO_TESTNET_CHAIN : ALEO_MAINNET_CHAIN;
+  return tokens.filter((token) => token.chainName !== excludedAleoChain);
+}
```
Xaroz: this seems wrong, you are filtering all the other tokens out that are not the configured Aleo chains

Xaroz (commented):
test fails and the UI preview also fails

src/features/warpCore/warpCoreConfig.ts
troykessler: Good catch — fixed in 663b22c3. `filterAleoTokensByNetwork` now also prunes each surviving token's `connections` entries that pointed at the excluded chain, and added a regression test (both `aleoNetwork` settings) that runs the filtered tokens through `WarpCore.FromConfig` to catch this class of bug going forward.

scripts/unlink-monorepo.js
troykessler: Fixed in 663b22c3 — the workspace YAML doc is now only mutated in memory at that point; the actual write moved down to right after the `failedToRestore` check, alongside the `package.json` write.

src/features/warpCore/warpCoreConfig.ts
troykessler: Not a bug — the filter keeps everything except the excluded chain. `tokens.filter((token) => token.chainName !== excludedAleoChain)` drops only tokens whose `chainName` equals the *other* Aleo network (`aleotestnet` when configured for mainnet, or vice versa); every non-Aleo token and the matching Aleo chain's tokens pass through unchanged. That said, paulbalaji's comment above caught a real related bug in the same function (dangling `connections` to the excluded chain breaking `WarpCore.FromConfig`) — fixed in 663b22c3, which also adds tests asserting non-Aleo tokens are preserved.

paulbalaji (approved):
Rereviewed the latest head `663b22c3`. The Aleo filter now prunes excluded-network connections and passes `WarpCore.FromConfig` against the full published registry in both modes; unlink failure atomicity also holds in isolated fixtures. No actionable findings. Build, typecheck, lint, unit tests, smoke E2E, and both full E2E shards are green.

PR #1158, opened 2026-07-13
fix: handle native alt-vm balance aliases

Xaroz:
## Summary

- Resolve Aleo native zero-address balance reads through the chain native denom instead of passing the zero alias downstream.
- Resolve Radix native zero-address balance reads through the Radix resource denom while keeping component token balances on the Hyp adapter path.
- Add focused balance routing coverage for Aleo and Radix native aliases.

### DIFF
--- a/src/features/balances/aleo.ts
+++ b/src/features/balances/aleo.ts
@@ -1,4 +1,5 @@
 import { Token, TokenStandard, type MultiProtocolProvider } from '@hyperlane-xyz/sdk';
+import { isZeroishAddress } from '@hyperlane-xyz/utils';
 
 import { logger } from '../../utils/logger';
 import type { BalanceToken } from './types';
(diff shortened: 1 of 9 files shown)


src/features/balances/aleo.ts
```diff
@@ -98,3 +102,15 @@ export function directAleoBalanceDenom(args: {
...
+
+function nativeDenomFor(
+  multiProvider: MultiProtocolProvider,
+  chainName: string | undefined,
+): string | undefined {
+  if (!chainName) return undefined;
+  return multiProvider.tryGetChainMetadata(chainName)?.nativeToken?.denom;
+}
+
+function isZeroAddress(address: string): boolean {
```
xeno097: why not use isZeroishAddress from the utils package?

xeno097 (approved):

paulbalaji (approved):
Reviewed current head 44afeb02. No blocking code findings. Exact-head unit tests (211/211), typecheck, lint, and formatting passed locally. Please ensure the normal CI workflow runs after stack integration or retargeting, since this stacked base does not trigger ci.yml.

src/features/balances/aleo.ts
Xaroz: Fixed in c1fdda35. Verified `isZeroishAddress` matches the native zero aliases we need while leaving `credits`, Aleo token programs, and Radix components untouched, then replaced the local helper.

PR #1165, opened 2026-07-17
feat(rpc): rank-based fallback transport to demote faulty RPCs

haggis-hyperlane:
## Summary

Faulty RPCs (e.g. drpc returning HTTP 429) were being re-hit as the primary on every request instead of being demoted, because viem's `fallback()` transport defaults to `rank: false` — it only advances to the next provider *within* a single failed request, then resets to primary-first. A rate-limited endpoint therefore stayed at the front of the queue and got retried up to 3× (429 is classified retryable) before every failover.

This wires viem's fallback **ranking** into the Nexus RPC paths so faulty endpoints are periodically sampled and reordered by stability (success rate, weighted 0.7) + latency, automatically demoting a 429ing/unhealthy RPC below healthy ones.

## Changes

- **`src/utils/rpcTransport.ts`** (new): shared `rankedFallbackTransport()` helper — `fallback(urls.map(http), { rank: { interval: 10_000 } })`. Interval widened from viem's 4s default to reduce background ping load on public RPCs.
- **`src/features/wallet/context/EvmWalletContext.tsx`**: wagmi per-chain `client()` now uses the ranked helper instead of an unranked `fallback(...)`.
- **`src/features/balances/evm.ts`**: `fetchChainBalances` previously used only `rpcUrls[0]` with a single `http()` transport (no fallback at all). Upgraded to ranked fallback across all configured RPC URLs.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm lint` — 0 errors (3 pre-existing warnings unrelated)
- [x] `pnpm format` applied
- [x] `pnpm test` — 254 passed (254)

### DIFF
--- a/src/features/balances/evm.ts
+++ b/src/features/balances/evm.ts
@@ -7,11 +7,11 @@ import {
   decodeFunctionResult,
   encodeFunctionData,
   erc20Abi,
-  http,
   multicall3Abi,
 } from 'viem';
 
 import { logger } from '../../utils/logger';
+import { rankedFallbackTransport } from '../../utils/rpcTransport';
 import { getTokenKey } from '../tokens/utils';
 import { TokenEntry, fetchSdkBalance } from './tokens';
 
@@ -226,13 +226,17 @@ export async function fetchChainBalances(
   evmAddress: Hex,
   chainAddresses: ChainMap<ChainAddresses>,
 ): Promise<Record<string, bigint>> {
-  const rpcUrl = multiProvider.tryGetChainMetadata(group.chainName)?.rpcUrls?.[0]?.http;
-  if (!rpcUrl) {
+  const rpcUrls =
+    multiProvider
+      .tryGetChainMetadata(group.chainName)
+      ?.rpcUrls?.map((r) => r.http)
+      .filter((url): url is string => !!url) ?? [];
+  if (!rpcUrls.length) {
     logger.warn(`No RPC URL for chain ${group.chainName}, skipping balance fetch`);
     return {};
   }
 
-  const client = createPublicClient({ transport: http(rpcUrl) });
+  const client = createPublicClient({ transport: rankedFallbackTransport(rpcUrls) });
   const batchAddress = getBatchAddress(group.chainName, chainAddresses);
   const balanceOfCallData = encodeFunctionData({
     abi: erc20Abi,
--- a/src/features/balances/evm.ts
+++ b/src/features/balances/evm.ts
@@ -7,11 +7,12 @@ import {
   decodeFunctionResult,
   encodeFunctionData,
   erc20Abi,
+  fallback,
+  http,
   multicall3Abi,
 } from 'viem';
 
 import { logger } from '../../utils/logger';
-import { rankedFallbackTransport } from '../../utils/rpcTransport';
 import { getTokenKey } from '../tokens/utils';
 import { TokenEntry, fetchSdkBalance } from './tokens';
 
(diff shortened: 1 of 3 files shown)


src/features/balances/evm.ts
```diff
@@ -226,13 +226,17 @@ export async function fetchChainBalances(
...
+      .tryGetChainMetadata(group.chainName)
+      ?.rpcUrls?.map((r) => r.http)
+      .filter((url): url is string => !!url) ?? [];
+  if (!rpcUrls.length) {
     logger.warn(`No RPC URL for chain ${group.chainName}, skipping balance fetch`);
     return {};
   }
 
-  const client = createPublicClient({ transport: http(rpcUrl) });
+  const client = createPublicClient({ transport: rankedFallbackTransport(rpcUrls) });
```
paulbalaji: Could we avoid creating the ranked client here on every balance refresh? `useTokenBalances` runs this path per EVM chain every 30s, while viem 2.54's `rankTransports` starts an immediate recursive probe loop with no teardown. Each refresh therefore leaves another client pinging every configured RPC every 10s for the rest of the page lifetime, so background traffic grows continuously. Please cache one client/transport per chain + RPC URL set (or keep ranking disabled for this short-lived path) so the number of rankers stays bounded.

paulbalaji (commented):
One lifecycle issue below. Static checks, unit tests, smoke E2E, and other live checks are otherwise green.

Xaroz:
@claude can you review this PR plus check the performance impact of this?

paulbalaji (commented):
The earlier balance-client probe-loop issue is fixed, and exact-head CI plus smoke E2E are green. One cross-base ordering issue remains below, so I am holding approval.

src/utils/rpcTransport.ts
```diff
@@ -0,0 +1,17 @@
...
+// them by stability (success rate) and latency. A faulty endpoint — e.g. one returning
+// HTTP 429 rate limits — is demoted so healthy endpoints are tried first, instead of
+// re-hitting the same faulty primary on every request (viem's default rank:false).
+// Interval is widened from the 4s default to reduce background ping load on public RPCs.
+const RANK_OPTIONS = { interval: 10_000 } as const;
+
+export function rankedFallbackTransport(httpUrls: string[]): Transport {
+  // With a single endpoint there is nothing to rank; skip the probe loop entirely.
+  if (httpUrls.length === 1) return http(httpUrls[0]);
+  return fallback(
```
paulbalaji, line 13: Could we preserve the configured primary RPC here? `nexus` now includes #1166 (`b71e5682`), which prepends `NEXT_PUBLIC_RPC_OVERRIDES` so the premium endpoint stays primary and public dRPC is fallback. Viem's ranker scores probe stability plus latency from `net_listening`; when both probes succeed, a faster public URL moves ahead of the configured override. Real application 429s do not feed this score, so this can recreate the dRPC-first path #1166 fixed. I reproduced it with two successful mock transports: after one ranking sample, the faster public transport handled `eth_blockNumber` despite the override being first. Please preserve explicit-primary ordering while healthy (for example, disable latency-based reordering for override-aware configs or separate override-aware transport construction) and add a regression test covering #1166's prepended override.

Xaroz:
@claude do another check

haggis-hyperlane:
Closing — superseded by #1166 (use rpc override as primary endpoint).

The 429 root cause is better solved deterministically by #1166's override-as-primary: a keyed premium endpoint is tried first, public URLs (e.g. drpc) only as fallback. viem's `rank` doesn't achieve this — it reorders on background `net_listening` probes rather than real request outcomes, so a rate-limited-but-alive RPC is never demoted, and latency reordering could float a faster public RPC above the #1166 override primary (regressing it).

To fully cover the NES route's 429 exposure, add a keyed `bsc` override alongside the existing `base` Alchemy key in `NEXT_PUBLIC_RPC_OVERRIDES` (config/secret change, no code).

PR #1166, opened 2026-07-18
fix: use rpc override as primary endpoint

haggis-hyperlane:
## Summary
- `NEXT_PUBLIC_RPC_OVERRIDES` was appended to the **end** of the EVM `rpcUrls` list. Nexus's viem `fallback()` transport uses `rank=false`, so it always tries `rpcUrls[0]` (the public primary, e.g. `base.drpc.org`) first and only reaches the override after the primary exhausts its retries — meaning a configured keyed/premium endpoint was effectively never used.
- Prepend the override for all protocols so a keyed premium endpoint is the **primary** RPC, with the registry's public URLs kept as fallback (EVM fallback transport).

## Context
Needed so the newly-configured domain-restricted Alchemy key for `base` is actually used as primary on Nexus, resolving the recurring HTTP 429 rate-limit errors NESA hit on the base leg of NES transfers (public `base.drpc.org` was rate-limiting).

## Test plan
- [ ] `pnpm typecheck` (passes locally)
- [ ] `pnpm lint` (clean; pre-existing console warnings only)
- [ ] After deploy: confirm base RPC requests hit the Alchemy endpoint and NESA transfers succeed

### DIFF
--- a/src/features/chains/metadata.ts
+++ b/src/features/chains/metadata.ts
@@ -10,7 +10,6 @@ import {
   objFilter,
   objMap,
   promiseObjAll,
-  ProtocolType,
   tryParseJsonOrYaml,
 } from '@hyperlane-xyz/utils';
 import { z } from 'zod';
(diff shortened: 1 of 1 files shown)


Skunkchain (approved):
No blocking findings. The change makes configured RPC overrides primary while retaining registry endpoints as fallbacks, preserves existing non-EVM behavior, and does not embed credentials. I also pushed the formatter-only correction. Local Node 24 typecheck, lint, and all 254 unit tests pass; GitHub build, security, format, test, and E2E smoke checks are green.

PR #586, opened 2025-05-22
chore: add SVM rent estimate

Xaroz:
Add SVM chains rent estimate amount on top of the IGP payment quoting, **this is purely presentational and the UI does not add on top of the actual transactions, the rent comes from the payment itself**. This is mean purely as a way to represent the actual fee paid for SVM bridging as close as possible

You can check the transactions and look at the rent paid to see how the estimates are taken:
[Solana](https://solscan.io/tx/53mvkbN21SM7PCv7CY98CDCXqxEhCXMPjtdsVZT6YNuRg98DEji7tkzXPsqjXpMGW3rAsdzZmwERv8CLrBX1BnsC#solBalanceChange): 0.00187224 + 0.00224112 = 0.00411336
[Eclipse](https://eclipsescan.xyz/tx/4xza8yohGKqBNd5QytFWRKLCgycJpYqohnhpKUB4Aen8Yx83jA28pPSE9naT7SHuvcUT2ytximkWF4G3559kvSj#balanceChange): 0.000018292 + 0.000021896 = 0.00004019
[Soon](https://explorer.soo.network/tx/3LwvSmxkca9tFz9CF8YAmLjE1ZL9f1EiV9zCBaSZacr82kACb1zrG3eA49pCKtDFHpszoWGJy3qkJMSPEmiJYNvH): 0.000001932 + 0.000001614 = 0.00000355
[SonicSVM](https://explorer.sonic.game/tx/3fRkH2h2To6ZPnk2PdpkmUs27S79EAVueCcuisV6EcZEi5tbvgyLkaWnDpsLSvhnUaw6rJyqEhcAXE46hiQvg47y?cluster=custom&customUrl=https%3A%2F%2Fapi.mainnet-alpha.sonic.game): 0.00224112 + 0.00187224 = 0.00411336 

as you can see below, the IGP shown would be inaccurate because it does not take into account the rent:
image
when you sign, you would notice this:
Screenshot 2025-05-22 at 11 29 04 AM

and this is the estimation after adding the rent:
image

PR #1170, opened 2026-07-20
feat: SVM offchain fee quoting + same-chain swap fee fixes

xeno097:
## Summary

Brings the SVM offchain fee-quoting flow to `main` (previously only on `nexus`) and fixes fee display for same-chain SVM cross-collateral swaps.

Depends on `@hyperlane-xyz/sdk|utils|widgets` **38.0.0** (bumped from 36.0.0) for the `FeeQuotingV2Client` / `SealevelQuotedTransferProvider` APIs.

## What's included

**SVM offchain quoting**
- `useSvmQuotedTransfer` + `/api/v2/quote/[endpoint]` proxy: discovers the route `fee_config`, builds a `SealevelQuotedTransferProvider`, and eagerly fetches the priced warp + IGP quote so `ReviewDetails` shows the real fee before submit.
- Wires the quoted transfer into `useTokenTransfer` and displays offchain fees for EVM cross-collateral and SVM origins.

**Same-chain swap fee fix**
- A same-chain SVM cross-collateral swap (e.g. `solanamainnet` USDCFEE → USDTFEE) showed a spurious "Interchain Gas" fee: `getInterchainQuote` unconditionally added the SVM rent estimate onto the (genuinely zero) IGP quote.
- `getInterchainQuote` now skips the rent add for same-chain transfers.
- Sender-paid ATA-creation rent is folded into `localQuote` (checked via the Sealevel adapter) only when the recipient's destination-token ATA is missing — Solana's gas estimate omits rent deposits. No protocol-specific fee field is introduced.
- `getTotalFee` skips zero-amount quotes so a `0` fee never appears in the breakdown.

**Ported from #1109** (`fix: warp core config merge logic`)
- Type-aware merge in `reduceOptions` (`warpCoreConfig.ts`): arrays concat, objects deep-merge via `objMerge`, scalars last-wins skipping nullish. Includes the review-requested `objMerge` change and the accompanying tests.

## Testing
- `pnpm typecheck`, `pnpm lint`, `pnpm build` pass.
- Unit tests: `fees.test.ts` (incl. `getInterchainQuote` + zero-skip), `useSvmQuotedTransfer.test.ts` (ATA-rent branches), `warpCoreConfig.test.ts` (`reduceOptions`).
- Manually verified same-chain SVM CCR swap: no interchain fee, no `0.00000 SOL` entry, rent shown only when the ATA needs creating.


src/features/transfer/TransferTokenForm.tsx
```diff
@@ -668,9 +693,11 @@ function ButtonSection({
 
     // Wait for any in-flight offchain quote to settle so a quick Send-click
     // during the first-load / refetch window doesn't fall through to the
-    // plain transferRemote path.
+    // plain transferRemote path. EVM and SVM offchain paths are mutually
+    // exclusive (only one of `quotedCallsParams` / `svmQuotedTransfer` is
+    // populated based on origin protocol).
     const quotedCallsParams = await getQuotedCallsParams();
-    await triggerTransactions(values, routeOverrideToken, quotedCallsParams);
+    await triggerTransactions(values, routeOverrideToken, quotedCallsParams, svmQuotedTransfer);
```
paulbalaji: This only awaits the EVM quote. During initial `svmFeeConfig` discovery, `svmQuotedTransfer` is still null while Send remains enabled, so a quick click passes no provider and `WarpCore` builds the plain transfer path instead. Please expose an async SVM getter/refetch analogous to `getQuotedCallsParams`, or disable Send until discovery settles, and cover deferred discovery in a test.

src/features/transfer/useSvmQuotedTransfer.ts
```diff
@@ -0,0 +1,291 @@
...
+        sender: sender!,
+        recipient,
+        amount: debouncedAmount,
+      }),
+    enabled: shouldFetchFees,
+    refetchInterval: FEE_QUOTE_REFRESH_INTERVAL,
+  });
+
+  return {
+    quotedTransfer,
```
paulbalaji, line 229: `quotedTransfer` stays active after the fee query returns a typed `not_configured` response or exhausts retries. `ReviewDetails` then sees `fees: null` and shows the on-chain fallback, but Send still passes this provider and `buildQuotedTransferTxs` repeats the failed v2 request. Please gate the provider for supported no-quote/error states, or expose an async getter so display and submit share one fallback policy.

src/features/transfer/useSvmQuotedTransfer.ts
```diff
@@ -0,0 +1,291 @@
...
+    sender,
+    recipient,
+    destinationToken,
+  });
+
+  // Local gas estimate so the UI matches what the user will see in their
+  // wallet. Mirrors `useQuotedCalls`'s post-quote local-gas estimate; throws
+  // here drop the whole offchain result, letting the consumer fall through
+  // to on-chain quoting.
+  const localQuote = await warpCore.getLocalTransferFeeAmount({
```
paulbalaji: This estimates a different transaction from the one submitted: SDK `getLocalTransferFeeAmount` has no `quotedTransfer` input, so it builds a plain SVM transfer, while submit prepends `SubmitFeeQuote` and possibly `SubmitIgpQuote` plus extra signers. Local gas (and the related balance/Max validation path) can therefore disagree with the signed transaction. Please add a quoted-provider estimation/validation path and cover transient mode.

src/features/transfer/useSvmQuotedTransfer.ts
```diff
@@ -0,0 +1,291 @@
...
+}: {
+  warpCore: WarpCore;
+  originToken: Token;
+  destinationToken: IToken;
+  destination: string;
+  recipient: string;
+}): Promise<bigint | undefined> {
+  if (originToken.chainName !== destination) return undefined;
+
+  const rentLamports = chainsRentEstimate[originToken.chainName];
```
paulbalaji: `chainsRentEstimate` is not an ATA-rent estimate: #586 derived the Solana value (4,113,360 lamports) by summing two cross-chain message/gas account rents. Reusing it here misstates same-chain ATA creation, and Token-2022 account sizes can differ further. Please query `getMinimumBalanceForRentExemption` for the destination token-account layout, or introduce a separately justified ATA-specific estimate.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -224,6 +238,16 @@ async function executeTransfer({
...
     });
 
+    // DEBUG: dump the built tx shape so we can correlate wallet "unknown tx"
+    // failures with what `getTransferRemoteTxs` actually produced. Especially
+    // useful while wiring SVM offchain quoting end-to-end.
+    // eslint-disable-next-line no-console
+    console.log('[svm-quoted DEBUG] Built transfer txs:', {
+      count: txs.length,
+      summaries: txs.map((tx, i) => summarizeTx(tx, i)),
+      raw: txs,
```
paulbalaji: Please remove this raw log before merge. SDK 38 SVM quoted transaction entries include `extraSigners: Keypair[]`; Solana `Keypair` exposes its secret bytes via enumerable `_keypair.secretKey`, so `raw: txs` puts generated signer private material in browser devtools. Remove both temporary debug logs and `summarizeTx`, rather than logging signer-bearing objects.

paulbalaji (commented):
Reviewed exact head `d50edd1d6c75fa45bdd056f54126f61cfa1131bd`. Confirmed the existing unresolved EVM cross-collateral `destinationToken` thread; additional SVM lifecycle, fee-estimation, rent, and signer-material findings are inline. Focused local tests passed (66/66); current build/typecheck/lint/test/format checks pass.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -237,26 +232,25 @@ async function fetchQuotedCallsFees(
...
+  // quoteExecute eth_call independently — both consume the same `quotes` +
+  // `clientSalt` so the result is deterministic and display ↔ submit match.
+  const { igpQuote, tokenFeeQuote } = await warpCore.getQuotedTransferFee({
+    quotedTransfer,
     originTokenAmount,
     destination,
     sender,
     recipient,
-    quotedCalls: baseQuotedCallsParams,
   });
```
paulbalaji, line 256: `destinationToken` is still missing here. The existing thread was auto-resolved as addressed by `a22c267`, but that commit did not touch this file. SDK 38 only selects the cross-collateral `targetRouter` / `transferRemoteTo` path when `destinationToken` is passed, while submit does pass it, so preview and submit still build different commands. Please include `destinationToken` and cover EVM CC.

src/features/transfer/useSvmQuotedTransfer.ts
```diff
@@ -0,0 +1,351 @@
...
+        amount: debouncedAmount,
+      }),
+    enabled: shouldFetchFees,
+    refetchInterval: FEE_QUOTE_REFRESH_INTERVAL,
+  });
+
+  // The display-time quote settled with an error (typed `not_configured` /
+  // retries exhausted). Submit's `buildQuotedTransferTxs` would repeat the same
+  // failing request, so treat the route as not quote-enabled and fall through.
+  const isQuoteUnavailable = shouldFetchFees && !isFeesLoading && isFeesError;
```
paulbalaji: `isFeesError` is too broad to mean ‘route is not quote-enabled.’ It also covers transient 502/timeouts, invalid quotes, transaction construction/simulation failures, and ATA/mint RPC errors; returning null silently switches submit to the plain path. TanStack also retains prior `fees` after a failed background refetch, so Review can show the offchain quote while submit is unquoted. Please only downgrade explicit no-quote responses and keep display/provider/getter state coherent; unexpected failures should stay blocking.

src/features/transfer/useSvmQuotedTransfer.ts
```diff
@@ -0,0 +1,351 @@
...
+    destinationToken,
+  });
+
+  // Estimate the ACTUAL quoted transaction the user will submit: it prepends
+  // `SubmitFeeQuote` / `SubmitIgpQuote` ixs and extra signers that the SDK's
+  // `getLocalTransferFeeAmount` (no `quotedTransfer` input) omits by building a
+  // plain transfer. Build the same txs as submit and sum their per-tx fees,
+  // reusing the native token from `igpQuote`. A throw here drops the whole
+  // offchain result, letting the consumer fall through to on-chain quoting.
+  const quotedTxs = await warpCore.getTransferRemoteTxs({
```
paulbalaji, line 277: This fixes ReviewDetails estimation, but not the related validation/Max paths: `useFetchMaxAmount` still calls `getMaxTransferAmount` without its supported `feeEstimate`, and `validateForm` calls `validateTransfer`, which internally estimates the plain transfer. They can therefore accept or compute Max against different fees from the quoted transaction submitted here. Please thread the quoted estimate through Max and add a quoted validation path, with token-fee/transient coverage.

paulbalaji (commented):
Rereviewed exact head `a22c267`. The quick-send discovery race, signer-bearing transaction logs, ATA-rent calculation, and ReviewDetails quoted-transaction estimation are fixed. Three current issues remain inline; the existing v2 proxy timeout thread also remains valid. Focused local tests: 65/65; CI and E2E are green.

Xaroz:
@claude do a general review of this, also, is it plausible to have one code path for the hook called useQuotedParams() which depending on the protocol calls the correct offchain fee path instead of having two hooks that passes through components?

src/features/transfer/useTokenTransfer.ts
xeno097: Removed in `a22c267` — both `[svm-quoted DEBUG]` logs and the now-unused `summarizeTx` helper are gone.

src/features/transfer/useSvmQuotedTransfer.ts
xeno097: Fixed in `a22c267` — replaced the `chainsRentEstimate` reuse with the real rent-exempt minimum for the receive account (165B classic SPL, extension-aware size derived from the mint for Token-2022).

src/features/transfer/TransferTokenForm.tsx
xeno097: Fixed in `a22c267` — added an awaited `getQuotedTransfer()` (mirrors the EVM `getQuotedCallsParams`) so Send waits out in-flight `svmFeeConfig` discovery instead of falling through to the plain path.

src/features/transfer/useSvmQuotedTransfer.ts
xeno097: Addressed in `a22c267`, then refined in `b48b2c9` per your follow-up below (only downgrade on explicit no-quote).

src/features/transfer/useSvmQuotedTransfer.ts
xeno097: Fixed in `a22c267` — local gas now estimates the actual quoted txs via `getTransferRemoteTxs({ quotedTransfer })` (which include SubmitFeeQuote/SubmitIgpQuote + extra signers) instead of the plain transfer.

src/features/transfer/useQuotedCalls.ts
xeno097: Resolved by merging `main` (`9f0dbb5`): main threads `destinationToken` into `getQuotedTransferFee` independently, so it is present now and preview matches submit for EVM CC.

src/features/transfer/useSvmQuotedTransfer.ts
xeno097: Fixed in `b48b2c9` — downgrade is now gated on the typed `FeeQuotingNoQuoteAvailableError` only. Transient failures (502/timeout, RPC, tx-construction) keep the provider so submit stays on the quoted path and fails loudly; and since we no longer downgrade on transient errors, the retained `fees` after a failed background refetch stay coherent with the kept provider.

src/features/transfer/useSvmQuotedTransfer.ts
xeno097: Confirmed pre-existing — the EVM quoted-calls path on `main` has the same gap (`getMaxTransferAmount` is never passed `feeEstimate`, and this branch does not touch `maxAmount.ts` or the `validateTransfer` call). Tracking as a follow-up: thread the quoted `feeEstimate` through Max (needs a probe-quote since Max is usually clicked before an amount is entered) plus a quoted validation path — ideally via an SDK `validateTransfer({ feeEstimate })` change so the app side stays a one-liner instead of duplicating balance logic.

src/features/transfer/useQuotedCalls.ts
```diff
@@ -237,26 +236,25 @@ async function fetchQuotedCallsFees(
...
```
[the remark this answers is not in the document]
xeno097: Addressed via merge with `main` (`9f0dbb5`) — `destinationToken` is now threaded into `getQuotedTransferFee`.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -366,3 +390,51 @@ const txCategoryToStatuses: Record<WarpTxCategory, [TransferStatus, TransferStat
...
+    return {
+      ...base,
+      kind: 'v0',
+      ixCount: Array.isArray(compiledIxs) ? compiledIxs.length : undefined,
+      altTableCount: Array.isArray(altLookups) ? altLookups.length : undefined,
+    };
+  }
+
+  return base;
+}
```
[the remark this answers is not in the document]
xeno097: Removed in `a22c267` (along with the debug logs that used it).

src/pages/api/v2/quote/[endpoint].ts
```diff
@@ -0,0 +1,131 @@
...
+    // endpoint === QuoteV2Endpoint.Igp
+    const quote = await client.getIgpQuote({
+      origin,
+      router,
+      destination: destinationDomainId,
+      salt,
+      txSubmitter,
+    });
+    const response: QuoteV2Response = { quote };
+    return res.status(200).json(response);
```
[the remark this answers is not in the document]
Comment (xeno097) [reply src/pages/api/v2/quote/[endpoint].ts:130]: Added in `c6baeca0` — a 10s response timeout returns 504. Note the `FeeQuotingV2Client` exposes no AbortSignal, so this bounds the response, not the in-flight upstream fetch.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -208,14 +213,23 @@ async function executeTransfer({
 
     updateTransferStatus(transferIndex, (transferStatus = TransferStatus.CreatingTxs));
 
+    // eslint-disable-next-line no-console
+    console.log('[svm-quoted DEBUG] about to call getTransferRemoteTxs', {
+      hasQuotedTransfer: !!quotedTransfer,
+      hasQuotedCalls: !!quotedCallsParams,
+      hasAttestation: !!attestationResult,
+    });
```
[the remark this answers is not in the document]
xeno097: Removed in `a22c267`.

src/features/transfer/useTokenTransfer.ts
```diff
@@ -224,6 +238,16 @@ async function executeTransfer({
...
 
+    // DEBUG: dump the built tx shape so we can correlate wallet "unknown tx"
+    // failures with what `getTransferRemoteTxs` actually produced. Especially
+    // useful while wiring SVM offchain quoting end-to-end.
+    // eslint-disable-next-line no-console
+    console.log('[svm-quoted DEBUG] Built transfer txs:', {
+      count: txs.length,
+      summaries: txs.map((tx, i) => summarizeTx(tx, i)),
+      raw: txs,
+    });
```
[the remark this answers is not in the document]
xeno097: Removed in `a22c267`.

Xaroz (approved):

Xaroz:
@claude review the new split path into a single useQuotedCalls, make sure to check for regressions, performance issues with the new paths and if the behavior is correct

paulbalaji (commented):
Rereviewed exact head `b48b2c9547e4e8e25ae61aee190e3a9136504087`, including the unified EVM/SVM provider split. The quick-send, destination-token, actual-tx gas estimation, ATA-rent, logging, typed no-quote, and timeout fixes are present; 65/65 focused tests and all current CI/E2E pass.

The current transient-error thread at `useSvmQuotedTransfer.ts:229` remains valid: after an initial transient quote failure with no cached fees, Review falls back to onchain fees while submit retains/retries the quoted provider, so a retry success can sign a different fee/path than was reviewed. Keep that error blocking or make Review and submit consume the same awaited result. The acknowledged Max/validation gap is pre-existing and can remain a tracked follow-up. No duplicate inline comment added.

xeno097:
Implemented in the universal router api

PR #1173, opened 2026-07-27
fix: pin qr to 0.5.5 to fix WalletConnect QR crash

haggis-hyperlane:
## Summary
Opening the WalletConnect QR modal on Nexus crashed with **"Fatal Error Occurred / invalid border=0"**, tripping the app error boundary. Root cause is a transitive dependency floating forward:

- RainbowKit `2.2.10` renders the WC QR via `cuer@0.0.3`, whose `create()` calls `encodeQR(value, 'raw', { border: 0, ... })` — border hardcoded to `0`.
- `cuer` depends on `qr` with a loose `"~0"` range. Upstream `qr@0.6.0` (published 2026-04-28) **added** `if (border <= 0) throw new Error(\`invalid border=${border}\`)`. `qr@0.5.5` only validated the border *type* and accepted `0`.
- A routine lockfile regen (`chore: merge main into nexus #1163`, 2026-07-17) bumped the lockfile's `qr` from `0.5.5` → `0.6.0`, baking the incompatibility into the deployed build.

This pins `qr` to `0.5.5` via the `pnpm-workspace.yaml` override. Dependency-only change; no app code touched.

## Verification
Ran `cuer`'s exact QR-encode path against both versions:
- `qr@0.6.0`: `encodeQR(..., { border: 0 })` → **THREW: invalid border=0** (reproduces the customer crash)
- `qr@0.5.5` (pinned): `encodeQR(..., { border: 0 })` → **OK, 37-row grid**
- `cuer.create(...)` end-to-end with pinned qr → **OK, 33×33 grid**

## Test plan
- [ ] On a preview build, open the wallet modal and select WalletConnect; confirm the QR renders and no error boundary appears.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -23,6 +23,7 @@ overrides:
   pbkdf2: 3.1.3
   form-data: 4.0.4
   '@ledgerhq/errors': 6.31.0
+  qr: 0.5.5
 
 patchedDependencies:
   '@provablehq/sdk@0.11.3': 3b2a635396ffc676f020ff1d35ebeca1e38f3b02007c8f6379a8737f8c76cff8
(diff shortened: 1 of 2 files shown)


Xaroz:
@claude review

Xaroz (approved):

PR #1178, opened 2026-07-31
perf: reduce Radix and Starknet browser bundles

Xaroz:
## Summary

- update `@hyperlane-xyz/sdk`, `@hyperlane-xyz/utils`, and `@hyperlane-xyz/widgets` from `39.0.0` to `39.1.0`
- consume the released Radix browser-runtime lazy loading and Starknet runtime/deployment-artifact split
- keep the Radix toolkit and gateway client stable across React renders
- keep Starknet provider and chain configuration stable, while using narrow public SDK and registry imports
- tighten route bundle budgets so the reduction cannot regress silently

## Runtime changes

Radix provider construction now loads its browser implementation only when the SDK provider is used. The private-key signer remains outside the normal browser path. The UI also avoids recreating `RadixDappToolkit` and `GatewayApiClient` on every render, preventing duplicate listeners and wallet-state resets.

Starknet runtime consumers now receive only the ABI and class-hash data required by adapters. Full Sierra/CASM deployment artifacts remain available to deployment tooling but are no longer included in the browser runtime. The UI preserves the same registry-derived fallback chain and wallet connectors while stabilizing the provider configuration.

## Bundle benchmarks

Clean Turbopack production builds, comparing `main` with Hyperlane packages `39.0.0` against this branch with `39.1.0`:

| Route | Previous raw | New raw | Raw reduction | Previous gzip | New gzip | Gzip reduction |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `/` | 126.22 MiB | 45.64 MiB | 63.8% | 23.62 MiB | 7.71 MiB | 67.4% |
| `/embed` | 128.04 MiB | 47.45 MiB | 62.9% | 23.95 MiB | 8.05 MiB | 66.4% |
| `/blocked` | 100.86 MiB | 27.70 MiB | 72.5% | 18.05 MiB | 5.02 MiB | 72.2% |

The route budgets now cap `/` at 47 MiB raw / 8.2 MiB gzip, `/embed` at 49 MiB / 8.6 MiB, and `/blocked` at 29 MiB / 5.4 MiB.

## Validation completed

- formatting
- linting (no errors; three existing `no-console` warnings)
- TypeScript typecheck
- 254 unit tests
- production build and bundle-budget check
- all 76 Chromium production E2E tests, including Radix and Starknet auto-connect and wallet-modal flows


Xaroz:
@claude review the changes, look for regression, bugs and see if benchmarks are correct

scripts/check-bundle-size.mjs
```diff
@@ -4,9 +4,9 @@ import { gzipSync } from 'node:zlib';
 const STATS_PATH = '.next/diagnostics/route-bundle-stats.json';
 const MIB = 1024 * 1024;
 const ROUTE_BUDGETS = {
-  '/': { raw: 130 * MIB, gzip: 25 * MIB },
-  '/embed': { raw: 132 * MIB, gzip: 25.5 * MIB },
-  '/blocked': { raw: 104 * MIB, gzip: 19 * MIB },
+  '/': { raw: 47 * MIB, gzip: 8.2 * MIB },
+  '/embed': { raw: 49 * MIB, gzip: 8.6 * MIB },
+  '/blocked': { raw: 29 * MIB, gzip: 5.4 * MIB },
 };
```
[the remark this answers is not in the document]
Xaroz: Fixed in `8f98d082`. Bundle limits now pass through `mibToBytes`, which rounds every threshold to integer bytes without changing the displayed MiB budgets.

Xaroz:
## Review follow-up

Implemented the actionable review suggestions in `8f98d082`:

- **Radix client lifetime:** replaced `useMemo` with a lazy `useState` initializer, so the retained toolkit and gateway client are component state rather than a disposable memoized value.
- **Static Radix configuration:** moved the unchanged toolkit options to module scope and kept client construction in a small factory.
- **Bundle-budget precision:** converted MiB limits through `mibToBytes`, explicitly rounding every threshold to integer bytes. The effective limits and benchmark output remain unchanged.

Validation completed after the changes:

- formatting, linting, and TypeScript typecheck
- 254 unit tests
- production build and bundle-budget check
- 8 production Chromium wallet tests covering Radix auto-connect, Radix overlay, Starknet modal, and the other protocol modals

Xaroz:
@claude follow-up review

src/features/wallet/context/RadixWalletContext.tsx
```diff
@@ -7,22 +7,31 @@ import {
...
 
-  const gatewayApi = GatewayApiClient.initialize(rdt.gatewayApi.clientConfig);
+  return {
+    gatewayApi: GatewayApiClient.initialize(rdt.gatewayApi.clientConfig),
+    rdt,
+  };
+}
+
+export function RadixWalletContext({ children }: PropsWithChildren<unknown>) {
+  const [{ gatewayApi, rdt }] = useState(createRadixClients);
```
paulbalaji: Medium: `useState(createRadixClients)` does not guarantee one construction here. This app enables React StrictMode, and React 19 invokes function state initializers twice in development. `RadixDappToolkit` installs subscriptions/listeners and exposes `destroy()`, so the discarded first instance cannot be disposed here and can duplicate wallet events—the lifecycle issue this change aims to prevent. Use a lifecycle-safe retained resource/singleton pattern and add StrictMode coverage proving only one live toolkit (or disposal of every discarded instance).

paulbalaji (commented):
Reviewed exact head `8f98d082a87a6956aeed97b7f527c31b56b2d0ed`. The 39.1.0 bundle reduction reproduces the PR measurements exactly (`/`: 45.64/7.71 MiB raw/gzip; `/embed`: 47.45/8.05; `/blocked`: 27.70/5.02), tightened budgets pass, the Starknet imports/memoization look correct, and exact-head CI/E2E plus 254 unit tests are green. One Radix lifecycle issue remains inline.

src/features/wallet/context/RadixWalletContext.tsx
Xaroz: Fixed in `4188ddfe`.

Radix clients now come from a lazy page-lifetime singleton, so Strict Mode's repeated renders reuse the same toolkit instead of constructing a discarded listener-owning instance. Added focused coverage proving repeated acquisition creates one toolkit and gateway client. Unit tests, production build and bundle budgets, and both Radix production-browser tests pass.

paulbalaji (approved):
Re-reviewed exact head `4188ddfed7f6ecb1b1831d64e2ae99b1a280c062`. The Radix page-lifetime singleton resolves the Strict Mode duplicate-construction issue, has focused regression coverage, and is not reached during SSR. Local focused tests, typecheck, and lint pass; exact-head CI is green with 255 unit tests, production build/bundle budgets, security checks, wallet smoke, and both E2E shards. No remaining findings.

PR #1179, opened 2026-08-03
feat: update Nexus transfer analytics events

Xaroz:
## Summary

Updates Nexus transfer analytics for the engine-owned transfer flow so swap/bridge routes preserve both selected token identities and all existing CTA/event paths stay wired.

- Emit token fields as `address|symbol` via `originToken` and `destinationToken` instead of the old single source-token shape.
- Keep `Token Selected` wired from token selection using the selected token plus the current opposite-side token; skip until both sides are available so the event stays complete.
- Keep chain selection tracking wired from the unified token/chain modal.
- Keep wallet connection initiated tracking wired from both the protocol modal and connect-aware submit CTA, and wallet connected tracking from the wallet connection observer.
- Restore `Transaction Submitted` from the transfer execution path after the origin route tx confirms.
- Avoid double-emitting submitted + failed analytics if post-submission bookkeeping throws.
- Keep `transactionHash` as the confirmed origin tx hash, which matches the existing dbt join path.
- Keep `recipient` on submitted and failed transaction events.
- Track `Unsupported Route Selected` once per token pair only after the raw quote response returns `routes: []`.
- Track `Transaction Submission Failed` for both form validation failures and transfer execution/broadcast failures.
- Add focused tests for the packed chain/token analytics helpers.

## Validation

- `pnpm format`
- `pnpm vitest src/features/analytics/utils.test.ts src/features/transfer/engine/useTransfer.test.ts --watch false`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm test`

### DIFF
--- a/src/features/analytics/types.ts
+++ b/src/features/analytics/types.ts
@@ -34,8 +34,8 @@ export type EventProperties = {
   };
   [EVENT_NAME.TRANSACTION_SUBMITTED]: {
     chains: string;
-    tokenAddress: string;
-    tokenSymbol: string;
+    originToken: string;
+    destinationToken: string;
     amount: string;
     walletAddress: string;
     transactionHash: string;
@@ -51,8 +51,8 @@ export type EventProperties = {
   };
   [EVENT_NAME.TRANSACTION_SUBMISSION_FAILED]: {
     chains: string;
-    tokenAddress: string;
-    tokenSymbol: string;
+    originToken: string;
+    destinationToken: string;
     amount: string;
     walletAddress: string | null;
     recipient: string;
--- a/src/features/analytics/utils.test.ts
+++ b/src/features/analytics/utils.test.ts
@@ -1,0 +1,18 @@
+import { describe, expect, test } from 'vitest';
+
+import { getAnalyticsChains, getAnalyticsToken } from './utils';
+
+describe('analytics event helpers', () => {
+  test('packs chains in dbt-compatible order', () => {
+    expect(
+      getAnalyticsChains(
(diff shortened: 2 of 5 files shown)


Xaroz:
@claude review

Xaroz:
@claude thanks, addressed the concrete UI fixes in 969f4c36.

- Prevented double-emitting `Transaction Submitted` + `Transaction Submission Failed` by skipping failed analytics after submitted analytics has fired.
- Changed unsupported-route dedupe from a single last key to a `Set`, so it is actually once per token pair for the form session.
- Renamed the chain helper params to avoid token/chain naming confusion.

Keeping `Unsupported Route Selected` keyed to raw `routes: []` intentionally: this event is meant to mean the engine returned no route. Security-filtered routes are a different case and I do not want to mix unsafe-route filtering into the unsupported-route funnel metric.

Also keeping `Transaction Submitted` post-confirm intentionally because dbt joins through the confirmed origin tx hash. The 8-property ceiling is noted; we are at the limit by design to keep `recipient` plus both selected token identities.

xeno097 (approved):

paulbalaji (approved):
Reviewed exact head `969f4c36beffc4fb3b936ea0dac7c1ea01520e00`. The packed chain/token fields match the companion DBT parsing contract, submitted/failed events are emitted at the correct execution boundaries, and unsupported-route tracking uses the raw quote result with per-pair deduplication. Local exact-head validation passed: 255 tests, typecheck, lint, production build, and bundle checks. No actionable findings.

PR #1184, opened 2026-08-07
fix: preserve disabled chain gate

Xaroz:
Fixes the engine-backed transfer flow so local disabled-chain metadata is still honored when `config.shouldDisableChains` is enabled.

Changes:
- Derive disabled state for `/v1/chains` entries from registry chain metadata instead of hardcoding `disabled: false`.
- Ignore disabled-chain tokens during transfer initial-value hydration so deep links/defaults do not prefill unavailable chains.
- Enforce the same disabled-chain check in form validation before quote/send.
- Add focused tests for engine-returned disabled chains in picker mapping, hydration, and validation.

Validation:
- `pnpm vitest --run src/features/transfer/engine/useFormInitialValues.test.ts src/features/chains/hooks.test.ts src/features/transfer/engine/validate.test.ts`
- `pnpm test`
- `pnpm typecheck`
- `pnpm lint` (passes with existing mock-wallet console warnings)
- `pnpm build`

### DIFF
--- a/src/features/transfer/engine/useFormInitialValues.ts
+++ b/src/features/transfer/engine/useFormInitialValues.ts
@@ -1,9 +1,13 @@
+import type { MultiProtocolProvider } from '@hyperlane-xyz/sdk';
 import { useMemo } from 'react';
 
 import { WARP_QUERY_PARAMS } from '../../../consts/args';
 import { config } from '../../../consts/config';
 import { getQueryParams } from '../../../utils/queryParams';
+import { useMultiProvider } from '../../chains/hooks';
+import { isChainDisabled } from '../../chains/utils';
 import { useTokens } from '../../tokens/hooks';
+import type { UiToken } from '../../tokens/types';
 import type { TransferFormValues } from './types';
 
 const EMPTY_VALUES: TransferFormValues = {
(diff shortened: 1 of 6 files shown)


Xaroz:
Botanix is disabled [here](https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/blob/d6ef01a16ebbd61ff0afe265ddc2ffe8a0d1f322/src/consts/chains.ts#L65) on nexus so it will still show up in the current preview

src/features/transfer/engine/useFormInitialValues.ts
```diff
@@ -41,15 +45,17 @@ export function useFormInitialValues(): TransferFormValues {
     const { originId, destinationId } = readInitialIds();
     const origin = originId ? tokens.find((t) => matchesId(t, originId)) : undefined;
     const destination = destinationId ? tokens.find((t) => matchesId(t, destinationId)) : undefined;
+    const enabledOrigin = isTokenChainEnabled(origin, multiProvider);
+    const enabledDestination = isTokenChainEnabled(destination, multiProvider);
```
paulbalaji: **Low — cover the disabled deep-link/default hydration path**

This is one of the three behaviors the PR is intended to restore, but neither new suite exercises it: `hooks.test.ts` covers picker mapping and `validate.test.ts` covers submit-time rejection. Please add a focused regression test that supplies disabled origin and destination tokens from URL/default IDs and asserts these fields remain empty. Otherwise this hydration guard could regress while the current tests continue passing. Example-based coverage is sufficient here; property-based testing would not add much for this small boolean boundary.

paulbalaji (approved):
Reviewed exact head `1b81f19b`. The fix consistently restores the existing disabled-chain policy in engine chain mapping, initial token hydration, Continue validation, and the pre-broadcast Send validation. Exact-head build, typecheck, lint, format, unit/security checks, smoke E2E, and both full E2E shards are green. One non-blocking hydration-test gap is inline; no correctness or security blockers found. Note that the Nexus Vercel deployment was ignored, so the next main-to-nexus sync still needs Nexus-specific runtime verification.

src/features/transfer/engine/useFormInitialValues.ts
Xaroz: Good call, this is valid. The picker mapping and submit validation are covered, but the hydration branch itself is not. I’ll add a focused regression for `useFormInitialValues` that mocks disabled origin/destination tokens from the initial IDs and asserts the chain/token fields stay empty.

Xaroz: Done in the updated commit. I went with the helper-extraction path so the hydration logic is covered directly without adding React test utilities: `getInitialValuesFromTokens` now has a focused regression for disabled origin/destination initial IDs staying empty.

src/features/transfer/engine/useFormInitialValues.ts
```diff
@@ -90,6 +106,15 @@ function matchesId(t: { chainName: string; address: string }, id: string): boole
   return normalizeTokenAddress(t.address) === normalizeTokenAddress(address);
 }
 
+function isTokenChainEnabled<T extends { chainName: string }>(
+  token: T | undefined,
+  multiProvider: Pick<MultiProtocolProvider, 'tryGetChainMetadata'>,
+): T | undefined {
+  if (!token) return undefined;
+  const metadata = multiProvider.tryGetChainMetadata(token.chainName);
+  return isChainDisabled(metadata ?? null) ? undefined : token;
```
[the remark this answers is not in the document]
paulbalaji: This would make hydration inconsistent with the engine-driven picker and validation. Missing provider metadata is not equivalent to a disabled chain: `toTransferChainInfo` deliberately falls back to the engine display/protocol data and `isChainDisabled(null)` remains false, preserving permissionless engine-discovered chains. We should reject only metadata explicitly marked `Disabled`; the current implementation does that consistently.

paulbalaji (approved):
Rereviewed exact head `b82b0a59`. The prior hydration-test gap is resolved: `getInitialValuesFromTokens` is extracted and directly covers disabled origin/destination IDs; the Botanix metadata override is restored with domain ID 3637. The later missing-metadata suggestion is not applicable because engine-discovered permissionless chains intentionally remain enabled without provider metadata, consistently across picker, hydration, and validation. Exact-head build, tests, typecheck, lint, format, security review, smoke E2E, both full E2E shards, and report merge are green. No new findings. Nexus-specific Vercel remains ignored; verify after the next main-to-nexus sync.

PR #1186, opened 2026-08-10
feat: classify submitted Nexus transactions as bridge or swap

Xaroz:
## Summary

- classify submitted Nexus transactions using `isBridgeOnlyRoute`
- record bridge or swap in the existing packed `chains` analytics field to stay within the Vercel property limit
- classify swap-only and swap-plus-bridge routes as `swap`

### DIFF
--- a/src/features/analytics/utils.test.ts
+++ b/src/features/analytics/utils.test.ts
@@ -12,6 +12,16 @@ describe('analytics event helpers', () => {
     ).toBe('arbitrum|42161|base|8453');
   });
 
+  test('packs transaction type after the existing chain fields', () => {
+    expect(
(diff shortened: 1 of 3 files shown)


paulbalaji (approved):
Clean at exact head 705214fda5ef2c9ba67d7c2f2e568e8712f37458. The route classifier matches the paired hyperlane-dbt #81 contract: bridge-only routes emit bridge, while routes containing a swap emit swap; appending the value preserves existing four-field consumers. Local exact-head validation passed 294 tests, typecheck, lint with zero errors, formatting, and diff checks. GitHub build, tests, typecheck, lint, format, security, smoke E2E, and both full E2E shards are green. No actionable findings.

PR #1191, opened 2026-08-14
perf: defer Solana wallet adapters

Xaroz:
## Summary

- Load the seven Solana wallet adapters only when a user chooses Solana.
- Preserve auto-connect by loading adapters on startup only when a prior Solana wallet selection exists.
- Route every app-level connect entry point through the deferred activation flow.
- Keep E2E mock-wallet behavior unchanged.

## Benchmarks

Production build, root route:

| Metric | Before | After | Delta |
| --- | ---: | ---: | ---: |
| First-load JS, raw | 15,755,499 B | 15,693,017 B | -62,482 B (-0.40%) |
| First-load JS, gzip | 4,057,487 B | 4,047,745 B | -9,742 B (-0.24%) |
| First-load chunks | 53 | 53 | unchanged |

The deferred Solana adapter chunk is 183,367 B raw / 43,848 B gzip. In a clean browser session it was absent from the initial page load; selecting Solana requested the chunk and then displayed Phantom, Backpack, Solflare, Salmon, Connect by Drift, Trust, and Ledger.

## Validation

- pnpm typecheck
- pnpm lint (passes with two pre-existing E2E mock console warnings)
- pnpm test (294 tests)
- E2E_USE_PROD=1 pnpm exec playwright test tests/wallet-connect/protocol-wallet-modals.spec.ts --project=chromium --reporter=list (2 tests)
- pnpm build
- pnpm check:bundle
- Manual production-browser verification of clean load and deferred Solana wallet modal

No dependencies, package manifests, lockfiles, or unrelated documentation files changed. Cosmos adapter startup remains a separate follow-up because the current CosmosKit provider snapshots its wallet list at mount.

### DIFF
--- a/src/components/buttons/ConnectAwareSubmitButton.tsx
+++ b/src/components/buttons/ConnectAwareSubmitButton.tsx
@@ -1,15 +1,13 @@
 import { ProtocolType } from '@hyperlane-xyz/utils';
 import { useTimeout } from '@hyperlane-xyz/widgets';
-import {
-  useAccountForChain,
-  useConnectFns,
-} from '@hyperlane-xyz/widgets/walletIntegrations/multiProtocol';
+import { useAccountForChain } from '@hyperlane-xyz/widgets/walletIntegrations/multiProtocol';
 import { useFormikContext } from 'formik';
 import { useCallback } from 'react';
 
 import { EVENT_NAME } from '../../features/analytics/types';
 import { trackEvent } from '../../features/analytics/utils';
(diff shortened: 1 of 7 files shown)


paulbalaji (commented):
Confirmed the existing inline concern: a rejected adapter-chunk import cannot reach onWalletsLoadError and falls into the app-wide fatal boundary, with reload as the only recovery. Successful lazy loading, modal activation, build, bundle limits, static checks, and exact-head E2E are otherwise clean.

PR #1192, opened 2026-08-14
perf: virtualize token picker rows

Xaroz:
## Summary

- Replace the token picker hard cap with dependency-free fixed-row virtualization.
- Render only the visible rows plus four-row overscan while retaining the full scroll height.
- Reset and remeasure the window when token search, chain filters, item count, or viewport size changes.
- Keep every returned token browsable instead of hiding tokens after the first 50.
- Teach the shared E2E picker helper to scroll through virtualized rows before selecting a token.

## Benchmarks

Automated 200-token production-browser scenario:

| Metric | Before | After | Delta |
| --- | ---: | ---: | ---: |
| Mounted token rows | 50 | fewer than 20 | at least 62% fewer |
| Tokens reachable without filtering | 50 | 200 | +300% |

The production test scrolls to the end, verifies TOK199 is visible, and confirms the mounted row count remains below 20 at both the top and bottom.

On the default local production dataset, browser inspection measured 20 mounted rows before and 11 at the top / 10 after scrolling with virtualization, a 45–50% reduction. The picker layout and row appearance were also checked visually.

Bundle size remains effectively unchanged, as expected for a rendering optimization: the root route passes at 15.03 MiB raw / 3.88 MiB gzip.

## Implementation

The custom windowing hook is 91 lines and uses only React plus the browser `ResizeObserver` API. Rows already have a fixed 60 px height and 8 px gap, so the 68 px virtual size does not require runtime row measurement or a third-party virtualization package.

No dependencies were added. `package.json` and `pnpm-lock.yaml` are unchanged, so this PR adds no package-install or supply-chain exposure.

The E2E failures were caused by tests locating Solana USDC before its row was mounted. The shared picker helper now scans overlapping virtualized viewports until the requested rendered button appears, then selects that exact button. This preserves the existing token and route semantics without adding production effects, state, or memoization.

## Commits

- `95683949` — `perf: virtualize token picker rows`
- `40b1982d` — `test: support virtualized token selection in e2e`

## Validation

- `pnpm typecheck`
- `pnpm lint` (passes with two pre-existing E2E mock console warnings)
- `pnpm test` (299 tests)
- `pnpm exec playwright test tests/token-selection tests/chain-selection --project=chromium --reporter=list` (19 tests)
- `E2E_USE_PROD=1 pnpm exec playwright test tests/token-selection/virtualized-list.spec.ts --project=chromium --reporter=list`
- Eight affected and adjacent wallet E2E scenarios together (8 passed)
- Three key E2E scenarios repeated twice (6 passed)
- Full wallet E2E suite: all picker/wallet scenarios passed; one unrelated amount-entry test flaked, then passed twice in isolation
- `pnpm build`
- `pnpm check:bundle`
- Local browser verification: Solana initially mounted 11 rows with USDC absent; scrolling mounted and selected USDC, then closed the picker with the Solana USDC route applied

Only the token-list implementation, directly related tests, and the shared picker test helper are included.


paulbalaji (commented):
Production virtualization looks sound. I reproduced keyboard-only Tab traversal through all 200 mocked tokens, reaching TOK199 without pointer input, so I could not reproduce that inline concern. The other existing inline is valid: the shared E2E helper can false-pass if the picker closes before the requested token is found or clicked. All exact-head CI and focused local validation are green.

PR #1193, opened 2026-08-14
perf: defer Cosmos wallet modal

Xaroz:
## Summary

- `perf:` replace the Chakra-wrapped CosmosKit provider with the lite provider
- `perf:` load CosmosKit's existing wallet modal and selected-wallet context only when the Cosmos picker opens
- `test:` cover the header-to-Cosmos wallet picker flow and its wallet options

## Benchmarks

Production `/` bundle, using the same checkout and configuration:

| Variant | Raw JS | Gzip JS | Chunks |
| --- | ---: | ---: | ---: |
| `main` | 15,755,499 B | 4,072,660 B | 53 |
| This PR | 14,590,298 B | 3,739,672 B | 54 |
| Change | -1,165,201 B (-7.40%) | -332,988 B (-8.18%) | +1 |

A static-modal lite-provider variant saved only 91,125 gzip bytes (-2.24%), so the conditional split accounts for another 241,863 gzip bytes without replacing CosmosKit's wallet behavior.

`@cosmos-kit/react-lite@2.18.1` was already in the lockfile transitively; this PR promotes the same pinned version to a direct dependency and adds no new package implementation.

## Verification

- production build and bundle limits pass
- TypeScript passes
- 294 unit tests pass
- three Chromium protocol-wallet modal tests pass, including the new Cosmos regression
- verified in the rendered browser: Cosmos wallet list, missing-Keplr install state, back navigation, close, and reopen
- no wallet was connected and no transaction was submitted

### DIFF
--- a/src/features/wallet/context/CosmosWalletContext.tsx
+++ b/src/features/wallet/context/CosmosWalletContext.tsx
@@ -1,13 +1,14 @@
-import { ChakraProvider, extendTheme } from '@chakra-ui/react';
 import { GasPrice } from '@cosmjs/stargate';
+import type { WalletModalProps } from '@cosmos-kit/core';
 import { wallets as cosmostationWallets } from '@cosmos-kit/cosmostation';
 import { wallets as keplrWallets } from '@cosmos-kit/keplr';
 import { wallets as leapWallets } from '@cosmos-kit/leap';
-import { ChainProvider } from '@cosmos-kit/react';
+import { ChainProvider } from '@cosmos-kit/react-lite';
 import { cosmoshub } from '@hyperlane-xyz/registry';
 import { MultiProtocolProvider } from '@hyperlane-xyz/sdk';
 import { getCosmosKitChainConfigs } from '@hyperlane-xyz/widgets/walletIntegrations/cosmos';
 import '@interchain-ui/react/styles';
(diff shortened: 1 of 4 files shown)


src/features/wallet/context/CosmosWalletContext.tsx
```diff
@@ -17,12 +18,29 @@ import { E2EAutoConnectCosmos } from '../_e2e/E2EAutoConnectCosmos';
...
+    return function CosmosKitWalletModal(props: WalletModalProps) {
+      return (
+        <SelectedWalletRepoProvider>
+          <DefaultModal {...props} />
+        </SelectedWalletRepoProvider>
+      );
+    };
   },
-});
+  { ssr: false },
```
paulbalaji, line 37: Next's production dynamic loader renders null after a rejected chunk load and retains the failed subscription until retry. Here the Cosmos view is already open, so a stale or transient chunk failure leaves no modal; reopening reuses the rejection and only a page reload recovers. Could we surface the error and expose retry (or close the view) from a loading handler?

paulbalaji (commented):
Normal Cosmos modal lifecycle, dependency resolution, bundle gate, tests, and exact-head CI are clean. One lazy-load recovery issue inline.

PR #1194, opened 2026-08-14
fix: resolve SVM synthetic balances and route display

Xaroz:
## Summary

- derive Sealevel synthetic Token-2022 mints from Warp router program IDs before reading balances
- use the SDK Sealevel token adapter for single-token balance validation while preserving batched UI reads
- show the selected destination token for terminal cross-chain route nodes, fixing Aleo → Solana ALEO routes displaying SOL
- add focused regression coverage for both issues

### DIFF
--- a/src/features/balances/sealevel.ts
+++ b/src/features/balances/sealevel.ts
@@ -1,4 +1,9 @@
-import { TokenStandard, type MultiProtocolProvider } from '@hyperlane-xyz/sdk';
+import {
+  BaseSealevelAdapter,
+  SealevelTokenAdapter,
+  TokenStandard,
+  type MultiProtocolProvider,
+} from '@hyperlane-xyz/sdk';
 import {
   AccountLayout,
   ASSOCIATED_TOKEN_PROGRAM_ID,
@@ -14,6 +19,7 @@ import type { BalanceToken } from './types';
 import { getBalanceTokenKey } from './types';
 
 const ZERO_ADDRESS = /^0x0+$/i;
+const HYP_SYNTHETIC_MINT_SEEDS = ['hyperlane_token', '-', 'mint'];
 
 export async function fetchSealevelChainBalances(
   rpcUrl: string,
@@ -31,6 +37,29 @@ export async function fetchSealevelChainBalances(
       nativeTokens.push(token);
       continue;
     }
+
+    if (token.standard === TokenStandard.SealevelHypSynthetic) {
+      try {
+        const mint = deriveSealevelHypSyntheticMint(token.address);
+        splEntries.push({
+          token,
+          ata: getAssociatedTokenAddressSync(
+            mint,
+            ownerKey,
+            true,
+            TOKEN_2022_PROGRAM_ID,
+            ASSOCIATED_TOKEN_PROGRAM_ID,
+          ),
+        });
+      } catch (err) {
+        logger.warn(
+          `Could not resolve synthetic balance account for ${token.symbol}`,
+          err as Error,
+        );
+      }
+      continue;
+    }
+
     let mint: PublicKey;
     try {
       mint = new PublicKey(token.address);
@@ -86,7 +115,17 @@ export async function readSealevelTokenBalance(
   ) {
     return BigInt(await connection.getBalance(ownerKey));
   }
+  if (args.standard === TokenStandard.SealevelHypSynthetic) {
+    const adapter = new SealevelTokenAdapter(args.chainName, multiProvider, {
+      token: deriveSealevelHypSyntheticMint(args.tokenAddress).toBase58(),
+    });
+    return adapter.getBalance(args.owner);
+  }
   return fetchSplBalance(connection, ownerKey, args.tokenAddress);
+}
+
+export function deriveSealevelHypSyntheticMint(warpRouter: string): PublicKey {
+  return BaseSealevelAdapter.derivePda(HYP_SYNTHETIC_MINT_SEEDS, warpRouter);
 }
 
 export function isSealevelNativeBalance(token: {
--- a/src/features/balances/sealevel.ts
+++ b/src/features/balances/sealevel.ts
@@ -104,6 +104,13 @@ export async function readSealevelTokenBalance(
 ): Promise<bigint> {
   const rpcUrl = multiProvider.tryGetChainMetadata(args.chainName)?.rpcUrls?.[0]?.http;
   if (!rpcUrl) throw new Error(`Missing Sealevel RPC URL for ${args.chainName}`);
+  if (args.standard === TokenStandard.SealevelHypSynthetic) {
+    const adapter = new SealevelTokenAdapter(args.chainName, multiProvider, {
+      token: deriveSealevelHypSyntheticMint(args.tokenAddress).toBase58(),
+    });
+    return adapter.getBalance(args.owner);
+  }
+
   const connection = new Connection(rpcUrl, 'confirmed');
   const ownerKey = new PublicKey(args.owner);
   if (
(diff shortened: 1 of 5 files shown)


Xaroz:
@claude review

src/features/balances/sealevel.ts
```diff
@@ -31,6 +37,29 @@ export async function fetchSealevelChainBalances(
...
+    if (token.standard === TokenStandard.SealevelHypSynthetic) {
+      try {
+        const mint = deriveSealevelHypSyntheticMint(token.address);
+        splEntries.push({
+          token,
+          ata: getAssociatedTokenAddressSync(
+            mint,
+            ownerKey,
+            true,
+            TOKEN_2022_PROGRAM_ID,
```
haggis-hyperlane, line 50: Non-blocking: the batched path hardcodes `TOKEN_2022_PROGRAM_ID` for synthetics, while `readSealevelTokenBalance` below routes through `SealevelTokenAdapter`, which detects the token program dynamically from the mint owner (`SealevelTokenAdapter.ts:364-371`). This matches the SDK's `SealevelHypSyntheticAdapter`, which also assumes Token-2022, so it's correct for all current Hyperlane Sealevel synthetics — just noting that a hypothetical classic-SPL synthetic would be found by the single-token path but missed here.

src/features/balances/sealevel.ts
```diff
@@ -86,9 +115,19 @@ export async function readSealevelTokenBalance(
...
+    const adapter = new SealevelTokenAdapter(args.chainName, multiProvider, {
+      token: deriveSealevelHypSyntheticMint(args.tokenAddress).toBase58(),
+    });
+    return adapter.getBalance(args.owner);
+  }
   return fetchSplBalance(connection, ownerKey, args.tokenAddress);
 }
 
+export function deriveSealevelHypSyntheticMint(warpRouter: string): PublicKey {
+  return BaseSealevelAdapter.derivePda(HYP_SYNTHETIC_MINT_SEEDS, warpRouter);
```
haggis-hyperlane, line 129: Minor: this re-derives what the SDK already exposes as `SealevelHypSyntheticAdapter.deriveMintAuthorityAccount()` (seeds `['hyperlane_token','-','mint']`). Fine to keep for a light dependency, but reusing the SDK helper would avoid drift if those seeds ever change.

haggis-hyperlane (commented):
## Review Summary

Reviewed at `9cc438929bcbcf65db32daa51a08c2f739cf8b80`.

The balance fix is correct and matches the diagnosed root cause. Prior to this PR, `fetchSealevelChainBalances` treated the engine's token `address` as the SPL mint, but for `SealevelHypSynthetic` tokens the engine returns the **warp-route program ID** (`wrappedAddress: null`), not a mint — so ATAs were derived against a non-mint and every SVM synthetic balance (Eclipse, Solana, …) read as 0/missing.

I verified the fix end-to-end:
- The derivation seeds `['hyperlane_token','-','mint']` and `TOKEN_2022_PROGRAM_ID` exactly match the SDK's `SealevelHypSyntheticAdapter.deriveMintAuthorityAccount()` / `deriveAssociatedTokenAccount()` (`typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts:1634-1650`).
- On-chain check for BLEND/solanamainnet: program `4iQfUG5…` is owned by the BPF Upgradeable Loader (a program, not a mint); the derived mint PDA = `BWsnyEa1XtsNRdgPaDoA1WVUonF7BBGZTd2zc72NQsWT`, owned by the Token-2022 program (a real mint). The PR's `deriveSealevelHypSyntheticMint` reproduces exactly this.

The route-display fix is coherent too: `buildFlowNodes` and `CrossChainEdge` now fall back to the user-selected `dstToken` for the terminal destination instead of `step.asset` (origin asset), which is what caused Aleo→Solana ALEO to render as SOL. Tests cover both terminal and non-terminal bridge cases.

Test coverage is focused and correct. No blocking issues — the notes below are non-blocking.

## Notes

- The single-token path (`readSealevelTokenBalance`) uses the base `SealevelTokenAdapter`, which dynamically detects the mint's owning token program (`SealevelTokenAdapter.ts:364-371`), whereas the batched path hardcodes `TOKEN_2022_PROGRAM_ID`. Both are correct today since all Hyperlane Sealevel synthetics mint under Token-2022 — just flagging the asymmetry (see inline).

src/features/balances/sealevel.ts
Xaroz: Confirmed and covered in d29fb754. The regression test now exercises SealevelTokenAdapter end-to-end: it reads the derived mint owner, selects Token-2022, and queries the expected ATA. The batched path keeps the SDK synthetic invariant and preserves one batched RPC read.

src/features/balances/sealevel.ts
Xaroz: Keeping BaseSealevelAdapter.derivePda here. SealevelHypSyntheticAdapter requires mailbox plus the resolved mint; mailbox lives separately in store chainAddresses rather than MultiProtocolProvider metadata, and the mint is the value being derived. Threading both into balance reads solely for this instance helper would add circular coupling. A standalone SDK helper would be the clean upstream improvement.

Xaroz:
@claude Review follow-up in d29fb754: confirmed the SDK base token adapter dynamically reads the derived mint owner and selects Token-2022. The regression test now exercises that real path and asserts the exact ATA; the unused Connection allocation was also removed. The full Hyp synthetic adapter was intentionally skipped because its mailbox and resolved-mint inputs are not carried by MultiProtocolProvider metadata in this engine-driven balance path.

xeno097 (approved):

src/features/transfer/engine/routeSelection/RouteSelectionModal.tsx
```diff
@@ -163,7 +163,7 @@ function RouteFlowDiagram({
   // picked Base→Viction) are never fetched by the token picker, so
   // tokens like WETH on Arb would otherwise stay unresolved.
   useRouteChainTokens(steps);
-  const nodes = buildFlowNodes(steps);
+  const nodes = buildFlowNodes(steps, { destinationTokenAddress: dstToken?.address });
```
paulbalaji, line 166: `dstToken.address` is the native zero-address sentinel for native selections. `buildFlowNodes` then makes it the terminal bridge node, while `showUnwrap` below only matches the trusted wrapped-native address, so terminal native routes lose the explicit Unwrap edge. Could we use `trustedWrappedNativeAddressForToken(dstToken) ?? dstToken?.address` here and cover a terminal native destination?

src/features/transfer/engine/routeSelection/RouteSelectionModal.tsx
```diff
@@ -477,7 +487,7 @@ function CrossChainEdge({
       ? getTokenByKeyFromMap(tokenMap, tokenKey(step.destChain, nextStep.tokenIn))
       : nextStep?.type === 'bridge'
         ? getTokenByKeyFromMap(tokenMap, tokenKey(step.destChain, nextStep.asset))
-        : asset;
+        : (resolvedDestAsset ?? asset);
 
   const amountIn = formatStepAmount(step.amountIn, asset?.decimals ?? 18);
   const amountOut = formatStepAmount(step.amountOut, destAsset?.decimals ?? asset?.decimals ?? 18);
```
paulbalaji, line 493: `amountOut` now correctly uses destination decimals, but its tooltip still appends the source `symbol` below. For bridges whose destination representation has a different symbol, the value is mislabeled. Could Amount out use `destAsset?.symbol ?? symbol`?

paulbalaji (commented):
Post-merge follow-up review: balance resolution and the current SDK invariants look sound. Two small route-display follow-ups inline.

PR #1196, opened 2026-08-17
fix: don't double-count same-token IGP fee in balance validation

haggis-hyperlane:
## Summary
- Bridge transfers on **native-token warp routes** (e.g. GNET galactica→solana) were falsely blocked with `Insufficient <TOKEN> balance (need X more)` even when the wallet held enough.
- Root cause: for native routes the interchain gas fee is paid in the **same** token being sent. The engine already nets that fee out of the output (`amountOut = amountIn − igp`) and sets `tx.value = amountIn`, so the fee is **already included in `amountIn`**. `validateBalances` added the same-token IGP again on top, over-requiring by exactly the fee amount.
- ERC20 collateral/synthetic routes were unaffected because their IGP is paid in **native** (a different token → `sameTokenIgp = 0`, validated on the separate native path). This is why prior EVM testing didn't surface it — the bug is specific to native-source (same-token-IGP) routes, independent of VM.

### Change
- `validateBalances`: require `amountIn` (already gross of same-token fees) for the source-token check instead of `amountIn + sameTokenIgp`; likewise use `amountIn` (not `amountIn + sameTokenIgp`) for `quotedNativeDebit`. The native path still bounds by the real `tx.value` via `max(txValue, ...)`.
- Added a regression test mirroring the GNET scenario (`amountOut = amountIn − igp`, `tx.value = amountIn`, balance exactly `amountIn`) that fails before the fix and passes after.

### Live confirmation
- Galactica GNET router `0x8dD4105615773b9e28a83304Ea3739ecA8C45de9`, `quoteGasPayment(solana) = 233.18 GNET`, origin `scale = 1`. Observed: Send `1224` → Receive `990.82` (= 1224 − 233.18), balance `1224.72` → false "need 232.46 more".

## Test plan
- [x] `vitest run src/features/transfer/engine/validate.test.ts` (11 passed)
- [x] Regression test verified failing pre-fix, passing post-fix
- [x] `tsc --noEmit` clean
- [x] `oxlint ./src` clean (only 2 pre-existing unrelated warnings)

### DIFF
--- a/src/features/transfer/engine/validate.ts
+++ b/src/features/transfer/engine/validate.ts
@@ -235,7 +235,6 @@ export async function validateBalances(args: {
 
   const igpByToken = aggregateIgp(bestRoute.feeBreakdown.components);
   const srcKey = balanceKey(srcToken.chainId, srcToken.address);
-  const sameTokenIgp = igpByToken.get(srcKey) ?? 0n;
   const nativeFee = igpByToken.get(balanceKey(srcChainInfo.id, NATIVE_ADDRESS)) ?? 0n;
 
   let srcBalance: bigint | null;
@@ -257,11 +256,16 @@ export async function validateBalances(args: {
     return null;
   }
 
-  if (srcBalance != null && amountIn + sameTokenIgp > srcBalance) {
+  // amountIn is the gross debit from the sender: the engine already nets any
+  // same-token fee out of the output (amountOut = amountIn − same-token fees),
+  // so an interchain fee denominated in the source token (e.g. native-token
+  // warp routes where IGP is paid in the native gas token) is already included
+  // in amountIn. Adding it again would double-count the fee.
+  if (srcBalance != null && amountIn > srcBalance) {
     return {
       amount: formatInsufficientBalanceMessage({
         base: `Insufficient ${srcToken.symbol} balance`,
-        deficit: amountIn + sameTokenIgp - srcBalance,
+        deficit: amountIn - srcBalance,
         decimals: srcToken.decimals,
         symbol: srcToken.symbol,
       }),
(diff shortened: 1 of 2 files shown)


paulbalaji (commented):
One correctness issue: the fix is valid for EVM-native routes, but the unconditional accounting change under-validates separately paid native IGP on non-EVM sdkWarp routes.

src/features/transfer/engine/validate.ts
```diff
@@ -257,11 +256,16 @@ export async function validateBalances(args: {
     return null;
   }
 
-  if (srcBalance != null && amountIn + sameTokenIgp > srcBalance) {
+  // amountIn is the gross debit from the sender: the engine already nets any
+  // same-token fee out of the output (amountOut = amountIn − same-token fees),
+  // so an interchain fee denominated in the source token (e.g. native-token
+  // warp routes where IGP is paid in the native gas token) is already included
+  // in amountIn. Adding it again would double-count the fee.
+  if (srcBalance != null && amountIn > srcBalance) {
```
paulbalaji, line 264: URE only folds native IGP into `amountIn` for direct/EVM-like source semantics. Non-EVM bridge-only routes use `sdkWarp`, and URE intentionally sets `sdkNativeIgpPaidSeparately`, leaving `igpTokenFee = 0` and `amountOut === amountIn`; the SDK transaction pays IGP separately. Since typed SDK txs are not `isChainRouteTx`, `txValue` cannot recover that debit here. With this unconditional check, a native Solana/Cosmos/Starknet wallet holding exactly `amountIn` passes preflight but lacks the IGP. Please scope the no-double-count path to routes where IGP is included, retaining `amountIn + sameTokenIgp` for non-EVM `sdkWarp`, and cover `amountOut === amountIn` with a typed SDK-tx regression.

PR #1197, opened 2026-08-17
fix: avoid double-counting embedded interchain fees

Xaroz:
## Summary

- avoid double-counting IGP already embedded in the route input amount
- validate only separately funded IGP plus source-network transaction fees
- use each max route's engine-calculated source transaction fee for validation and fee display
- estimate source fees for manually entered amounts through the SDK multi-protocol provider
- wait while quote-time allowance checks are pending; if a read fails, reserve the two-transaction classic ERC20 worst case and re-check exactly before any broadcast
- require explicit `fee.igpIncludedInAmountIn` metadata
- share the prepared transaction between fee estimation and wallet execution, including the configured Solana provider
- preserve approval, Starknet, ERC20, synthetic, swap, and multi-transaction handling

## Max flow

1. Clicking **Max** cancels any ordinary quote in flight and calls `POST /v1/quote/max` once.
2. The UI fills the returned amount and seeds the ordinary quote cache with the same executable max response, avoiding a second `/quote` call caused by the input update.
3. While the request and filled amount remain unchanged, that response stays authoritative until `expiresAt`: neither `/quote` nor `/quote/max` is refreshed, and the input is not rewritten.
4. At expiry, the UI calls ordinary `/quote` for the fixed input and resumes its normal 25-second refresh cadence. If current fees no longer fit the balance, validation blocks the transfer and the user can press **Max** again.
5. Editing the amount, token pair, recipient, or slippage exits the preserved Max intent and uses ordinary quoting.
6. Send uses the same execution path for normal and max quotes. It performs fresh exact approval/revoke checks and aborts before broadcasting if those checks cannot resolve.

## Dependencies

- hyperlane-xyz/universal-router-engine#87 provides explicit IGP funding metadata
- hyperlane-xyz/universal-router-engine#92 provides maximum quotes and per-route source transaction fees
- SDK 42.0.0 includes balance-independent fee estimation from hyperlane-xyz/hyperlane-monorepo#9313, the Solana message-fee fix from hyperlane-xyz/hyperlane-monorepo#9270, and the EIP-1559 fee fix from hyperlane-xyz/hyperlane-monorepo#9279
- ordinary Ethereum and Tron fees use the SDK estimate directly; local EVM-like pricing remains only for conservative approval/revoke gas budgets
- Starknet max quotes currently use the full source-token balance with a zero source-fee placeholder until unsigned server-side estimation is available


paulbalaji (dismissed):
Verified embedded-versus-external IGP accounting across EVM/non-EVM native routes, ERC20/router aliases, swaps, and the optional `igpTokenFee` rollout contract with URE #87. Focused tests/typecheck/format and exact-head CI, including E2E, are green.

src/features/transfer/engine/sourceFee.ts
```diff
@@ -0,0 +1,108 @@
...
+    multiProvider,
+    chainName,
+    protocol: protocol as ProtocolType,
+    sender,
+    routeTxs,
+  });
+  const publicKey = await senderPubKey;
+  const estimates = await Promise.all(
+    transactions.map((transaction) =>
+      multiProvider.estimateTransactionFee({
```
paulbalaji, line 97: This still under-reserves Solana fees. SDK 41's Solana estimator only returns `unitsConsumed * recentPrioritizationFee`; it never includes the mandatory per-signature base fee. I reproduced `fee: 0n` with zero priority fee, so a native near/full-balance route can pass `nativeRequired` and then fail in the wallet. Please fix the SDK estimator (ideally via `getFeeForMessage`) or account for the base fee here, and add a non-mocked zero-priority regression test.

src/features/transfer/engine/sourceFee.ts
```diff
@@ -0,0 +1,108 @@
...
+  // A pending approval, including SDK routes that carry approval + transfer
+  // transactions, makes the transfer impossible to simulate against current
+  // state. Preserve the existing combined gas budget for those routes.
+  const routeTxs = getRouteTxs(route);
+  if (isEVMLike(protocol) && (approvalPending || routeTxs.length > 1)) {
+    return estimateEvmLikeFeeForGasUnits(
+      multiProvider,
+      chainName,
+      EVM_LIKE_APPROVAL_ROUTE_GAS_BUDGET,
+    );
```
paulbalaji: This 600k fallback can be lower than the work we actually send. URE can budget 600k+ for the routed transfer itself, and `useTransfer` may send one or two approval transactions before it. Because this branch replaces the route estimate with a flat total, validation can pass, consume approval gas, then fail the transfer for insufficient native balance. Please reserve route execution plus approval/revoke separately (and cover a 600k+ route and revoke+approve).

paulbalaji (commented):
Re-reviewed exact head `ea471238`. Two source-fee gaps remain; comments inline.

Local focused tests (37), typecheck, lint, and diff check pass. Build/test/typecheck/lint/security CI pass. Both failed E2E jobs were canceled during `playwright install-deps chromium` after Ubuntu mirror stalls, before application tests ran, so they appear environmental but are not green evidence.

src/features/transfer/engine/sourceFee.ts
```diff
@@ -0,0 +1,108 @@
...
+  gasUnits: bigint,
+): Promise<bigint> {
+  const feeData = await multiProvider.getEthersV5Provider(chainName).getFeeData();
+  const maxFee = feeData.maxFeePerGas ? BigInt(feeData.maxFeePerGas.toString()) : undefined;
+  const priorityFee = feeData.maxPriorityFeePerGas
+    ? BigInt(feeData.maxPriorityFeePerGas.toString())
+    : undefined;
+  const legacyFee = feeData.gasPrice ? BigInt(feeData.gasPrice.toString()) : undefined;
+  const gasPrice = maxFee && priorityFee ? maxFee + priorityFee : legacyFee;
+  if (gasPrice == null) throw new Error(`No EVM-like gas price available for ${chainName}`);
```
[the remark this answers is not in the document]
Xaroz: Fixed in 2799513e: maxFeePerGas is now treated as the full EIP-1559 cap; the priority fee is no longer added twice.

src/features/transfer/engine/sourceFee.ts
Xaroz: Fixed in 2799513e: the fallback now reserves the full route gas budget plus 55k for each approve/revoke transaction. Tests cover a 700k route and revoke + approve.

src/features/transfer/engine/TransferForm.tsx
```diff
@@ -234,31 +244,67 @@ function TransferFormContent() {
...
+          senderPubKey,
+          route: bestRoute.raw,
+          approvalPending,
+        });
+      } catch (err) {
+        logger.warn('Unable to estimate source transaction fee', err as Error);
+        return {
+          form: 'Unable to estimate source transaction fee; reduce the amount and try again',
+        };
+      }
```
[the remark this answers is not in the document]
Xaroz: Fixed in 2799513e: estimation failures now ask the user to retry without incorrectly blaming the transfer amount.

src/features/transfer/engine/sourceFee.ts
Xaroz: The SDK fix is merged in hyperlane-monorepo#9270. The latest published SDK (41.2.0) predates that merge, so this remains pending until the release is available.

src/features/transfer/engine/sourceFee.ts
```diff
@@ -0,0 +1,116 @@
...
+    multiProvider,
+    chainName,
+    protocol: protocol as ProtocolType,
+    sender,
+    routeTxs,
+  });
+  const publicKey = (await senderPubKey)?.replace(/^0x/, '');
+  const estimates = await Promise.all(
+    transactions.map((transaction) =>
+      multiProvider.estimateTransactionFee({
```
paulbalaji, line 97: The local approval fallback now handles EIP-1559 correctly, but no-approval EVM routes still delegate here to SDK 41.0.0, whose `computeEvmTxFee` uses `maxFeePerGas + maxPriorityFeePerGas`. I reproduced gas price `3` / fee `1,800,000` for a 600k route with max fee `2` and priority `1`; the cap is `1,200,000`. That can reject a fundable transfer and makes the stacked Max flow reserve too much. Please fix/pin the SDK path too (or calculate EVM fee caps locally) and cover it with a non-mocked EIP-1559 regression test.

paulbalaji (commented):
Re-reviewed exact head `2799513`.

The approval fallback now preserves the route gas budget and separately counts approve/revoke transactions; that prior finding is resolved. The Solana base-fee thread remains open as acknowledged: this PR is pinned to SDK 41.0.0, and the published SDK still predates hyperlane-monorepo#9270. One additional SDK-path issue inline.

Local focused tests (40), typecheck, lint, format, and diff check pass. Exact-head build/test/typecheck/lint/format/security, smoke E2E, and E2E shard 1 pass; shard 2 is still running in Playwright system dependency installation.

src/features/transfer/engine/sourceFee.ts
Xaroz: Moved this fix to the shared SDK in [hyperlane-monorepo#9279](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/9279): EIP-1559 estimates now use `maxFeePerGas` as the full cap. This UI stays on the generic SDK estimator; I will bump it after the SDK release containing #9270 and #9279.

Xaroz: Fixed in a564e495. The UI now uses SDK-estimated gas units with `maxFeePerGas` as the EIP-1559 cap while hyperlane-monorepo#9279 is pending, with regression coverage.

src/features/transfer/engine/sourceFee.ts
Xaroz: Fixed in a564e495. Upgraded to SDK 41.3.0 (including hyperlane-monorepo#9270) and added a zero-priority Solana base-fee regression test.

Xaroz:
@claude review

paulbalaji:
@cursor review

src/features/transfer/engine/TransferForm.tsx
```diff
@@ -234,31 +254,94 @@ function TransferFormContent() {
...
+    latestMaxRequestKeyRef.current = maxRequestKey;
+  }, [maxRequestKey]);
+
+  const onMax = useCallback(async () => {
+    if (!srcToken) return;
+    const requestKey = latestMaxRequestKeyRef.current;
+    try {
+      const response = await requestMaxQuote();
+      if (latestMaxRequestKeyRef.current !== requestKey) return;
+      await setFieldValue('amount', formatUnits(BigInt(response.amount), srcToken.decimals), false);
```
paulbalaji: The max endpoint can still return an amount that this client immediately rejects when a separately funded IGP is paid in the source ERC20. URE #92 quotes the full non-native source balance and only checks native funding; it never subtracts same-token external IGP or checks another non-native fee token. I reproduced balance `1000` + source-token IGP `5` returning `1000` here (the executable amount is `995`), after which WUI's `amountIn + sameTokenIgp` validation fails. Please make `/quote/max` solve funding per token for the final route before applying `response.amount`.

src/features/transfer/engine/useQuote.ts
```diff
@@ -31,16 +31,52 @@ function randomBytes32(): Hex {
...
+    params.slippageBps ?? null,
+  ] as const;
+}
+
+export function cacheMaxQuote(
+  queryClient: QueryClient,
+  params: MaxQuoteParams,
+  response: MaxQuoteResponse,
+): void {
+  queryClient.setQueryData(quoteQueryKey({ ...params, amount: BigInt(response.amount) }), response);
```
paulbalaji: This atomic max result only survives until the 25s query refresh. The same cache entry then refetches through plain `/quote`, drops `sourceTransactionFee`, and validation switches to a fresh client estimate instead of the fee used to derive the native Max amount. If fees rise, Max fails its own validation. Preserve max intent and refresh via `/quote/max` (or retain a conservative executable fee bound) rather than replacing this response with a normal quote.

src/features/transfer/engine/useQuote.ts
```diff
@@ -309,19 +392,24 @@ function compoundSlippageMin(output: bigint, slippageBps: number, swapStepCount:
 }
 
 export function isQuoteRequestReady(v: TransferFormValues, sender: string | undefined): boolean {
+  if (!isMaxQuoteRequestReady(v, sender)) return false;
+  if (!v.amount || Number(v.amount) <= 0) return false;
+  return true;
+}
+
+export function isMaxQuoteRequestReady(v: TransferFormValues, sender: string | undefined): boolean {
```
paulbalaji, line 555: URE #92 explicitly rejects Starknet max quotes because server-side fee estimation has no connected account, but this readiness check enables Max for every protocol. That exposes an action known to fail for Starknet. Please gate the capability with clear UX or retain the connected-account estimation path for Starknet.

paulbalaji (commented):
Rereviewed the exact head and the dependent URE #92 implementation. The earlier Solana, EIP-1559, and approval-budget findings are fixed. I found three remaining Max integration issues below. WUI CI and focused local validation are green.

src/features/transfer/engine/TransferForm.tsx
Xaroz: Fixed on the engine side in [URE #92](https://github.com/hyperlane-xyz/universal-router-engine/pull/92), commit `8c01fc6`. `/quote/max` now groups external IGP by token, subtracts source-token IGP from the common amount, reads balances for other fee tokens, and filters the final quoted routes by complete per-token funding. The UI therefore only applies an executable returned amount.

src/features/transfer/engine/useQuote.ts
Xaroz: Fixed in `7635b3bc`. The UI retains Max intent for the exact request and atomic amount, so the 25-second refresh calls `/quote/max` again instead of plain `/quote`. It preserves `sourceTransactionFee`, atomically updates the input/cache if the executable maximum changes, and clears Max intent on manual edits. Regression coverage exercises the intent match.

src/features/transfer/engine/useQuote.ts
Xaroz: Fixed in `7635b3bc`. Max stays disabled until the source protocol is known, and Starknet remains disabled with an explanatory tooltip. Typed Starknet transfers retain the connected-account fee-estimation path.

Xaroz:
@claude Review follow-up

Addressed the major and actionable minor findings in `7635b3bc`:

- Max intent now survives refresh; the 25-second refresh uses `/quote/max`, retains `sourceTransactionFee`, and updates the atomic input/cache if the executable amount changes.
- Unresolved approval state reserves one approval transaction instead of simulating a transfer expected to revert.
- Max formatting uses the same token-decimals source as the quote cache key.
- Starknet Max is gated until server-side account-aware estimation exists.
- Zero-valued EVM fee data remains valid.

Confirmed no Solana change is needed: the SDK provider builder already constructs `Connection(..., "confirmed")`. I kept the category-less multi-transaction fallback as a conservative fee reserve, and kept exact byte-wrapper matching to avoid reinterpreting arbitrary objects that happen to contain `encoding` and `data`.

Validation: 342 tests, typecheck, lint, production build, format, and diff checks pass. The only lint output is two pre-existing mock-wallet `console` warnings.

src/features/transfer/engine/useQuote.ts
Xaroz: Follow-up in `aa9fad21`: `/quote/max` now runs only when Max is clicked. Its response remains authoritative until `expiresAt`; only then does the UI resume plain `/quote` refreshes for the filled amount. Ordinary quote work is canceled/paused while Max is pending, and periodic input rewrites were removed.

Xaroz:
@paulbalaji follow-up on the Max lifecycle after your review:

1. Clicking **Max** cancels any in-flight ordinary quote and calls `/quote/max` once.
2. The returned amount and routes are cached together, so filling the input does not trigger `/quote`.
3. While the request and amount stay unchanged, the max response remains authoritative through `expiresAt`; there is no periodic `/quote/max`, no ordinary `/quote`, and no second input rewrite.
4. At expiry, the UI calls ordinary `/quote` for the fixed amount and resumes normal refreshes. From that point validation uses the fresh route/fee data; if the balance no longer covers it, the user must press **Max** again.
5. A manual amount, token, recipient, or slippage change exits the preserved Max intent immediately.

Implemented in `aa9fad21`. This keeps the engine-reserved `sourceTransactionFee` for the full executable lifetime without polling the expensive Max endpoint.

src/features/transfer/engine/useQuote.ts
```diff
@@ -97,21 +181,115 @@ export function useQuote({ values, sender, pause }: UseQuoteArgs) {
...
+      amount: amountAtomic,
+      sender: engineSender,
+      recipient: engineRecipient,
+      slippageBps: values.slippageBps,
+    }),
+    queryFn: async ({ signal }) => {
+      if (activeMaxQuoteIntent && maxQuoteIntentRef.current === activeMaxQuoteIntent) {
+        // The max response remains authoritative until its expiry. This query
+        // runs at that deadline, clears max mode, then resumes normal quoting.
+        maxQuoteIntentRef.current = null;
```
paulbalaji: This moves the original self-invalidating Max case from the 25s refresh to `expiresAt`, but does not remove it. The plain `/quote` response below drops `sourceTransactionFee`; validation then estimates current source gas against the unchanged native Max amount. If gas rose since Max was calculated, that amount immediately fails its own funding check. At expiry, either refresh atomically through `/quote/max` and reconcile the amount, or explicitly mark Max stale and require recalculation instead of silently converting it to an ordinary exact-input quote.

src/features/transfer/engine/TransferForm.tsx
```diff
@@ -229,36 +248,98 @@ function TransferFormContent() {
...
+
+  const onMax = useCallback(async () => {
+    if (!srcToken || sourceTokenDecimals == null) return;
+    const requestKey = latestMaxRequestKeyRef.current;
+    try {
+      const response = await requestMaxQuote();
+      if (latestMaxRequestKeyRef.current !== requestKey) return;
+      await setFieldValue(
+        'amount',
+        formatUnits(BigInt(response.amount), sourceTokenDecimals),
```
paulbalaji, line 344: URE #92 still derives this amount from the *lowest* route-specific maximum, rather than the greatest executable route. Its exact-head test has balance `1000`, a direct route requiring `10`, and a swap route requiring `35`, yet asserts `amount === 965`; the direct route can transfer `990`. Because WUI then selects the preferred/direct route, an expensive alternative can make Max underfill arbitrarily. The endpoint should choose the greatest executable route-specific amount and filter incompatible routes, or return route-specific maxima and bind the selected route.

paulbalaji (commented):
Rereviewed exact head `aa9fad21`, the linked Claude follow-up, and current URE #92 (`43f491f`). The prior per-token funding, decimals, Starknet, Solana, EIP-1559, and approval-route-budget findings are fixed. Three remaining Max/funding semantics are inline. Exact-head WUI CI, 343 local tests, and linked URE CI are green.

src/features/transfer/engine/approval.ts
```diff
@@ -23,6 +23,15 @@ export type ApprovalPhase = (typeof ApprovalPhase)[keyof typeof ApprovalPhase];
 
 export type ApprovalStatus = { phase: ApprovalPhase };
 
+export function getApprovalTransactionCount(status: ApprovalStatus, hasApproval: boolean): number {
+  if (status.phase === ApprovalPhase.NeedsRevoke) return 2;
+  if (status.phase === ApprovalPhase.NeedsApprove) return 1;
+  // While the allowance query is unresolved, reserve one approval instead
+  // of simulating a transfer that is expected to revert without allowance.
+  if (hasApproval && status.phase === ApprovalPhase.Idle) return 1;
```
paulbalaji: `Idle` conflates the initial allowance load with a failed/unresolved allowance read, so one approval is not a conservative bound. `useTransfer` re-reads allowance and may send both revoke and approve for a nonzero insufficient USDT-like allowance. If this query stays unresolved through validation, native funding can be short by one transaction. Gate validation until allowance settles successfully, or reserve two transactions while approval state is unknown.

src/features/transfer/engine/useQuote.ts
Xaroz: I think we want it to be like this, if we call /quote/max on refresh it would mean that the input could change which in my opinion is weird UX behavior

paulbalaji (approved):
Approved on exact head aa9fad21. Prior blocking findings are fixed and CI is green. The remaining inline comments are non-blocking follow-ups/product tradeoffs.

src/features/transfer/engine/TransferForm.tsx
Xaroz: Keeping this behavior as an intentional tradeoff. `/quote/max` returns one common amount that every returned route can fund. That can underfill relative to the cheapest/preferred route, but it preserves all executable alternatives with one stable input and avoids route-bound maxima, additional final quote passes, or changing the input when a route is selected. We can revisit per-route maxima as a separate API/UI design if maximizing each route becomes more important than the simpler common-amount contract.

src/features/transfer/engine/approval.ts
Xaroz: Fixed in `7eaa7d9e`. The UI no longer guesses an approval count while the initial allowance read is unresolved. Approval routes keep Continue in the existing `Fetching quote…` state until the check returns the exact zero/one/two-transaction phase; a failed initial read blocks validation as `Approval check failed`. Execution still re-reads allowance immediately before sending.

Xaroz: @paulbalaji Follow-up in b1609dd5 and engine PR #92 commit a7a318d: pending allowance reads still gate Continue; failed quote-time reads now reserve the conservative worst case (revoke + approve, plus Permit2 in the engine when applicable) instead of failing the quote. Send performs fresh exact allowance checks for both normal and max quotes, before any wallet transaction, and aborts if they cannot resolve.

paulbalaji (approved):
Reapproved exact head b1609dd5. The two-commit allowance delta fixes the unknown-allowance under-reservation: pending checks gate validation, failures reserve revoke plus approve, and execution rechecks before broadcasting. No new findings. The remaining Max-lifecycle thread is a non-blocking product tradeoff.

src/features/transfer/engine/sourceFee.ts
```diff
@@ -0,0 +1,116 @@
...
+  approvalTransactionCount,
+}: {
+  multiProvider: MultiProtocolProvider;
+  chainName: string;
+  sender: string;
+  senderPubKey?: Promise<HexString | undefined> | HexString;
+  route: RouteResponse;
+  approvalTransactionCount: number;
+}): Promise<bigint> {
+  if (route.sourceTransactionFee) return BigInt(route.sourceTransactionFee.amount);
```
troykessler: The embedded Max fee is based on the allowance snapshot taken by `/quote/max`, but `useApprovalStatus` keeps refreshing afterward. If the endpoint saw sufficient allowance and that allowance is revoked or consumed before Send, the current status can become `NeedsApprove`/`NeedsRevoke` while this early return ignores `approvalTransactionCount`. Validation can then accept an exact native Max amount, after which `useTransfer` spends native gas on the newly required approval(s) and leaves the transfer underfunded. Please invalidate/recompute the Max quote when the approval requirement grows, or carry enough metadata to reconcile this embedded fee with the fresh approval count.

troykessler (commented):
One remaining funding edge in the Max path.

src/features/transfer/engine/sourceFee.ts
Xaroz: Fixed in c65e2431. `estimateRouteSourceFee` now reuses the embedded `/quote/max` fee only while the refreshed allowance requires no approval. If `useApprovalStatus` changes to `NeedsApprove` or `NeedsRevoke`, send-time validation recomputes the EVM-like source fee with the current approval transaction count and returns to edit mode before broadcasting if the Max amount is underfunded. Added regression coverage for allowance changing from sufficient at quote time to approval-required before Send.

src/features/transfer/engine/sourceFee.ts
```diff
@@ -0,0 +1,122 @@
...
+  approvalTransactionCount,
+}: {
+  multiProvider: MultiProtocolProvider;
+  chainName: string;
+  sender: string;
+  senderPubKey?: Promise<HexString | undefined> | HexString;
+  route: RouteResponse;
+  approvalTransactionCount: number;
+}): Promise<bigint> {
+  if (route.sourceTransactionFee && approvalTransactionCount === 0) {
```
paulbalaji, line 57: **[P1] Validate the fee with the execution-time allowance**

This guard still trusts `approvalTransactionCount` from `useApprovalStatus`, whose value may be up to 10 seconds stale. `onSendTransactions` does not refetch it; after validation reuses the embedded fee here, `useTransfer.execute` performs independent fresh `isApproveRequired` / `isRevokeApprovalRequired` reads and may broadcast approval transactions. If allowance is consumed or revoked within that window, exact Max still passes without approval gas, then approval can consume the headroom and leave the transfer underfunded. Please make send-time validation await the same fresh allowance result that drives execution (or build one execution plan before any broadcast) so fee validation and submitted transactions cannot diverge.

paulbalaji (commented):
Rereviewed exact head c65e2431. The patch handles an allowance change already observed by the polling hook, but the execution-time allowance race remains inline. Focused tests (15), typecheck, and CI are green.

src/features/transfer/engine/useQuote.ts
Xaroz: Fixed in 8ba06dae. Max mode now stays authoritative through its TTL; at expiry it is marked stale and requires an explicit Max recalculation. It no longer falls through to an ordinary exact-input quote or changes the input amount automatically.

src/features/transfer/engine/sourceFee.ts
Xaroz: Fixed in 8ba06dae. Send now reads one fresh approval plan before validation, derives the exact revoke/approve transaction count from it, re-estimates the source fee when necessary, and passes that same plan into execution so validation and broadcast cannot diverge.

Xaroz:
@claude Please review the latest changes.

- Ordinary quote source fees are estimated in the frontend through the SDK's per-VM transaction fee dispatch, with sender-balance overrides where supported.
- Max quote fees remain API-authoritative unless the fresh approval plan changes the submitted transaction count.
- Send-time validation and execution share one fresh approval plan.
- Expired Max quotes require explicit recalculation instead of silently becoming ordinary quotes.
- Main's lazy, registry-address-verified vault collateral resolution is preserved.

paulbalaji (approved):
Rereviewed exact head `8ba06dae`, including the merge result with main/#1208. The prior P1 is resolved: Send reads one fresh approval plan, derives the fee-validation transaction count from it, and passes that same plan into execution. Expired Max quotes require explicit recalculation; ordinary quotes use balance-independent SDK fee dispatch; and the bounded, abort-aware, trusted-router vault resolution is preserved. No new findings.

Local validation passed: 371 tests, typecheck, lint, Oxfmt, diff check, production build, and bundle budgets. Exact-head CI is fully green, including both E2E shards and smoke coverage.

Non-blocking follow-up: add a component-level regression asserting that the exact `ApprovalPlan` read at Send is used for both fee validation and `useTransfer.execute`; current coverage tests the plan-count helper and fee paths separately.

PR #1198, opened 2026-08-17
fix: reserve source fees for native max

Xaroz:
## Summary

- prepare the same source transactions for wallet execution and `multiProvider.estimateTransactionFee`
- keep the native Max flow in one hook: quote the full balance, estimate its source transactions, reserve gas and external native funding, then set the reduced amount for the final quote
- validate and display source gas for native, ERC20/synthetic, swap, bridge, multi-transaction, and EVM/Tron approval routes
- keep embedded IGP and token fees inside `amountIn`; group separately funded fees by chain and token
- use the connected Starknet account estimator and pass Cosmos public keys through the shared estimator

## Dependencies

Stacked on #1197. The IGP fix depends on hyperlane-xyz/universal-router-engine#87 for authoritative `igpIncludedInAmountIn` metadata; the UI retains rollout inference for older deployments.

No additional engine field is needed for source gas. The UI estimates the returned transaction plan. Complete Solana native-Max support also requires hyperlane-xyz/hyperlane-monorepo#9270 and a subsequent UI SDK version bump.


src/features/transfer/engine/sourceGas.ts
```diff
@@ -0,0 +1,80 @@
...
+            chainNameOrId: chainName,
+            transaction,
+            sender,
+            senderPubKey: publicKey,
+          });
+        }),
+      );
+      return estimates.reduce((total, estimate) => total + BigInt(estimate.fee), 0n);
+    } catch (err) {
+      logger.warn('Failed to estimate source transaction fee', err as Error);
```
paulbalaji: SDK 41’s Solana estimator only multiplies consumed compute units by the recent priority fee and omits the base message/signature fee. I reproduced a 200k-CU estimate returning `fee: 0` when priority pricing is zero. This makes validation treat the transfer as fee-free, while native Max rejects the same route at `gasCost <= 0`. Could we include `getFeeForMessage` (plus priority fee), or reject/replace zero Solana estimates with a conservative protocol-specific value before returning this total?

paulbalaji (commented):
One inline finding on Solana source-fee accounting. Local full tests, typecheck, lint, format, build, and bundle budgets passed.

Xaroz:
@claude review

Xaroz:
@paulbalaji Fixed your Solana source-fee finding in `ce398b22` and hyperlane-xyz/hyperlane-monorepo#9270 (`f281826857`). The UI now rejects zero source-fee totals (with the existing EVM gas-budget fallback), and the SDK uses `getFeeForMessage`, which already includes the base fee plus any priority fee encoded in the message. Added regressions for zero priority pricing and versioned Solana transactions.

src/features/transfer/engine/sourceGas.ts
Xaroz: Fixed in `ce398b22` and hyperlane-xyz/hyperlane-monorepo#9270 (`f281826857`). The UI now rejects zero estimates; the SDK uses `getFeeForMessage` for the base fee plus any priority fee encoded in the message. Added zero-priority and versioned-message regressions.

src/features/transfer/engine/sourceFee.ts
```diff
@@ -0,0 +1,167 @@
...
+          sender,
+          senderPubKey: publicKey,
+        }),
+      ),
+    );
+    const total = estimates.reduce((sum, estimate) => sum + BigInt(estimate.fee), 0n);
+    if (total <= 0n) throw new Error('Source fee estimate is empty');
+    return total;
+  } catch (cause) {
+    if (protocol !== ProtocolType.Ethereum) {
```
paulbalaji: Tron has the same state-dependent `estimateGas` problem as Ethereum here. When `approvalAmounts` is non-empty, approval and route transactions are estimated concurrently against the current allowance, so the route simulation can revert until approval executes. Ethereum takes the gas-budget fallback, but this branch turns the equivalent Tron case into a hard source-fee error and blocks Continue. The Tron test currently mocks all estimates as successful; covering a route estimate that rejects while approval is pending would expose this.

src/features/transfer/engine/types.ts
```diff
@@ -94,22 +94,18 @@ export interface TransferFormValues {
...
-// A single fee component, attributable to a specific token. Engine emits
-// these per cross-chain step: one for the route fee (in the step's `asset`)
-// and one for the IGP (in `step.fee.igpToken`, which can be native or an
-// ERC20). Origin/dest chain + token address let consumers resolve
-// decimals and symbol from the token map.
+// Engine emits bridge and IGP components; the UI appends the source-chain
+// transaction fee after preparing and estimating the route transaction.
 export interface FeeComponent {
-  category: 'bridge' | 'igp';
+  category: 'bridge' | 'igp' | 'source';
```
paulbalaji, line 100: The new `source` category also reaches `ReviewDetails`, where the current two-way label maps every non-`bridge` component to `Interchain Gas`. That makes source wallet gas appear as destination/relayer gas in review mode even though the fee modal labels it correctly. Please give `source` its own `Source Gas` label there.

paulbalaji (commented):
The original Solana finding is addressed: this head rejects empty estimates, and #9270 provides the authoritative message fee; the documented SDK bump remains required for complete Solana native-Max support. Local exact-head focused/full tests, typecheck, lint/format, production build, and bundle budgets pass. Two current-head findings inline.

src/features/transfer/engine/sourceFee.ts
Xaroz: Fixed in f8bb8d84. The fallback now uses isEVMLike, so Tron approval-dependent routes use the same conservative gas-budget fallback as Ethereum. Added a regression test for route simulation failure while approval is pending.

src/features/transfer/engine/types.ts
Xaroz: Fixed in f8bb8d84. ReviewDetails now uses the same exhaustive fee-category label map as the fee breakdown modal, including source as Source Gas.

Xaroz:
@claude review please

src/features/transfer/engine/useSourceFee.ts
```diff
@@ -0,0 +1,80 @@
...
+  estimate,
+  cacheKey,
+}: {
+  route: AugmentedRoute | undefined;
+  estimate: (route: RouteResponse) => Promise<bigint>;
+  cacheKey: readonly unknown[];
+}) {
+  const query = useQuery({
+    queryKey: ['sourceFee', ...cacheKey, route?.raw ?? null],
+    queryFn: async () => requireSourceFee(await estimate(route!.raw)),
```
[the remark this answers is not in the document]
Xaroz: Fixed in 6231727d. Replaced the full route object with stable primitive route, amount, gas, and transaction key fields, with regression coverage for invalidation and refresh cache behavior.

Xaroz:
@claude Addressed in 783e938d: Source Gas now uses the augmented breakdown in Review, route transaction lookup uses the shared helper, token keys parse safely, and the EVM-like Max fallback is documented. The query-key finding was already fixed in 6231727d.

paulbalaji (approved):
Re-reviewed exact head `783e938d5881992f27ad28414ccf6b10c9a936c7`. All prior findings are resolved: Solana zero-estimate gating and SDK follow-up, Tron EVM-like fallback, Source Gas labeling/display, and compact query keys. No new findings.

Local exact-head validation: 47 files / 347 tests, typecheck, lint, changed-file format, production build, and bundle budgets passed. GitHub security/deployment checks are green; the normal test matrix is absent because this is stacked on #1197. #1197, URE #87, and the #9270 SDK follow-up remain merge/rollout ordering constraints.

PR #1199, opened 2026-08-17
fix: prompt destination wallet connect when recipient is empty

haggis-hyperlane:
## Summary
When the origin wallet is connected but no recipient exists (destination-protocol wallet not connected and no custom recipient typed):
- point `ConnectAwareSubmitButton` at the destination chain so it prompts **"Connect wallet"** for the destination protocol (e.g. Aleo) instead of showing the empty-quote **"Route is not supported"**; and
- gate the engine quote on the effective recipient the same way it's gated on the sender, so no `/v1/quote` fires until a recipient is known.

Once the destination wallet connects (or a custom recipient is entered), the recipient fills in and the normal quote flow resumes.

## Test plan
- [x] `pnpm typecheck`, `pnpm lint`, `useQuote` unit tests pass
- [ ] Manual: connect EVM origin → ETH→ETH Aleo → button reads "Connect wallet", no /quote call → connect Aleo wallet → quote proceeds
- [ ] Regression: origin-not-connected still prompts origin; EVM→EVM single-wallet unaffected

### DIFF
--- a/src/features/transfer/engine/TransferForm.tsx
+++ b/src/features/transfer/engine/TransferForm.tsx
@@ -1198,11 +1198,8 @@ function ButtonSection({
   const dstMetadata = dstChainName ? multiProvider.tryGetChainMetadata(dstChainName) : undefined;
   const dstDisplay = dstMetadata?.displayName || dstMetadata?.name || dstChainName || 'destination';
 
-  // Once the origin wallet is connected but no recipient exists — the destination
-  // wallet isn't connected and no custom recipient was entered — the engine quote
-  // comes back empty and would otherwise read "Route is not supported". Point the
-  // submit button at the destination chain so it prompts to connect that protocol's
-  // wallet instead of implying the route is unavailable.
+  // Origin connected but recipient still missing: prompt the destination wallet connect
+  // instead of the empty-quote "Route is not supported".
   const promptDestConnect = originConnected && needsRecipient && hasTokens && !!dstChainName;
   const connectChainName = promptDestConnect && dstChainName ? dstChainName : srcChainName;
 
--- a/src/features/transfer/engine/TransferForm.tsx
+++ b/src/features/transfer/engine/TransferForm.tsx
@@ -1196,11 +1196,13 @@ function ButtonSection({
 }) {
   const multiProvider = useMultiProvider();
   const dstMetadata = dstChainName ? multiProvider.tryGetChainMetadata(dstChainName) : undefined;
-  const dstDisplay = dstMetadata?.displayName || dstMetadata?.name || dstChainName || 'destination';
+  const dstDisplay = dstMetadata?.displayName ?? dstMetadata?.name ?? dstChainName ?? 'destination';
 
-  // Origin connected but recipient still missing: prompt the destination wallet connect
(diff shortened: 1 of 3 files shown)


Xaroz:
@claude review

src/features/transfer/engine/useQuote.ts
```diff
@@ -310,7 +310,10 @@ function compoundSlippageMin(output: bigint, slippageBps: number, swapStepCount:
 
 function isQuoteRequestReady(v: TransferFormValues, sender: string | undefined): boolean {
   // Non-empty checks only — engine validates / normalizes per-protocol address shapes.
+  // Recipient is the effective recipient (custom input or connected destination wallet);
+  // gate on it like sender so we don't quote a route the user can't yet receive.
   if (!sender) return false;
+  if (!v.recipient) return false;
   if (v.srcChain == null || v.dstChain == null) return false;
```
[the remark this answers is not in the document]
haggis-hyperlane: Added unit tests in `useQuote.test.ts` covering `isQuoteRequestReady`: disabled when recipient is empty, and enabled once it's set (plus the sender gate). Exported the function to test it directly. Done in 26b27795.

haggis-hyperlane:
Addressed the review feedback in 26b27795:

- **Copilot / Claude — missing recipient-gate tests:** exported `isQuoteRequestReady` and added `useQuote.test.ts` cases asserting it stays disabled when the recipient is empty and enables once set (also covers the sender gate).
- **Claude #1 — `promptDestConnect` hid the "Enter amount" hint:** now gated on `hasAmount`, so the button shows "Select tokens" → "Enter amount" first, then the destination "Connect wallet" prompt once an amount is entered.
- **CodeRabbit — `??` vs `||`:** switched `dstDisplay` to nullish coalescing per the repo guideline.

Skipped (with reason):
- **Claude #3 — redundant-looking `&& dstChainName`:** kept as-is; it's needed for TS to narrow `string | undefined` → `string`.

`pnpm vitest` (15 passing), `tsc --noEmit`, and `oxlint` all clean on the touched files.

Xaroz:
@claude review again

Xaroz (approved):

src/features/transfer/engine/TransferForm.tsx
```diff
@@ -1225,7 +1238,7 @@ function ButtonSection({
   if (!isReview) {
```
paulbalaji, line 1238: This still deadlocks when a Cosmos wallet has an address for some other Cosmos chain but not this destination. `effectiveRecipient` stays empty because address lookup is chain-specific, while `ConnectAwareSubmitButton` uses the protocol-wide account `isReady`; it therefore renders disabled `Fetching quote...` and never opens connect. Making this connect decision chain-address-aware (and covering a partially connected Cosmos account) would close the remaining path.

paulbalaji (commented):
Quote gating and the distinct-protocol destination-connect path look correct, and exact-head CI plus local tests/build are green. One chain-specific Cosmos readiness edge case inline.

src/features/transfer/engine/TransferForm.tsx
Xaroz: Fixed in #1201.

PR #1201, opened 2026-08-17
fix: make wallet connect checks chain-aware

Xaroz:
## Summary

- Resolve wallet readiness against the selected chain address.
- Keep the connect action available when a Cosmos wallet is connected to another chain only.
- Add regression coverage for the partially connected Cosmos case.

Follow-up to #1199 and [this review comment](https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/1199#discussion_r3798616143).


paulbalaji (approved):
Reviewed exact head `9bc9cf8c`. No findings. Chain-specific address readiness fixes the partial Cosmos connection deadlock while preserving the existing protocol connect flow. Local 304 tests, typecheck, lint, formatting, and production build pass; GitHub build/test/lint/typecheck/security, both Playwright shards, and wallet smoke pass.

PR #1209, opened 2026-08-24
feat: support Solana fee-aware routes

xeno097:
## Summary

- parse bridge fee inclusion and local-native fee fields returned by the universal router API, while defaulting missing IGP-inclusion metadata to the legacy conservative behavior
- account correctly for IGP already embedded in the input amount
- surface structured collateral, input-fee, quote, and transaction-encoding failures instead of reporting every failure as an unsupported route
- show local Solana transaction costs in the fee breakdown
- select `GithubRegistry` or `HttpClientRegistry` from a parsed and validated registry URL, and normalize trailing slashes for HTTP registries
- extend route-security coverage for fee-aware Sealevel cross-collateral routes

## Dependencies

- API support: https://github.com/hyperlane-xyz/universal-router-engine/pull/95
- browser HTTP registry reads: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/9332

## Testing

- `pnpm test` (46 files, 348 tests)
- `pnpm typecheck`
- `pnpm lint` (only two existing `_e2e` console warnings)
- `oxfmt --check ./src`

## Manual verification

Submitted Solana to Base `USDCFEE/sol` transfers through the local UI, router API, and registry using both execution paths:

- [Fee-quoting service unavailable: on-chain fee fallback](https://explorer.hyperlane.xyz/message/0x69ffd2845669699fbbb664d64fb69c256129fbfcf58be68411d387716af3f9c8)
- [Fee-quoting service running: transient quoted-fee path](https://explorer.hyperlane.xyz/message/0x193b2279d02d3787e90ba33a69e627accb776181bd40e95396702bb75312fc0b)


src/features/registry.ts
```diff
@@ -0,0 +1,18 @@
...
+import type { IRegistry } from '@hyperlane-xyz/registry';
+
+export function createConfiguredRegistry(config: {
+  registryUrl?: string;
+  registryBranch?: string;
+  registryProxyUrl?: string;
+}): IRegistry {
+  if (!config.registryUrl && !config.registryBranch) return new PartialRegistry({});
+  if (config.registryUrl && !config.registryUrl.toLowerCase().includes('github')) {
+    return new HttpClientRegistry(config.registryUrl);
```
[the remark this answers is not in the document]
xeno097: Fixed in f11e88e6. Registry selection now parses the URL, requires the exact `github.com` hostname, and accepts only repository-shaped paths with an optional `/tree/{branch}` suffix. Non-GitHub URLs containing `github` and malformed GitHub paths remain `HttpClientRegistry`; regression tests cover both cases. Full suite: 346 tests passing, plus typecheck and lint.

src/features/api/types.test.ts
```diff
@@ -3,6 +3,28 @@ import { describe, expect, test } from 'vitest';
...
+      QuoteResponseSchema.parse({
+        expiresAt: Math.floor(Date.now() / 1000) + 30,
+        routes: [],
+        rejections: [
+          {
+            code: 'insufficient_destination_collateral',
+            message: 'Insufficient destination liquidity for this amount',
+            srcChain: 8453,
+            dstChain: 1399811149,
+            srcToken: '0x2079997E72ffE2B4E455D94Cd88b5Eca0D9155e4',
```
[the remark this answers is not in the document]
xeno097: Fixed in f11e88e6. The EVM fixture address is now lowercase; the case-sensitive Solana address is unchanged.

paulbalaji (commented):
Reviewed exact head `03d75b19` with independent transfer, registry, adversarial, and Kimi passes. Two additional findings are inline; I did not duplicate the existing CodeRabbit registry-classification thread. Validation: full Vitest 343/343 on the feature head; focused exact-head tests 157/157 plus the final 11/11 delta tests; typecheck, lint, Oxfmt, production build, and diff check passed. The latest CI rerun is still in progress after the format-only follow-up; `e2e-wallet-full` is skipped on PRs. Live production API and local wallet-E2E reproductions confirm the schema failure. Fee arithmetic and route-security review found no additional issues.

src/features/api/types.ts
```diff
@@ -179,6 +179,7 @@ export const QuoteBridgeStepSchema = z.object({
     tokenFee: BigIntString,
     igpToken: TokenAddress,
     igpAmount: BigIntString,
+    igpIncludedInAmountIn: z.boolean(),
```
paulbalaji: [P1] Keep this field rollout-compatible. The default production router currently omits `igpIncludedInAmountIn`; parsing a live Ethereum→Optimism response with this exact schema fails with `Required` at each bridge fee, so every non-empty quote is rejected before rendering if WUI lands ahead of the backend. The shared `tests/e2e-wallet/helpers/quote.ts` mock also omits it and locally reproduces the Zod failure; `e2e-wallet-full` is skipped on this PR. Please use `z.boolean().default(false)` (matching the old conservative debit behavior) and add a legacy-response test, or enforce and verify the backend rollout before this deploys.

src/features/registry.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { IRegistry } from '@hyperlane-xyz/registry';
+
+export function createConfiguredRegistry(config: {
+  registryUrl?: string;
+  registryBranch?: string;
+  registryProxyUrl?: string;
+}): IRegistry {
+  if (!config.registryUrl && !config.registryBranch) return new PartialRegistry({});
+  if (config.registryUrl && !isGithubRepositoryUrl(config.registryUrl)) {
+    return new HttpClientRegistry(config.registryUrl);
```
paulbalaji: [P2] Normalize trailing slashes before constructing this client. `HttpClientRegistry` concatenates its base URL and endpoint; with the common `NEXT_PUBLIC_REGISTRY_URL=http://localhost:3334/`, `getMetadata()` requests `http://localhost:3334//metadata` (reproduced with registry 25.4.0), and the same applies to addresses and routes. Those paths can return 404, causing the UI to fall back to published data instead of the configured custom route. Strip trailing `/` characters and cover this case.

src/features/api/types.ts
xeno097: Fixed in f826dac9.  now defaults to , preserving the conservative legacy debit behavior when the backend field is absent. Added a legacy-response regression test. Full UI suite: 46 files / 348 tests; typecheck, lint, and formatting pass.

src/features/registry.ts
xeno097: Fixed in f826dac9. The configured registry URL is normalized by stripping trailing slashes before either registry client is constructed. Added coverage for  resolving to . Full UI suite and checks pass.

paulbalaji (approved):
Rereviewed exact head f826dac9. The prior rollout-compatibility and HTTP registry trailing-slash findings are fixed: missing igpIncludedInAmountIn defaults conservatively to false, and configured registry URLs are normalized before client construction. Focused registry/API tests passed 13/13, typecheck passed, exact-head CI is green, and all review threads are resolved/outdated. No remaining blocking findings.

PR #1212, opened 2026-08-26
ci: migrate workflows to Depot CI

paulbalaji:
## Summary

- move the main CI workflow and all three reusable E2E workflows from `.github/workflows/` to `.depot/workflows/`
- run every migrated job on pinned Depot Ubuntu 24.04 sandboxes sized to match this public repo's GitHub-hosted runners (4 vCPU / 16 GB)
- preserve triggers, branch filters, dependencies, matrices, caches, artifacts, notifications, and required check names
- preserve required CI for fork PRs with an unprivileged GitHub Actions fallback
- build the Next.js app explicitly on E2E cache misses and key shared build caches by exact commit plus embedded config versions
- build and default the app to the current URE Cloudflare endpoint, `https://hyperswaps-api.hyperlane.xyz`; the retired `router.services.hyperlane.xyz` endpoint now resolves to `192.0.2.1`
- keep GitHub-native Claude review, branch-sync PR creation, and dependency-update automation in GitHub Actions

## Controlled benchmark

The benchmark used the same application tree and one identical serial workload on both providers: frozen install, format/diff check, lint, 403 unit tests, and production build. Both cold runs used the fresh `wui-depot-migration-20260827-v2` cache key and logged cache misses; both warm runs logged primary-key cache hits. The runners were resource-equivalent: public-repo GitHub `ubuntu-24.04` and Depot `depot-ubuntu-24.04-4` both provide 4 vCPU / 16 GB. Durations are workflow start-to-finish.

| Cache state | GitHub Actions | Depot CI | Depot speedup |
| --- | ---: | ---: | ---: |
| Cold | 2m53s | 1m39s | 1.75x |
| Warm | 1m53s | 0m59s | 1.92x |

The earlier `0.64x` cold result was invalid: the default Depot sandbox had 2 vCPU / 8 GB while this public repo's GitHub runner had 4 vCPU / 16 GB. Next.js exposed the mismatch directly by using one worker on Depot versus three on GitHub. The private reference repositories use 2 vCPU / 8 GB GitHub runners, so their default Depot label was equivalent; WUI required the `-4` label.

Exact corrected runs:

- GitHub Actions cold: [`33024586966`](https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/actions/runs/33024586966)
- Depot CI cold: [`4ccs8n5tp9`](https://depot.dev/orgs/ptcjppsfhw/workflows/4ccs8n5tp9)
- GitHub Actions warm: [`33024757128`](https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/actions/runs/33024757128)
- Depot CI warm: [`htfc17hg17`](https://depot.dev/orgs/ptcjppsfhw/workflows/htfc17hg17)

Benchmark branches were based on the same source tree as this PR: `pbio/gha-ci-benchmark` and `pbio/depot-ci-benchmark`.

## Validation

At exact head `9d9bccc6`:

- current-head run [`3fjv2dlxk0`](https://depot.dev/orgs/ptcjppsfhw/workflows/3fjv2dlxk0) passed every executed job: format, lint, 403 tests, typecheck, both 33-test E2E shards, 5-test smoke gate, and merged reports
- the most recent full migration run [`p5qv137ndc`](https://depot.dev/orgs/ptcjppsfhw/workflows/p5qv137ndc) also passed build and the Chromium and Firefox wallet matrices; those path-filtered jobs intentionally skipped on the checkout-only follow-up
- fork-only GitHub Actions build, typecheck, format, lint, and test jobs use read-only permissions and disable checkout credential persistence
- all five shared `.next/` consumer keys require the exact full commit SHA; producers may restore only a config-scoped prefix as a build warm start
- Slack failure alerts capture and link to `DEPOT_JOB_URL`
- `main`, `nautilus`, `nexus`, `injective`, `trump`, and `ousdt` use strict exact-name `ci / ...` required checks with no app pin, allowing Depot or the fork fallback to satisfy them
- the application fallback and `.env.example` use `https://hyperswaps-api.hyperlane.xyz`
- local production build, 403 tests, lint, actionlint, oxfmt, and `git diff --check` passed
- branch is zero commits behind current `origin/main`; no rebase required

## Remaining cutover gate

- An organization owner must accept the pending Depot Code Access permission update: https://github.com/organizations/hyperlane-xyz/settings/installations/110670256
- Then import the repository-scoped `E2E_SLACK_WEBHOOK`. The Slack notification step safely skips when no webhook is configured, so this does not block current checks, but the secret must be imported before merge to preserve failure notifications.

### DIFF
--- a/.depot/workflows/ci.yml
+++ b/.depot/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-24.04
     steps:
       - uses: actions/checkout@v7
       - uses: pnpm/action-setup@v6
@@ -60,13 +60,13 @@ jobs:
     # TypeScript validation runs inside `next build`. Keep this dependent job
     # until the required `typecheck` branch-protection context is removed.
     needs: [build]
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-24.04
     steps:
       - name: Confirm build typechecked successfully
         run: echo "TypeScript validation completed in the required build job"
 
   format:
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-24.04
     steps:
       - uses: actions/checkout@v7
       - uses: pnpm/action-setup@v6
@@ -100,7 +100,7 @@ jobs:
           fi
 
   lint:
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-24.04
     steps:
       - uses: actions/checkout@v7
       - uses: pnpm/action-setup@v6
@@ -128,7 +128,7 @@ jobs:
         run: pnpm run lint
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-24.04
     steps:
       - uses: actions/checkout@v7
       - uses: pnpm/action-setup@v6
@@ -162,13 +162,13 @@ jobs:
       (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
       github.event_name == 'merge_group' ||
       github.event_name == 'workflow_dispatch'
-    uses: ./.github/workflows/e2e.yml
+    uses: ./.depot/workflows/e2e.yml
     secrets: inherit
 
   # Fast wallet-connected E2E subset, gates every PR.
   e2e-smoke:
     needs: [build]
-    uses: ./.github/workflows/e2e-smoke.yml
+    uses: ./.depot/workflows/e2e-smoke.yml
     secrets: inherit
 
   # Full wallet-connected E2E matrix (chromium/firefox/webkit). Main-only so
@@ -179,5 +179,5 @@ jobs:
       (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
       github.event_name == 'merge_group' ||
       github.event_name == 'workflow_dispatch'
-    uses: ./.github/workflows/e2e-wallet-full.yml
+    uses: ./.depot/workflows/e2e-wallet-full.yml
     secrets: inherit
--- a/.depot/workflows/ci.yml
+++ b/.depot/workflows/ci.yml
@@ -10,6 +10,8 @@ on:
 
 env:
   NEXT_PUBLIC_ROUTER_API_URL: https://hyperswaps-api.hyperlane.xyz
+  # Bump when NEXT_PUBLIC_WALLET_CONNECT_ID changes.
+  WALLET_CONNECT_ID_CACHE_VERSION: v1
 
 jobs:
   build:
@@ -49,7 +51,7 @@ jobs:
         uses: actions/cache@v6
         with:
           path: .next/
-          key: ${{ runner.os }}-nextjs-build-v3-${{ env.NEXT_PUBLIC_ROUTER_API_URL }}-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
+          key: ${{ runner.os }}-nextjs-build-v3-${{ env.NEXT_PUBLIC_ROUTER_API_URL }}-${{ env.WALLET_CONNECT_ID_CACHE_VERSION }}-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
 
       - name: build
         run: pnpm run build
--- a/.depot/workflows/ci.yml
+++ b/.depot/workflows/ci.yml
@@ -8,6 +8,9 @@ on:
   merge_group:
   workflow_dispatch:
 
+env:
+  NEXT_PUBLIC_ROUTER_API_URL: https://hyperswaps-api.hyperlane.xyz
+
 jobs:
   build:
     runs-on: depot-ubuntu-24.04
@@ -46,7 +49,7 @@ jobs:
         uses: actions/cache@v6
         with:
           path: .next/
-          key: ${{ runner.os }}-nextjs-build-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
+          key: ${{ runner.os }}-nextjs-build-v3-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
 
       - name: build
         run: pnpm run build
--- a/.depot/workflows/ci.yml
+++ b/.depot/workflows/ci.yml
@@ -13,7 +13,7 @@ env:
 
 jobs:
   build:
-    runs-on: depot-ubuntu-24.04
+    runs-on: depot-ubuntu-24.04-4
     steps:
       - uses: actions/checkout@v7
       - uses: pnpm/action-setup@v6
@@ -49,7 +49,7 @@ jobs:
         uses: actions/cache@v6
         with:
           path: .next/
-          key: ${{ runner.os }}-nextjs-build-v3-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
+          key: ${{ runner.os }}-nextjs-build-v3-${{ env.NEXT_PUBLIC_ROUTER_API_URL }}-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
 
       - name: build
         run: pnpm run build
@@ -63,13 +63,13 @@ jobs:
     # TypeScript validation runs inside `next build`. Keep this dependent job
     # until the required `typecheck` branch-protection context is removed.
     needs: [build]
-    runs-on: depot-ubuntu-24.04
+    runs-on: depot-ubuntu-24.04-4
     steps:
       - name: Confirm build typechecked successfully
         run: echo "TypeScript validation completed in the required build job"
 
   format:
-    runs-on: depot-ubuntu-24.04
+    runs-on: depot-ubuntu-24.04-4
     steps:
       - uses: actions/checkout@v7
       - uses: pnpm/action-setup@v6
@@ -103,7 +103,7 @@ jobs:
           fi
 
   lint:
-    runs-on: depot-ubuntu-24.04
+    runs-on: depot-ubuntu-24.04-4
     steps:
       - uses: actions/checkout@v7
       - uses: pnpm/action-setup@v6
@@ -131,7 +131,7 @@ jobs:
         run: pnpm run lint
 
   test:
-    runs-on: depot-ubuntu-24.04
+    runs-on: depot-ubuntu-24.04-4
     steps:
       - uses: actions/checkout@v7
       - uses: pnpm/action-setup@v6
--- a/.depot/workflows/ci.yml
+++ b/.depot/workflows/ci.yml
@@ -51,7 +51,9 @@ jobs:
         uses: actions/cache@v6
         with:
           path: .next/
-          key: ${{ runner.os }}-nextjs-build-v3-${{ env.NEXT_PUBLIC_ROUTER_API_URL }}-${{ env.WALLET_CONNECT_ID_CACHE_VERSION }}-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
+          key: ${{ runner.os }}-nextjs-build-v4-${{ env.NEXT_PUBLIC_ROUTER_API_URL }}-${{ env.WALLET_CONNECT_ID_CACHE_VERSION }}-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-nextjs-build-v4-${{ env.NEXT_PUBLIC_ROUTER_API_URL }}-${{ env.WALLET_CONNECT_ID_CACHE_VERSION }}-
 
       - name: build
         run: pnpm run build
--- a/.depot/workflows/e2e-wallet-full.yml
+++ b/.depot/workflows/e2e-wallet-full.yml
@@ -54,7 +54,7 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           path: .next/
-          key: ${{ runner.os }}-nextjs-build-v3-${{ env.NEXT_PUBLIC_ROUTER_API_URL }}-${{ env.WALLET_CONNECT_ID_CACHE_VERSION }}-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
+          key: ${{ runner.os }}-nextjs-build-v4-${{ env.NEXT_PUBLIC_ROUTER_API_URL }}-${{ env.WALLET_CONNECT_ID_CACHE_VERSION }}-${{ github.sha }}
 
       - name: Build on cache miss
         if: steps.build-cache.outputs.cache-hit != 'true'
@@ -100,6 +100,11 @@ jobs:
     env:
       E2E_SLACK_WEBHOOK: ${{ secrets.E2E_SLACK_WEBHOOK }}
     steps:
+      - name: Capture Depot job URL
+        if: ${{ env.E2E_SLACK_WEBHOOK != '' }}
+        id: depot-job
+        run: echo "url=$DEPOT_JOB_URL" >> "$GITHUB_OUTPUT"
+
       - name: Notify Slack on e2e-wallet failure
         if: ${{ env.E2E_SLACK_WEBHOOK != '' }}
         uses: slackapi/slack-github-action@v3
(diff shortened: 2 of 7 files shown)


Xaroz:
@claude review

paulbalaji:
Followed up on the latest Claude review at exact head `f8655e30`.

- Re-fetched `origin/main` (`597c0182`): this branch is zero commits behind, so no rebase was needed.
- Confirmed the required-context cutover issue. Updated `main` branch protection from the stale GitHub Actions checks to the reporting Depot Code Access checks: `ci / build`, `ci / lint`, `ci / test`, `ci / format`, and `ci / typecheck`. All five now resolve on this PR.
- Confirmed Depot still lacks `E2E_SLACK_WEBHOOK`. Pushed `f8655e30` to skip the Slack step safely when the secret is absent, preventing failure-on-failure; the secret import remains a pre-merge gate to preserve notifications.
- Updated the PR description to reflect the guarded behavior.

Validation: `oxfmt` clean, `git diff --check` clean, `actionlint` has only the two pre-existing SC2086 info messages, and current-head CI completed with 18 successful checks, 6 intentional skips, 0 failures/pending. CodeRabbit also completed cleanly.

.depot/workflows/ci.yml
```diff
@@ -162,13 +167,13 @@ jobs:
       (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
       github.event_name == 'merge_group' ||
       github.event_name == 'workflow_dispatch'
-    uses: ./.github/workflows/e2e.yml
+    uses: ./.depot/workflows/e2e.yml
```
Xaroz, line 172: **[P1] Preserve required CI for fork pull requests**

Depot's current [compatibility documentation](https://depot.dev/docs/ci/compatibility#pull-request-workflows-triggered-from-forks) says workflows triggered by `pull_request`/`pull_request_target` from forks are not supported yet. This public repository has received fork-based PRs (for example #534, #446, and #363). Once the GitHub CI workflow is removed, a new fork PR cannot emit `ci / build`, `ci / lint`, `ci / test`, `ci / format`, or `ci / typecheck`, so it gets no test feedback and remains blocked by required checks. Keep the PR gate on GitHub Actions until Depot supports forks, or explicitly make external fork PRs unsupported and document the replacement contribution flow.

Xaroz (commented):
## Review summary

**Verdict: conditional.** The current-head internal Depot run is healthy, all four workflows parse and format cleanly, `main` now requires the correct Depot app, and the prior router/WalletConnect cache findings are resolved. I found two cutover regressions that can block real PR classes, plus three correctness gaps in the migrated behavior.

### Blocking

- **P1:** Depot currently does not run pull-request workflows from forks, so removing GitHub CI drops all required checks for external contributions.
- **P1:** `nautilus`, `nexus`, `injective`, `trump`, and `ousdt` still require legacy GitHub Actions app/name combinations; only `main` has been cut over to Depot checks.

### Other correctness gaps

- The shared `.next` key omits production build inputs, allowing E2E consumers to restore stale output after config-only changes.
- Full-wallet failure alerts still link to a GitHub Actions run instead of the Depot job.
- CI uses the live router endpoint, but the application fallback and `.env.example` still point at the retired endpoint, hiding broken default deployments.

### Validation

Reviewed `597c0182..f8655e30`, all four changed files, prior review threads, relevant git history/blame, all six target branches' live protection rules, and the exact-head check runs. Oxfmt and YAML parsing pass. Current Depot generic E2E shards, smoke suite, and report merge pass. No authorization/value-transfer code changed; risk is CI coverage and gate integrity.

.depot/workflows/ci.yml
```diff
@@ -8,9 +8,14 @@ on:
...
 
+env:
+  NEXT_PUBLIC_ROUTER_API_URL: https://hyperswaps-api.hyperlane.xyz
+  # Bump when NEXT_PUBLIC_WALLET_CONNECT_ID changes.
+  WALLET_CONNECT_ID_CACHE_VERSION: v1
+
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-24.04-4
```
Xaroz, line 18: **[P1] Cut over every protected target branch, not only `main`**

This workflow still targets six branches, but live protection settings show only `main` requires the new `ci / ...` checks from Depot Code Access (app `219785`). `nautilus`, `nexus`, `injective`, `trump`, and `ousdt` still require legacy names such as `build`/`prettier` from GitHub Actions (app `15368`). After this workflow becomes authoritative for those targets, Depot's same-purpose checks cannot satisfy app-scoped legacy requirements, so sync and feature PRs will stay blocked. Update the app and exact check names for all five branches atomically with the migration, or retain their GitHub workflows until each cutover is complete. GitHub documents that an app-scoped required check [must come from the expected app](https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks#required-status-checks-from-unexpected-sources).

.depot/workflows/ci.yml
```diff
@@ -46,7 +51,7 @@ jobs:
         uses: actions/cache@v6
         with:
           path: .next/
-          key: ${{ runner.os }}-nextjs-build-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
+          key: ${{ runner.os }}-nextjs-build-v3-${{ env.NEXT_PUBLIC_ROUTER_API_URL }}-${{ env.WALLET_CONNECT_ID_CACHE_VERSION }}-${{ hashFiles('src/**', 'pnpm-lock.yaml', 'next.config.js', 'tsconfig.json') }}
```
Xaroz: **[P2] Include every production build input in the shared key**

The key still omits inputs that change `.next`, including `tailwind.config.js`, `postcss.config.js`, `package.json`, `pnpm-workspace.yaml`, the Sentry configs, and `scripts/fetch-fonts.mjs`. This matters because [cache entries are immutable](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-action-usage): on an exact hit, this job restores the old `.next` and rebuilds it locally, but the entry is not updated. The E2E consumers then restore that old entry, see `cache-hit == 'true'`, and skip their fallback build. A Tailwind-config-only PR can therefore pass `build` while E2E serves the previous CSS. Hash all production inputs in all five keys, or pass the current run's build to consumers as an artifact instead of treating an immutable cache as an artifact.

.depot/workflows/e2e-wallet-full.yml
```diff
@@ -84,12 +96,15 @@ jobs:
...
+    runs-on: depot-ubuntu-24.04-4
+    env:
+      E2E_SLACK_WEBHOOK: ${{ secrets.E2E_SLACK_WEBHOOK }}
     steps:
       - name: Notify Slack on e2e-wallet failure
+        if: ${{ env.E2E_SLACK_WEBHOOK != '' }}
         uses: slackapi/slack-github-action@v3
         with:
-          webhook: ${{ secrets.E2E_SLACK_WEBHOOK }}
+          webhook: ${{ env.E2E_SLACK_WEBHOOK }}
```
Xaroz, line 112: **[P2] Link failure alerts to the Depot job**

The payload below still constructs `${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}`, but this job no longer has a GitHub Actions run page; its check details point to Depot. On the exact path where this notification is useful, the alert's debugging link is therefore stale or nonexistent. Depot provides [`DEPOT_JOB_URL`](https://depot.dev/docs/ci/observability/depot-ci-logs#link-to-a-job-from-within-a-workflow) specifically for Slack/webhook notifications. Use that value (or an explicit Depot workflow URL) in the message.

.depot/workflows/ci.yml
```diff
@@ -8,9 +8,14 @@ on:
   merge_group:
   workflow_dispatch:
 
+env:
+  NEXT_PUBLIC_ROUTER_API_URL: https://hyperswaps-api.hyperlane.xyz
```
Xaroz, line 12: **[P2] Update the runtime default, not only the CI override**

This makes Depot builds use the live endpoint, but `src/consts/config.ts:19` and `.env.example:5` still default to `https://router.services.hyperlane.xyz`. I verified that hostname resolves to the TEST-NET address `192.0.2.1` and times out, while `https://hyperswaps-api.hyperlane.xyz/v1/chains` returns 200. CI will now pass while a template/local deployment without an explicit override remains broken. Update the application fallback and example in the same cutover, or remove the fallback and fail clearly when configuration is absent.

.depot/workflows/ci.yml
paulbalaji: Fixed in e12a6b1c. Added fork-only GitHub Actions build, typecheck, format, lint, and test jobs. Internal PR jobs skip under non-required names. Required ci contexts are app-unpinned so either Depot or the fork fallback can satisfy them.

.depot/workflows/ci.yml
paulbalaji: Fixed operationally. nautilus, nexus, injective, trump, and ousdt now strictly require ci build, lint, test, and format contexts with no app pin. main retains those plus typecheck.

.depot/workflows/ci.yml
paulbalaji: Fixed in e12a6b1c. All five shared Next.js cache keys use the exact full commit SHA. Producers may restore a config-scoped prefix only as a warm start; E2E consumers restore exact-SHA artifacts only.

.depot/workflows/e2e-wallet-full.yml
paulbalaji: Fixed in e12a6b1c. The workflow captures DEPOT_JOB_URL and the Slack failure payload links directly to that Depot job.

.depot/workflows/ci.yml
paulbalaji: Fixed in e12a6b1c. Both the application fallback and env example now use https://hyperswaps-api.hyperlane.xyz.

.github/workflows/ci-forks.yml
```diff
@@ -0,0 +1,112 @@
...
+env:
+  NEXT_PUBLIC_ROUTER_API_URL: https://hyperswaps-api.hyperlane.xyz
+
+jobs:
+  build:
+    name: ${{ github.event.pull_request.head.repo.fork && 'ci / build' || 'fork-ci-disabled / build' }}
+    if: ${{ github.event.pull_request.head.repo.fork }}
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v7
```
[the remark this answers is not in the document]
paulbalaji: Fixed in 9d9bccc6. All four fork-workflow checkout steps now disable credential persistence. Actionlint and formatting checks pass.

paulbalaji:
Applied every current review finding at exact head `9d9bccc6`:

- `e12a6b1c`: added unprivileged fork PR CI; updated all six target branches to strict exact-name, app-unpinned required checks; made shared Next.js cache consumption exact-SHA; linked Slack alerts to the Depot job; updated the router fallback and `.env.example`
- `9d9bccc6`: disabled checkout credential persistence in every fork CI job

Validation:

- current-head checks: 18 successful, 11 intentionally skipped, 0 failed/pending
- Depot run: https://depot.dev/orgs/ptcjppsfhw/workflows/3fjv2dlxk0
- local production build, 403 tests, lint, actionlint, oxfmt, and `git diff --check` passed
- branch remains zero commits behind `main`
- replied to every finding and resolved all review threads

Remaining operational cutover item: import `E2E_SLACK_WEBHOOK` after the pending Depot Code Access permission update is accepted. This does not block current CI; the notification job safely skips until configured.

.github/workflows/ci-forks.yml
```diff
@@ -0,0 +1,120 @@
+name: fork-ci
+
+on:
+  pull_request:
+    branches: [main, nautilus, nexus, injective, trump, ousdt]
```
Xaroz, line 5: [P1] Land workflows before switching these protections

Live protections for all six branches already require the new `ci / ...` contexts, but only this PR head contains the workflows that emit them. Existing PRs #857 (`injective`), #855 (`ousdt`), #867 (`trump`), and #1115 (`nexus`) currently expose no `ci / ...` checks. Merging this PR only to `main` still does not place these workflow files on the other base refs, so their PRs cannot satisfy the new requirements. Make the rollout atomic: land or cherry-pick the workflows on each protected base before changing its required contexts, or temporarily restore the old rules.

.github/workflows/ci-forks.yml
```diff
@@ -0,0 +1,120 @@
...
+
+permissions:
+  contents: read
+
+env:
+  NEXT_PUBLIC_ROUTER_API_URL: https://hyperswaps-api.hyperlane.xyz
+
+jobs:
+  build:
+    name: ${{ github.event.pull_request.head.repo.fork && 'ci / build' || 'fork-ci-disabled / build' }}
```
Xaroz, line 15: [P2] Keep the required gate pinned to a trusted app

Using the same `ci / ...` names for Depot and GitHub Actions led the live protection rules to `app_id: null`. That permits any installed app capable of publishing checks to satisfy these names, rather than requiring the expected CI provider. GitHub explicitly recommends associating required checks with the expected app when checks can come from unexpected sources. Preserve an app-scoped gate by exposing one authoritative result, for example a trusted GitHub Actions aggregator that selects the Depot or fork path. See [GitHub required-check troubleshooting](https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks#required-status-checks-from-unexpected-sources).

.github/workflows/ci-forks.yml
```diff
@@ -0,0 +1,120 @@
...
+      - uses: actions/setup-node@v7
+        with:
+          node-version-file: '.nvmrc'
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Test
+        run: pnpm run test
```
Xaroz, line 120: [P2] Preserve E2E coverage for fork PRs

The removed GitHub CI ran generic E2E for PRs targeting `main` and `e2e-smoke` for every PR (`.github/workflows/ci.yml:158-172` at the base commit). Depot does not run for fork PRs, while this fallback ends after unit tests, so external contributions now receive no browser or wallet-smoke coverage. Add secretless generic E2E and the smoke subset where credentials permit, or explicitly document and accept the reduced fork validation.

Xaroz (commented):
## Re-review summary

Reviewed updated head `9d9bccc6` against base `597c0182`.

All five findings from the prior review are structurally resolved: fork CI exists, all protections use the new names, cache consumers use exact-SHA keys, Slack uses a Depot URL, and the router default is fixed.

Three migration risks remain:

- **P1 rollout blocker:** live protections already require `ci / ...` on all six branches, but the workflows have not reached the five non-main base refs. Existing PRs targeting those branches cannot emit the required contexts.
- **P2 check-source integrity:** accepting the same names from Depot and GitHub Actions required removing the app pin, so another installed checks-capable app can satisfy the merge gate.
- **P2 coverage regression:** fork fallback restores the five required checks but omits the generic and wallet-smoke E2E jobs previously run for PRs.

## Verification

- Exact head and full diff inspected; prior review threads deduplicated.
- `git diff --check`, Oxfmt, and YAML parsing pass.
- Production build passes without exported WalletConnect/router overrides.
- Exact-head Depot required checks, generic E2E shards, and wallet smoke pass.
- Live protection rules and representative open release-branch PR checks inspected.

.github/workflows/ci-forks.yml
paulbalaji: Fixed operationally. Restored the legacy GitHub Actions app-pinned required checks on nautilus, nexus, injective, trump, and ousdt. main keeps its migrated ci contexts. Each release branch will cut over only after latest main, including the migrated workflows, is merged into that branch.

.github/workflows/ci-forks.yml
paulbalaji: No aggregator change in this PR. For the rollout phase, the five non-main branches are back on their legacy GitHub Actions checks pinned to app 15368. Each branch will move to migrated checks only after latest main, including the workflows, has landed there; its protection can then be updated selectively for the expected provider.

.github/workflows/ci-forks.yml
paulbalaji: No code change in this PR. The five non-main branches retain their existing GitHub Actions workflows and app-pinned protections, so their current fork coverage is unchanged. The main fork fallback remains core-check-only for now; E2E coverage for that path should be handled separately rather than expanding this migration during the branch rollout fix.

Xaroz (approved):

paulbalaji:
@Xaroz — is lack of fork-PR support actually a blocker for this repo (and our other public repos)? Depot CI currently does not run fork-triggered workflows, so a complete cutover means external fork PRs would not get automatic CI. The alternative is keeping a GitHub Actions fallback—or using Depot-hosted GitHub Actions runners—for fork PRs, which adds operational complexity. Before we standardize that fallback, do we need fork PR CI as a hard requirement in practice?

Xaroz:
@paulbalaji I'd say it is probably not necessary for now at least we don't get many fork PRs anyways

paulbalaji:
> @paulbalaji I'd say it is probably not necessary for now at least we don't get many fork PRs anyways

ya i thought as much, even if we do we can always just remake the pr for the ci (we've done this already with a few prs in monorepo because of limitations in triggering agent builds)

PR #1214, opened 2026-08-28
fix(routeSecurity): support Starknet origin approvals

haggis-hyperlane:
## Summary
- Nexus showed "Route is not supported" for Starknet origin USDC transfers (e.g. Starknet USDC -> Arbitrum) even though the registry connects both directions and the router engine returns a valid quote. The reverse (Arbitrum -> Starknet) worked because its origin approval is an EVM tx.
- Root cause: the client-side route-security validator only decoded EVM approvals (`decodeEvmApproval` needs `to`/`data`) and capped sdkWarp routes at one approval. Starknet origin routes return native approve calls (`contractAddress`/`entrypoint`/`calldata`) and emit two approvals (bridged token + IGP gas-fee token). Every candidate route was dropped in `useQuote`, and with no engine rejection the button fell back to the generic message.
- Fix: decode Starknet approve calldata, compare felts numerically, bind both approvals to the bridge router, bound amounts (bridged = input amount, fee = quoted `igpAmount`), and allow the second gas-fee approval for Starknet sources only. EVM/Sealevel behavior unchanged.

## Test plan
- [x] `npx vitest --run src/features/routeSecurity` (108 passed, incl. 4 new Starknet approval cases)
- [x] `npx tsc` clean
- [x] `npx oxlint ./src/features/routeSecurity` clean
- [ ] Manual: confirm Starknet USDC -> Arbitrum quotes to a Continue state in Nexus with the engine pointed at hyperswaps-api

### DIFF
--- a/src/features/routeSecurity/sdk.ts
+++ b/src/features/routeSecurity/sdk.ts
@@ -3,7 +3,7 @@ import { decodeFunctionData, erc20Abi, type Hex } from 'viem';
 
 import type { QuoteBridgeStep, RouteResponse, RouteTx } from '../api/types';
 import type { RouteSecurityValidationResult } from './types';
-import { firstBridge, sameTokenAddress } from './utils';
+import { firstBridge, isEngineNativeToken, sameTokenAddress } from './utils';
 
 const SDK_APPROVAL_CATEGORY = 'approval';
 const SDK_REVOKE_CATEGORY = 'revoke';
@@ -13,6 +13,7 @@ export function validateSdkRouteTx(
   route: RouteResponse,
   tx: Extract<RouteTx, { protocol: string }>,
   srcProtocol: ProtocolType,
+  nativeTokenAddress?: string,
 ): RouteSecurityValidationResult {
   if (route.executionKind !== 'sdkWarp') {
     return { valid: false, reason: 'SDK transaction shape is only supported for sdkWarp routes' };
@@ -33,7 +34,7 @@ export function validateSdkRouteTx(
   }
 
   if (tx.category === SDK_APPROVAL_CATEGORY || tx.category === SDK_REVOKE_CATEGORY) {
-    return validateSdkApprovalTx(route, tx, srcProtocol);
+    return validateSdkApprovalTx(route, tx, srcProtocol, nativeTokenAddress);
   }
 
   if (tx.category !== SDK_TRANSFER_CATEGORY) {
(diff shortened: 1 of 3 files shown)


src/features/routeSecurity/sdk.ts
```diff
@@ -85,6 +90,78 @@ function validateSdkApprovalTx(
...
+  }
+
+  const igpAmount = parseAmount(bridge.fee.igpAmount);
+  if (igpAmount == null || igpAmount === 0n) {
+    return { valid: false, reason: 'SDK approval token does not match route input token' };
+  }
+  if (approval.amount !== igpAmount) {
+    return { valid: false, reason: 'SDK approval amount does not match route fee amount' };
+  }
+  return { valid: true };
```
paulbalaji, line 134: **Medium — bind the fee/revoke `contractAddress` to a trusted token.** Right now any non-asset token passes when its amount equals `igpAmount`, while the earlier revoke branch accepts any token at amount zero. The route API is external and `useTransfer` submits every accepted call, so a compromised/malformed response can add an unrelated `approve` call and leave an unwanted router allowance. A live quote uses `fee.igpToken = 0x0` but approves canonical STRK, so resolve the sentinel from trusted chain/route metadata, require every approval/revoke token to equal the trusted asset or fee token, and add mismatched-token tests.

src/features/routeSecurity/sdk.ts
```diff
@@ -85,6 +90,78 @@ function validateSdkApprovalTx(
...
+  if (!sameFelt(approval.spender, bridge.router)) {
+    return { valid: false, reason: 'SDK approval spender does not match bridge router' };
+  }
+
+  if (tx.category === SDK_REVOKE_CATEGORY) {
+    if (approval.amount !== 0n) return { valid: false, reason: 'SDK revoke amount must be zero' };
+    return { valid: true };
+  }
+
+  if (sameFelt(approval.token, bridge.asset)) {
```
paulbalaji: **Medium — `StarknetHypNative` asset approvals still fall through to the fee branch.** URE exposes a native bridge asset as the `0x0` sentinel, while SDK 42 emits the approval against `chainMetadata.nativeToken.denom` for `amountIn`. This comparison is false, then the normal asset approval is compared with `igpAmount` and rejected. Normalize the trusted native asset contract before classifying asset vs fee, and cover a native-origin fixture.

paulbalaji (commented):
Reviewed exact head `66b2f38e2514bfcdfce91918d15afeec6d56c66a`. Two medium findings remain in the Starknet approval role/identity handling. Exact-head CI is green; locally, the focused route-security suite passed 63/63, typecheck passed, lint passed with two unrelated existing warnings, and `git diff --check` passed. Please resolve the trusted native token before classifying approvals and reject every approval/revoke whose token is not the trusted bridge asset or fee token.

troykessler (approved):
Reviewed full change. Approval validation is bound to trusted tokens, router, and quoted amounts; targeted tests pass.

PR #1219, opened 2026-08-28
fix: handle RPCs without gas state overrides

paulbalaji:
## Summary

- detect EVM RPCs that reject the third `eth_estimateGas` state-override argument
- fall back to the quote-provided origin gas budget and current fee data
- preserve balance-independent SDK estimation on compatible RPCs and surface unrelated errors

## Root cause

WUI requests `ignoreSenderBalance: true` from SDK 42.0.0. For EVM transactions this sends `eth_estimateGas(transaction, "latest", stateOverride)`. KiiChain RPC accepts at most two arguments, so source-fee estimation fails before transfer validation.

Live probe against `https://json-rpc.kiivalidator.com` with the reported transaction:

- two arguments: success, `0x329da` gas
- three arguments with balance override: `-32602 too many arguments, want at most 2`

## Verification

- `pnpm vitest src/features/transfer/engine/sourceFee.test.ts --run` (13 passed)
- `pnpm test` (48 files, 412 tests passed)
- `pnpm typecheck`
- `pnpm lint` (passes with two pre-existing console warnings)
- `pnpm exec oxfmt --check src/features/transfer/engine/sourceFee.ts src/features/transfer/engine/sourceFee.test.ts`
- `git diff --check`

Validated with Node 26.6.0.

### DIFF
--- a/src/features/transfer/engine/sourceFee.test.ts
+++ b/src/features/transfer/engine/sourceFee.test.ts
@@ -164,6 +164,30 @@ describe('estimateRouteSourceFee', () => {
     ).resolves.toBe(0n);
   });
 
+  test('uses quoted gas when an EVM RPC rejects the balance override argument', async () => {
+    prepareRouteTransactionMock.mockResolvedValue(typedTransaction(ProviderType.EthersV5));
+    const rpcError = new Error('too many arguments, want at most 2');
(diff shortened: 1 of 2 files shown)


troykessler (approved):
tested manually, works now

PR #1222, opened 2026-08-28
fix: handle synthetic balance fee estimation

Xaroz:
## Summary

- fall back to the route's quoted EVM gas only when an `EvmHypSynthetic` or `EvmHypSyntheticRebase` source burn exceeds the sender balance
- recognize both OpenZeppelin v4's revert string and v5's `ERC20InsufficientBalance` custom error
- let transfer validation report the actual token deficit instead of showing a network-fee toast
- preserve existing estimation behavior for native, ERC-20 collateral, xERC20, owner-collateral/yield, and unrelated RPC or contract errors

## Root cause

`ignoreSenderBalance` gives the sender enough native currency through an `eth_estimateGas` state override, but it cannot change ERC-20 storage. Estimating a synthetic Warp transfer still executes `_burn`, so an amount above the sender's token balance reverts before form validation can report the balance error.

### DIFF
--- a/src/features/transfer/engine/sourceFee.ts
+++ b/src/features/transfer/engine/sourceFee.ts
@@ -8,6 +8,7 @@ import type { FeeBreakdown } from './types';
 
 const EVM_LIKE_MIN_ROUTE_GAS_UNITS = 600_000n;
 const EVM_LIKE_APPROVAL_GAS_UNITS = 55_000n;
+const ERC20_INSUFFICIENT_BALANCE_SELECTOR = '0xe450d38c';
 const NATIVE_ADDRESS = '0x0000000000000000000000000000000000000000';
 
 export function sourceFeeRouteKey(route: RouteResponse): string {
@@ -52,7 +53,7 @@ export async function estimateRouteSourceFee({
   chainName: string;
   sender: string;
   senderPubKey?: Promise<HexString | undefined> | HexString;
-  sourceTokenStandard?: string;
+  sourceTokenStandard?: TokenStandard;
   route: RouteResponse;
   approvalTransactionCount: number;
 }): Promise<bigint> {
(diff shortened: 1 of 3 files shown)


Xaroz:
@claude review

xeno097 (approved):

src/features/transfer/engine/sourceFee.ts
```diff
@@ -108,15 +110,22 @@ export async function estimateRouteSourceFee({
...
 
-function isUnsupportedStateOverrideError(error: unknown): boolean {
-  return formatError(error).toLowerCase().includes('too many arguments, want at most 2');
+function shouldUseQuotedGasFallback(error: unknown, sourceTokenStandard?: string): boolean {
+  const message = formatError(error).toLowerCase();
+  return (
+    message.includes('too many arguments, want at most 2') ||
+    // A native balance override cannot fund a synthetic token burn. Balance
+    // validation reports the actual token deficit after this fee fallback.
+    (sourceTokenStandard === TokenStandard.EvmHypSynthetic &&
```
paulbalaji: **Medium:** Include `EvmHypSyntheticRebase` in this fallback. Hyperlane's `HypERC4626._transferFromSender` converts assets to shares, then calls `HypERC20._transferFromSender`, which invokes OpenZeppelin 4.9.3 `_burn`; insufficient shares produce this exact `ERC20: burn amount exceeds balance` revert. The current negative test therefore preserves the same misleading network-fee failure for over-balance rebase-synthetic transfers. Accept both synthetic standards and flip that case to positive coverage, unless there is a route invariant not represented in the current contract path.

paulbalaji (commented):
Reviewed exact head `75e6c4a0a496a0c5f63b1c8d2e7b7a8af3184032`, including all existing reviews, PR comments, current CI, contract implementation, and both production callers. One medium completeness issue remains inline: rebase synthetics use the same balance-failing burn path but are explicitly excluded. No other actionable findings.

Validation: focused 19/19, full Vitest 418/418, TypeScript, Oxlint, Oxfmt, and diff checks pass; exact-head CI/E2E/Vercel checks are green. Posted `COMMENT`, not `REQUEST_CHANGES`, per request.

src/features/transfer/engine/sourceFee.ts
Xaroz: Fixed in `6eb9e6f0`. I verified the `HypERC4626._transferFromSender` path converts assets to shares and delegates to `HypERC20._transferFromSender`, so the same balance-dependent `_burn` applies. The fallback now accepts both `EvmHypSynthetic` and `EvmHypSyntheticRebase`, with positive table-driven coverage for each; xERC20, collateral, native, and owner-collateral standards remain excluded.

Xaroz:
@claude @coderabbitai Resolved all review feedback in `6eb9e6f0`.

- Included `EvmHypSyntheticRebase` after verifying `HypERC4626._transferFromSender` delegates to the same balance-dependent `HypERC20` burn path.
- Added OpenZeppelin v5 `ERC20InsufficientBalance` support by inspecting nested revert data for selector `0xe450d38c`; the existing v4 revert-string path remains covered.
- Narrowed `sourceTokenStandard` to the SDK `TokenStandard` enum. The form only passes the two supported synthetic standards, keeping forward-compatible engine strings outside the estimator contract.
- Added the requested negative test proving an unrelated nested `All providers failed` error is rethrown unchanged.
- Preserved negative coverage for xERC20, collateral ERC-20, native, and owner-collateral/yield standards.

Validation completed on the exact pushed head: focused tests 21/21, full Vitest 420/420, formatting, lint, typecheck, production build, and diff checks all pass. Lint reports only the two pre-existing mock-wallet console warnings.

PR #1225, opened 2026-08-31
feat: block transfers with high price impact

Xaroz:
## Summary

- calculate expected price impact from input and output USD values
- block Continue and Send when fresh minimum-output pricing implies a 10% or greater loss
- fail open when current pricing is unavailable, stale, or failed while retaining cached display values
- add unit coverage for the calculation, minimum-output threshold, and failed-refresh behavior
- pin Vercel install and build commands to the repository's pnpm 11.20.0

### DIFF
--- a/src/features/balances/utils.ts
+++ b/src/features/balances/utils.ts
@@ -65,8 +65,9 @@ export function getFeePercentage(totalFeesUsd: number, transferUsd: number): str
   return `${pct.toLocaleString('en-US', { minimumFractionDigits: 2, maximumFractionDigits: 2 })}%`;
 }
 
-// Signed value change from source to destination. Negative means value lost
-// to fees, swap slippage, or spread. Missing prices stay unknown.
+// Signed percentage change from source to destination. Negative means value
+// lost to fees, swap slippage, or spread. Returns null when either value is
+// missing, non-finite, or outside the domain of a meaningful percentage.
 export function getPriceImpactPercentage(
   inputUsd: number | null,
   outputUsd: number | null,
--- a/src/features/balances/utils.ts
+++ b/src/features/balances/utils.ts
@@ -1,5 +1,5 @@
 import type { MultiProtocolProvider } from '@hyperlane-xyz/sdk';
-import { fromWei, fromWeiRounded } from '@hyperlane-xyz/utils';
+import { fromWei, fromWeiRounded, isNullish } from '@hyperlane-xyz/utils';
 
 import type { BalanceToken } from './types';
 import { balanceTokenKey, getBalanceTokenKey } from './types';
@@ -72,8 +72,8 @@ export function getPriceImpactPercentage(
   outputUsd: number | null,
 ): number | null {
   if (
-    inputUsd == null ||
-    outputUsd == null ||
+    isNullish(inputUsd) ||
+    isNullish(outputUsd) ||
     !Number.isFinite(inputUsd) ||
     !Number.isFinite(outputUsd) ||
     inputUsd <= 0 ||
--- a/src/features/balances/utils.ts
+++ b/src/features/balances/utils.ts
@@ -65,6 +65,25 @@ export function getFeePercentage(totalFeesUsd: number, transferUsd: number): str
   return `${pct.toLocaleString('en-US', { minimumFractionDigits: 2, maximumFractionDigits: 2 })}%`;
 }
 
+// Signed value change from source to destination. Negative means value lost
+// to fees, swap slippage, or spread. Missing prices stay unknown.
+export function getPriceImpactPercentage(
+  inputUsd: number | null,
+  outputUsd: number | null,
+): number | null {
+  if (
+    inputUsd == null ||
+    outputUsd == null ||
+    !Number.isFinite(inputUsd) ||
+    !Number.isFinite(outputUsd) ||
+    inputUsd <= 0 ||
+    outputUsd < 0
+  ) {
+    return null;
+  }
+  return ((outputUsd - inputUsd) / inputUsd) * 100;
+}
+
 export function getNativeTokenDenom(
   multiProvider: MultiProtocolProvider,
   chainName: string | undefined,
--- a/src/features/tokens/useTokenPrice.test.ts
+++ b/src/features/tokens/useTokenPrice.test.ts
@@ -1,0 +1,35 @@
+import { afterEach, describe, expect, test, vi } from 'vitest';
+
+import { useStore } from '../store';
+import { getFreshTokenUsdValue } from './useTokenPrice';
+
+const NOW = 2_000_000_000_000;
+const initialTokenPrices = useStore.getState().tokenPrices;
+
+afterEach(() => {
+  useStore.setState({ tokenPrices: initialTokenPrices });
+  vi.restoreAllMocks();
+});
+
+describe('getFreshTokenUsdValue', () => {
+  test('returns the USD value for a fresh price', () => {
+    expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - 1_000 }, '3', NOW)).toBe(6);
+  });
+
+  test('returns null for a stale price', () => {
+    expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - 16 * 60_000 }, '3', NOW)).toBeNull();
+  });
+
+  test('returns null after a failed refresh retains the cached price', () => {
+    vi.spyOn(Date, 'now').mockReturnValue(NOW);
+    useStore.setState({
+      tokenPrices: { ethereum: { usd: 2, fetchedAt: NOW - 1_000 } },
+    });
+
+    useStore.getState().mergeTokenPrices([], {}, ['ethereum']);
+
+    const entry = useStore.getState().tokenPrices.ethereum;
+    expect(entry).toEqual({ usd: 2, fetchedAt: NOW - 1_000, failedAt: NOW });
+    expect(getFreshTokenUsdValue(entry, '3', NOW)).toBeNull();
+  });
+});
--- a/src/features/tokens/useTokenPrice.test.ts
+++ b/src/features/tokens/useTokenPrice.test.ts
@@ -1,13 +1,14 @@
 import { afterEach, describe, expect, test, vi } from 'vitest';
 
 import { useStore } from '../store';
-import { getFreshTokenUsdValue } from './useTokenPrice';
+import { getFreshTokenUsdValue, schedulePriceExpiry } from './useTokenPrice';
 
 const NOW = 2_000_000_000_000;
 const initialTokenPrices = useStore.getState().tokenPrices;
 
 afterEach(() => {
   useStore.setState({ tokenPrices: initialTokenPrices });
+  vi.useRealTimers();
   vi.restoreAllMocks();
 });
 
@@ -31,5 +32,21 @@ describe('getFreshTokenUsdValue', () => {
     const entry = useStore.getState().tokenPrices.ethereum;
     expect(entry).toEqual({ usd: 2, fetchedAt: NOW - 1_000, failedAt: NOW });
     expect(getFreshTokenUsdValue(entry, '3', NOW)).toBeNull();
+  });
+
+  test('expires a fresh value without a price-entry mutation', () => {
+    vi.useFakeTimers();
+    vi.setSystemTime(NOW);
+    const entry = { usd: 2, fetchedAt: NOW };
+    let value = getFreshTokenUsdValue(entry, '3');
+
+    schedulePriceExpiry(entry.fetchedAt, () => {
+      value = getFreshTokenUsdValue(entry, '3');
+    });
+    vi.advanceTimersByTime(15 * 60_000 - 1);
+    expect(value).toBe(6);
+
+    vi.advanceTimersByTime(1);
+    expect(value).toBeNull();
   });
 });
--- a/src/features/tokens/useTokenPrice.test.ts
+++ b/src/features/tokens/useTokenPrice.test.ts
@@ -1,13 +1,10 @@
 import { afterEach, describe, expect, test, vi } from 'vitest';
 
-import { useStore } from '../store';
-import { getFreshTokenUsdValue, schedulePriceExpiry } from './useTokenPrice';
+import { getFreshTokenUsdValue, PRICE_CACHE_MS, schedulePriceExpiry } from './useTokenPrice';
 
 const NOW = 2_000_000_000_000;
-const initialTokenPrices = useStore.getState().tokenPrices;
 
 afterEach(() => {
-  useStore.setState({ tokenPrices: initialTokenPrices });
   vi.useRealTimers();
   vi.restoreAllMocks();
 });
@@ -17,21 +14,32 @@ describe('getFreshTokenUsdValue', () => {
     expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - 1_000 }, '3', NOW)).toBe(6);
   });
 
-  test('returns null for a stale price', () => {
-    expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - 16 * 60_000 }, '3', NOW)).toBeNull();
+  test('expires exactly at the freshness boundary', () => {
+    expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - PRICE_CACHE_MS + 1 }, '3', NOW)).toBe(
+      6,
+    );
+    expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - PRICE_CACHE_MS }, '3', NOW)).toBeNull();
   });
 
-  test('returns null after a failed refresh retains the cached price', () => {
-    vi.spyOn(Date, 'now').mockReturnValue(NOW);
-    useStore.setState({
-      tokenPrices: { ethereum: { usd: 2, fetchedAt: NOW - 1_000 } },
-    });
+  test('rejects a failed refresh and allows a later successful refresh', () => {
+    expect(
+      getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - 1_000, failedAt: NOW - 1_000 }, '3', NOW),
+    ).toBeNull();
+    expect(
+      getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - 1_000, failedAt: NOW - 90_000 }, '3', NOW),
+    ).toBe(6);
+  });
 
-    useStore.getState().mergeTokenPrices([], {}, ['ethereum']);
+  test('returns null when the price entry or price is unavailable', () => {
+    expect(getFreshTokenUsdValue(undefined, '3', NOW)).toBeNull();
+    expect(getFreshTokenUsdValue({ fetchedAt: NOW - 1_000 }, '3', NOW)).toBeNull();
+    expect(getFreshTokenUsdValue({ usd: 0, fetchedAt: NOW - 1_000 }, '3', NOW)).toBeNull();
+  });
 
-    const entry = useStore.getState().tokenPrices.ethereum;
-    expect(entry).toEqual({ usd: 2, fetchedAt: NOW - 1_000, failedAt: NOW });
-    expect(getFreshTokenUsdValue(entry, '3', NOW)).toBeNull();
+  test('prices amounts containing grouping separators', () => {
+    expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - 1_000 }, '1,234.56', NOW)).toBe(
+      2_469.12,
+    );
   });
 
   test('expires a fresh value without a price-entry mutation', () => {
@@ -43,7 +51,7 @@ describe('getFreshTokenUsdValue', () => {
     schedulePriceExpiry(entry.fetchedAt, () => {
       value = getFreshTokenUsdValue(entry, '3');
     });
-    vi.advanceTimersByTime(15 * 60_000 - 1);
+    vi.advanceTimersByTime(PRICE_CACHE_MS - 1);
     expect(value).toBe(6);
 
     vi.advanceTimersByTime(1);
--- a/src/features/tokens/useTokenPrice.tsx
+++ b/src/features/tokens/useTokenPrice.tsx
@@ -1,7 +1,7 @@
 import { isNullish } from '@hyperlane-xyz/utils';
 import { useDebounce } from '@hyperlane-xyz/widgets';
 import { useQuery } from '@tanstack/react-query';
-import { useMemo } from 'react';
+import { useEffect, useMemo, useReducer } from 'react';
 
 import { logger } from '../../utils/logger';
 import { useStore } from '../store';
@@ -149,7 +149,14 @@ export function useTokenUsdValue(token: UiToken | undefined, amount: string): nu
 export function useFreshTokenUsdValue(token: UiToken | undefined, amount: string): number | null {
   const id = token?.coinGeckoId;
   const entry = useStore((s) => (id ? s.tokenPrices[id] : undefined));
-  return useMemo(() => getFreshTokenUsdValue(entry, amount), [amount, entry]);
+  const [, invalidateFreshness] = useReducer((version: number) => version + 1, 0);
+
+  useEffect(
+    () => schedulePriceExpiry(entry?.fetchedAt, invalidateFreshness),
+    [entry?.fetchedAt, invalidateFreshness],
+  );
+
+  return getFreshTokenUsdValue(entry, amount);
 }
 
 function getTokenUsdValue(price: number | undefined, amount: string): number | null {
@@ -167,4 +174,16 @@ export function getFreshTokenUsdValue(
   if (isNullish(entry?.fetchedAt) || now - entry.fetchedAt >= PRICE_CACHE_MS) return null;
   if (!isNullish(entry.failedAt) && entry.failedAt >= entry.fetchedAt) return null;
   return getTokenUsdValue(entry.usd, amount);
+}
+
+export function schedulePriceExpiry(
+  fetchedAt: number | undefined,
+  onExpire: () => void,
+  now = Date.now(),
+): (() => void) | undefined {
+  if (isNullish(fetchedAt)) return undefined;
+  const delay = fetchedAt + PRICE_CACHE_MS - now;
+  if (delay <= 0) return undefined;
+  const timeoutId = setTimeout(onExpire, delay);
+  return () => clearTimeout(timeoutId);
 }
--- a/src/features/tokens/useTokenPrice.tsx
+++ b/src/features/tokens/useTokenPrice.tsx
@@ -4,23 +4,16 @@ import { useQuery } from '@tanstack/react-query';
 import { useEffect, useMemo, useReducer } from 'react';
 
 import { logger } from '../../utils/logger';
-import { useStore } from '../store';
+import { type TokenPriceCache, type TokenPriceCacheEntry, useStore } from '../store';
 import type { UiToken } from './types';
 
-interface PriceCacheEntry {
-  usd?: number;
-  fetchedAt?: number;
-  failedAt?: number;
-}
-
-type PriceCache = Record<string, PriceCacheEntry>;
 type MergeFn = (
   succeededIds: string[],
   fetched: Record<string, number>,
   failedIds: string[],
 ) => void;
 
-const PRICE_CACHE_MS = 15 * 60_000;
+export const PRICE_CACHE_MS = 15 * 60_000;
 const FAILED_BACKOFF_MS = 90_000;
 const COINGECKO_BATCH = 100;
 const CHUNK_DELAY_MS = 200;
@@ -70,7 +63,11 @@ async function fetchPricesBatched(ids: string[]): Promise<BatchedFetchResult> {
   return { requestedIds, prices };
 }
 
-async function fetchTokenPrices(ids: string[], cache: PriceCache, merge: MergeFn): Promise<null> {
+async function fetchTokenPrices(
+  ids: string[],
+  cache: TokenPriceCache,
+  merge: MergeFn,
+): Promise<null> {
   const now = Date.now();
   const idsToFetch = ids.filter((id) => {
     const entry = cache[id];
@@ -162,12 +159,20 @@ export function useFreshTokenUsdValue(token: UiToken | undefined, amount: string
 function getTokenUsdValue(price: number | undefined, amount: string): number | null {
   // Strip grouping separators — parseFloat("1,234.56") returns 1 otherwise.
   const parsedAmount = parseFloat(String(amount ?? '').replace(/,/g, ''));
-  if (!price || !Number.isFinite(parsedAmount)) return null;
+  if (
+    isNullish(price) ||
+    !Number.isFinite(price) ||
+    price <= 0 ||
+    !Number.isFinite(parsedAmount) ||
+    parsedAmount < 0
+  ) {
+    return null;
+  }
   return parsedAmount * price;
 }
 
 export function getFreshTokenUsdValue(
-  entry: PriceCacheEntry | undefined,
+  entry: TokenPriceCacheEntry | undefined,
   amount: string,
   now = Date.now(),
 ): number | null {
--- a/src/features/tokens/useTokenPrice.tsx
+++ b/src/features/tokens/useTokenPrice.tsx
@@ -1,3 +1,4 @@
+import { isNullish } from '@hyperlane-xyz/utils';
 import { useDebounce } from '@hyperlane-xyz/widgets';
 import { useQuery } from '@tanstack/react-query';
 import { useMemo } from 'react';
@@ -6,7 +7,13 @@ import { logger } from '../../utils/logger';
 import { useStore } from '../store';
 import type { UiToken } from './types';
 
-type PriceCache = Record<string, { usd?: number; fetchedAt?: number; failedAt?: number }>;
+interface PriceCacheEntry {
+  usd?: number;
+  fetchedAt?: number;
+  failedAt?: number;
+}
+
+type PriceCache = Record<string, PriceCacheEntry>;
 type MergeFn = (
   succeededIds: string[],
   fetched: Record<string, number>,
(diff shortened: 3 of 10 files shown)


src/features/transfer/engine/TransferForm.tsx
```diff
@@ -197,6 +203,17 @@ function TransferFormContent() {
 
   // Input USD — used by the fee % calculation in FeeSectionButton.
   const amountUsd = useTokenUsdValue(srcToken, values.amount);
+  const outputAmount = useMemo(() => {
+    if (!bestRoute || !dstToken) return '';
+    try {
+      return formatUnits(BigInt(bestRoute.raw.output), dstToken.decimals);
```
paulbalaji: [P1] Gate on the executable minimum, not the optimistic quote. This prices `bestRoute.raw.output`, but swap transactions may settle at `outputMin`. For example, a $100 input quoted at $91 passes this 10% gate; the existing 3% slippage preset permits an $88.27 minimum, so the transaction can legitimately lose 11.73%. That bypasses the stated loss cap. Please calculate the gating USD value from `outputMin` (or the worse of `output`/`outputMin`), while keeping `output` for the expected-value display if desired, and cover the crossing-threshold case.

src/features/transfer/engine/TransferForm.tsx
```diff
@@ -197,6 +203,17 @@ function TransferFormContent() {
...
+    } catch {
+      return '';
+    }
+  }, [bestRoute, dstToken]);
+  const outputUsd = useTokenUsdValue(dstToken, outputAmount);
```
paulbalaji, line 224: [P1] Avoid hard-gating on indefinitely stale prices. `useTokenUsdValue` returns any cached `usd` without checking `fetchedAt`; when a CoinGecko refresh fails, `mergeTokenPrices` deliberately retains the old value and only stamps `failedAt`. The new gate can therefore block or allow transfers using arbitrarily stale relative prices, despite the stated fail-open behavior when pricing is unavailable. Require fresh timestamps for enforcement (stale values can remain display-only) and test the failed-refresh path.

paulbalaji (commented):
Reviewed exact head `42dbcb00`. Two value-safety findings inline. Local validation passed: 426/426 tests, typecheck, lint, formatting, and diff check. Exact-head CI build/test/typecheck/lint/format/security/smoke/E2E passed; the Vercel template preview remains failed, and its logs were not available under my Vercel account.

src/features/transfer/engine/TransferForm.tsx
Xaroz: Fixed in f3507798. The gate now prices `outputMin`, while `output` remains the expected-value display; added the crossing-threshold test.

src/features/transfer/engine/TransferForm.tsx
Xaroz: Fixed in f3507798. Enforcement now requires fresh `fetchedAt` data and fails open after staleness or a failed refresh; cached prices remain display-only, with merge-path coverage.

paulbalaji (commented):
Rereviewed exact head `f3507798`. Both prior P1s are materially fixed: enforcement uses `outputMin`, failed/stale price entries fail open, and the Vercel preview is green. One remaining time-expiry issue is inline. Local validation passed: 430/430 tests, 39 focused tests, typecheck, lint, formatting, production build, and diff check. Exact-head CI/E2E/security/Socket/Vercel are green.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -134,10 +141,30 @@ export function useTokenPrices(): { prices: Record<string, number>; isLoading: b
...
-  }, [amount, price]);
+  return useMemo(() => getTokenUsdValue(price, amount), [amount, price]);
+}
+
+// Enforcement must fail open when the cached price is stale or its latest
+// refresh failed. Display-only callers can continue using useTokenUsdValue.
+export function useFreshTokenUsdValue(token: UiToken | undefined, amount: string): number | null {
+  const id = token?.coinGeckoId;
+  const entry = useStore((s) => (id ? s.tokenPrices[id] : undefined));
+  return useMemo(() => getFreshTokenUsdValue(entry, amount), [amount, entry]);
```
paulbalaji: [P1] Make freshness expire with wall-clock time. This memo only depends on `amount` and `entry`, while `Date.now()` is hidden inside `getFreshTokenUsdValue`; crossing the 15-minute boundary alone does not recompute it. Because price polling does not fetch while the page is unfocused and `refetchOnWindowFocus` is false, a tab suspended past the TTL can resume with the previously memoized non-null value and keep gating on stale prices until the entry or amount changes. Please schedule invalidation at `fetchedAt + PRICE_CACHE_MS`, or synchronously reread/evaluate freshness in both action guards, and add a fake-timer hook/action test for expiry without an entry mutation.

src/features/tokens/useTokenPrice.test.ts
```diff
@@ -0,0 +1,35 @@
...
+import { useStore } from '../store';
+import { getFreshTokenUsdValue } from './useTokenPrice';
+
+const NOW = 2_000_000_000_000;
+const initialTokenPrices = useStore.getState().tokenPrices;
+
+afterEach(() => {
+  useStore.setState({ tokenPrices: initialTokenPrices });
+  vi.restoreAllMocks();
+});
```
xeno097, line 10: **Blocker — this file fails 3/3 and makes `pnpm test` red.**

`useStore` is wrapped in zustand's `persist`, which resolves `window.localStorage` at store-creation time. `vitest.config.mts` sets no `environment`, so the run is `node`, and there is no `jsdom`/`happy-dom` in `devDependencies`. Every `useStore.setState` throws:

```
TypeError: Cannot read properties of undefined (reading 'setItem')
 ❯ src/features/tokens/useTokenPrice.test.ts:10:12   (afterEach)
 ❯ src/features/tokens/useTokenPrice.test.ts:25:14   (test 3)
```

Because this `afterEach` runs for every test, the two pure tests above (`returns the USD value for a fresh price`, `returns null for a stale price`) fail too — they never needed the store at all.

`src/features/store.test.ts` already establishes the working pattern: import only the pure reducers (`mergeKnownTokens`, `migratePersistedAppState`, …) and never touch the `useStore` singleton.

**Suggested fix:** drop the `useStore` import and this `afterEach`; keep tests 1–2 as direct calls to `getFreshTokenUsdValue`. For test 3, extract the `mergeTokenPrices` body as a pure exported reducer and assert it in `store.test.ts` alongside its siblings — the `{usd, fetchedAt, failedAt}` shape after a failed refresh is a store concern, not a `getFreshTokenUsdValue` one.

Two secondary issues in the same block: `vi.restoreAllMocks()` on line 11 is unreachable because line 10 throws first, so the `Date.now` mock leaks (harmless only because the mocking test happens to be last — swap the order). And `initialTokenPrices` is always `{}`, since `partialize` (store.ts:297) never persists `tokenPrices`.

src/features/transfer/engine/TransferForm.tsx
```diff
@@ -195,8 +205,25 @@ function TransferFormContent() {
...
 
-  // Input USD — used by the fee % calculation in FeeSectionButton.
+  // Cached prices remain display-only; fresh prices enforce the loss limit below.
   const amountUsd = useTokenUsdValue(srcToken, values.amount);
+  const [outputAmount, minimumOutputAmount] = useMemo<[string, string]>(() => {
+    if (!bestRoute || !dstToken) return ['', ''];
+    try {
+      return [
+        formatUnits(BigInt(bestRoute.raw.output), dstToken.decimals),
+        formatUnits(BigInt(bestRoute.raw.outputMin), dstToken.decimals),
```
xeno097: **The `outputMin` gate double-counts user-authorized slippage, so the slippage setting alone can block a zero-impact swap.**

To be clear about direction: gating on `outputMin` rather than the optimistic `output` was the right call and directly addresses the earlier P1 — I am not suggesting reverting it. The problem is that `outputMin` is not a market measurement. Per `useQuote.ts:538`, `compoundSlippageMin` makes it exactly `output × (1 − slippageBps/10000)^swapStepCount`, and `validateRouteAmounts` (:502-506) constrains `expectedMin <= outputMin <= output`. I verified against the fixture in `useQuote.test.ts:187-197` that the engine returns the floor to the wei: `313647743060936 / 320016062708842 = 0.99²` exactly.

So `minimumPriceImpactPct` *starts* at `−slippageBps/100 %` before a single cent of real impact:

- **1000 bps, 1 swap step** → `outputMin = 0.90 × output` → `−10` → `isPriceImpactTooHigh` uses `<=` → **blocked at zero real impact**. That is exactly `HIGH_SLIPPAGE_BPS` (`SlippagePanel.tsx:15`), a value the panel merely warns about.
- **600 bps × 2 legs** → `0.94² = 0.8836` → `−11.64%` → blocked at zero real impact.
- **3% preset × 2 legs** → `−5.91%` before fees; add a $2 IGP on a $60 transfer and you cross the cap.
- `MAX_SLIPPAGE_BPS = 5_000` (`SlippagePanel.tsx:14`), so roughly the upper 80% of a shipped control's range makes the form permanently unsendable on any pair, at any amount.

Pure-bridge routes are immune (`useQuote.ts:571` clamps `outputMin = output` for fixed-output), so this lands **only on swap routes** — the feature's entire target population. And the copy says "Price impact", pointing the user at the market rather than at the setting they just changed; the natural user response ("raise slippage to get my trade through") makes it strictly worse.

**Options that keep the `outputMin` gate intact:**
1. Subtract the authorized slippage before comparing: gate on `minimumPriceImpactPct + slippageBps/100 <= -MAX_PRICE_IMPACT_PCT`, so the cap measures *unauthorized* loss only.
2. Or cap `MAX_SLIPPAGE_BPS` strictly below `MAX_PRICE_IMPACT_PCT` so slippage alone can never reach the threshold.

Either way the message should name the real cause when slippage is the dominant term.

src/features/transfer/engine/TransferForm.tsx
```diff
@@ -195,8 +205,25 @@ function TransferFormContent() {
...
+      ];
+    } catch {
+      return ['', ''];
+    }
```
xeno097: **This silent `catch` now disables the safety guard, not just a display field.**

`['', '']` → `freshMinimumOutputUsd` is `null` → `minimumPriceImpactPct` is `null` → `isPriceImpactTooHigh(null)` is `false` → **transfer allowed**. A malformed `raw.outputMin` from the router API turns the protection off with no trace.

This exact pattern was harmless before this PR (the old `outputExact` memo only blanked a USD readout, and `outputDisplay` at :1127 still does), but the same swallow now gates value loss. A non-numeric `output` from the engine is a broken invariant, not expected user input — per CLAUDE.md, "NEVER add silent fallbacks for unexpected issues - they mask bugs."

Also note the shared `try` collapses both values: a bad `outputMin` also wipes `outputAmount`, so the card drops to `$0.00` and the impact percentage disappears at the same moment the guard goes off — the user loses the protection and the signal together.

**Suggested fix:** `catch (err) { logger.warn('Malformed route output amounts', err as Error); return ['', '']; }`. `logger` is already imported at line 26 and this file uses exactly that pattern at :406, :583, :937, :992.

src/features/transfer/engine/TransferForm.tsx
```diff
@@ -195,8 +205,25 @@ function TransferFormContent() {
...
+  }, [bestRoute, dstToken]);
+  const outputUsd = useTokenUsdValue(dstToken, outputAmount);
+  const priceImpactPct = getPriceImpactPercentage(amountUsd, outputUsd);
+  const freshAmountUsd = useFreshTokenUsdValue(srcToken, values.amount);
+  const freshMinimumOutputUsd = useFreshTokenUsdValue(dstToken, minimumOutputAmount);
+  const minimumPriceImpactPct = getPriceImpactPercentage(freshAmountUsd, freshMinimumOutputUsd);
+  const hasExcessivePriceImpact = isPriceImpactTooHigh(minimumPriceImpactPct);
```
xeno097, line 229: **The percentage shown to the user and the percentage that blocks them are different quantities, from different price sources, with nothing on screen reconciling them.**

- Displayed (`priceImpactPct`, rendered at :1171-1183): **cached** prices × `raw.output` (expected).
- Enforcing (`minimumPriceImpactPct`): **fresh** prices × `raw.outputMin` (worst case).

They diverge by the full slippage tolerance plus any cached/fresh divergence. `priceImpact.test.ts:22-27` encodes the gap as intended behavior (`-9%` passes, `-11.73%` blocks).

Concretely: the destination card renders `(-1.20%)` in plain grey — below the `-3%` danger colour at :1174 — while the button directly beneath reads `Price impact is 10% or higher` and is dead. Two authoritative-looking numbers ~200px apart, contradicting each other, with no explanation. The only conclusion available to a user is that the app is broken.

It runs the other way too: if the destination price entry is stale, `freshMinimumOutputUsd` is `null`, the block silently lifts, and the user can send while the card shows `-14%` in red — protection off exactly when the display is screaming.

**Suggested fix:** surface the enforced number rather than leaving it inferable only from a disabled button — e.g. a second line `worst case: -11.73%`, or interpolate the actual figure into the message (`Minimum received would lose 11.7%`). Also note the escalation ladder is split across two files with opposite sign conventions: `PRICE_IMPACT_WARN_PCT = -1` / `PRICE_IMPACT_DANGER_PCT = -3` here at :82-83 vs `MAX_PRICE_IMPACT_PCT = 10` in `priceImpact.ts`. Since the first two describe the displayed number and the third a different one, the visual escalation leads nowhere — red at -3%, then nothing until a block the red number never reaches.

xeno097 (commented):
## Review Summary

Multi-agent review of `f3507798` (six independent passes: numerical correctness, React/hooks, UX, adversarial bypass, test quality, scope/conventions). Findings below were cross-checked against each other; where two passes disagreed I re-verified against the code and report the resolution.

### The suite is red at this exact head

```
$ pnpm test
 Test Files  1 failed | 49 passed (50)
      Tests  3 failed | 427 passed (430)

TypeError: Cannot read properties of undefined (reading 'setItem')
 ❯ Object.setItem zustand/esm/middleware.mjs:270:42
 ❯ src/features/tokens/useTokenPrice.test.ts:10:12
```

All three tests in the new `useTokenPrice.test.ts` fail. Flagging explicitly because the prior review reported "430/430 tests" passing on this same SHA — the totals match (430) but 3 are failures here, so this may be an environment difference worth reconciling before merge. Reproduced twice locally, full details inline.

`pnpm typecheck` and `pnpm lint` are clean (2 pre-existing `no-console` warnings in `_e2e/`, untouched).

### The guard's inputs, not its wiring, are the problem

The plumbing is correct — every UI path to a broadcast consults `hasExcessivePriceImpact`, and the review→send race preserves `bestRoute` object identity, so there is no bypass. The issue is that the verdict collapses to "allow" in the cases that matter most and to "block" in cases that are perfectly legitimate:

- **Blocks legitimate transfers**: gating on `outputMin` means the user's own slippage setting counts against the 10% cap before any real market impact (inline on `TransferForm.tsx:215`). At `HIGH_SLIPPAGE_BPS` (1000) a zero-impact swap hits exactly `-10` and is blocked.
- **Allows risky ones**: every failure mode except literal-zero output fails open — no `coinGeckoId`, stale entry, `failedAt >= fetchedAt`, CoinGecko 429, `price === 0`, non-finite. Long-tail tokens with thin liquidity are the least likely to be CoinGecko-listed and the most likely to have genuine >10% impact, so coverage is inverted against risk.

### Reinforcing the open P1 on wall-clock freshness

`useTokenPrice.tsx:152` — agreed, and there is a second direction worth noting. Because `entry`'s reference only changes when `mergeTokenPrices` writes, and every write sets `fetchedAt: now`, at nearly every moment the memo *can* recompute `now - fetchedAt ≈ 0`. So the gate is not merely late — it can be **stuck ON** too: a memo that computed `<= -10%` keeps Continue/Send disabled until the next merge, with no way for the user to clear it. Whether you get stuck-open or stuck-closed depends on whether the user types after the price goes stale. Two review passes initially produced contradictory repros here; both are valid under different typing orders.

### UX: the block is a dead end

The message states a condition and offers no path out. The actions that would actually unblock a user — lower the amount, lower slippage, pick another route — are never suggested, and in review mode the slippage/routes controls are unmounted behind `{!isReview && (` at line 811. `FormWarningBanner` is already wired at :1029 and unused by this feature. Consider a banner naming the actual figure and the remedy, and instrumenting the block with `trackTransferValidationFailed` (fired on every other blocking condition but not this one) so you can measure false-positive rate rather than hearing about it in Discord.

### Observations Outside This PR

- `src/features/store.ts:262`: `mergeTokenPrices` writes `{usd: fetched[id], fetchedAt: now}` for every **requested** id, not every id CoinGecko actually returned. An id omitted from a 200 response caches as `{usd: undefined, fetchedAt: now}` with no `failedAt`, passes both freshness checks, and is not retried for 15 minutes. Pre-existing, but this PR makes it load-bearing for a safety guard.
- `src/features/tokens/useTokenPrice.tsx:105`: `refetchOnWindowFocus: false` plus react-query's default `refetchIntervalInBackground: false` means a backgrounded tab neither refetches on its interval nor catches up on return — this is what makes the staleness window in the open P1 unbounded rather than a 7.5-minute blip.
- `src/features/tokens/types.ts:37`: `decimals: t.decimals ?? 18` silently defaults a nullable API field. A 6-decimal token defaulted to 18 scales `outputMin` by 1e12 and produces `≈ -100%` impact — fail-closed, so not a safety hole, but an unexplainable dead end.
- `next.config.js:52`: `connect-src *` in an otherwise locked-down CSP. Unrelated to this PR; worth a separate ticket.

_Not re-flagging: CodeRabbit's nested-ternary nit on the Send label (1487-1491), and the two P1s already marked fixed in f3507798._

src/features/transfer/engine/TransferForm.tsx
```diff
@@ -536,6 +566,10 @@ function TransferFormContent() {
     if (!sender || !srcToken || !dstToken || !bestRoute || !values.srcChain || !values.dstChain) {
       return;
     }
+    if (hasExcessivePriceImpact) {
+      setIsReview(false);
+      return;
+    }
```
xeno097, line 604: **This guard is unreachable, and the failure it was written for goes unhandled.**

The Send button is `disabled` on the same `hasExcessivePriceImpact` from the same render (:1494), so a click cannot dispatch here. Same for the `onContinue` guard at :452 — `ConnectAwareSubmitButton` sets `type='button'` whenever disabled (`ConnectAwareSubmitButton.tsx:44-47`, "never submits when deliberately disabled"), so `submitCount` never advances and `FormSubmitDispatcher` (:1514) never fires. With multiple text inputs in the form, HTML implicit Enter-submission won't fire without a submit button either.

(Confirming the closure question, since it is easy to get wrong: `hasExcessivePriceImpact` **is** correctly present in both dep arrays, :508 and :717 — no stale read.)

So this is dead defensive code. Meanwhile the reachable failure gets nothing: `useQuote` is paused in review (`pause: isReview`, :183) but the price query is independent and keeps merging, so the guard can flip ON while the user sits on the review screen. The accent Send button silently turns grey and becomes an error string. No toast, no banner, no announcement — the only failure mode in this form that communicates purely by disabling a button, while `useToastError` is used for quote (:186), max-quote (:187), source-fee (:272) and transfer (:290) failures, and the sibling bailouts at :596 and :617 both set an error before collapsing.

**Suggested fix:** delete this block, and handle the false→true transition instead — fire a toast (`Price moved — this transfer now exceeds the 10% loss limit`) and drop out of review so the slippage and route controls remount.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -134,10 +141,30 @@ export function useTokenPrices(): { prices: Record<string, number>; isLoading: b
...
+}
```
xeno097, line 157: **"Cannot price" and "priced and safe" are the same value here, so coverage is inverted against risk.**

When `token.coinGeckoId` is undefined, `entry` is `undefined` → `getFreshTokenUsdValue` returns `null` → `getPriceImpactPercentage` returns `null` → `isPriceImpactTooHigh(null)` is `false` → **not blocked**. `coinGeckoId` is `z.string().optional()` in `TokenDiscoverySchema`, and the codebase already treats absence as routine (`FeeSectionButton.tsx:47-51` filters undefined ids; `utils.test.ts:81` has an explicit case).

The result: blue-chip pairs where real impact essentially never exceeds 10% get full protection, while `canSwap` long-tail tokens — least likely to be CoinGecko-listed, most likely to route through a thin pool at -40% — get none, silently, with an enabled Send button and no signal the guard doesn't apply.

The fail-open set is larger than just unlisted tokens: no cached entry, stale `fetchedAt`, `failedAt >= fetchedAt`, a CoinGecko 429 (free-tier `simple/price`, called direct from the browser with no key or proxy, one retry then silence for 90s), `price === 0`, and non-finite values all return `null`. `tokenPrices` is not persisted (`store.ts:297`), so enforcement is also off for the first seconds after every page load.

The fail-open default is a defensible design choice — the concern is that it is **invisible**. Rather than an alarming "protection unavailable" banner, the honest signal is at the display level: when impact can't be computed the card currently shows `$0.00` and omits the percentage (:1170-1171), which reads as "this transfer is worthless" rather than "we don't know". `Price unavailable` / `Impact unknown` would be accurate without being noisy.

**Suggested fix:** return a tri-state (`{status: 'ok' | 'unpriceable', pct}`) so callers must distinguish "verified safe" from "could not verify", and decide the policy explicitly at the call site instead of inheriting it from a `null`.

src/features/tokens/useTokenPrice.tsx
```diff
@@ -134,10 +141,30 @@ export function useTokenPrices(): { prices: Record<string, number>; isLoading: b
...
+  return parsedAmount * price;
+}
+
+export function getFreshTokenUsdValue(
+  entry: PriceCacheEntry | undefined,
+  amount: string,
+  now = Date.now(),
+): number | null {
+  if (isNullish(entry?.fetchedAt) || now - entry.fetchedAt >= PRICE_CACHE_MS) return null;
+  if (!isNullish(entry.failedAt) && entry.failedAt >= entry.fetchedAt) return null;
```
xeno097, line 180: **Both boundary comparisons here survive mutation, and the recovery branch is untested.**

I mutated the source and re-ran the suite (using a localStorage-independent copy of the assertions, since the file as committed doesn't execute — see the test-file comment):

| Mutation | Result |
|---|---|
| `now - entry.fetchedAt >= PRICE_CACHE_MS` → `>` | **survives** |
| `entry.failedAt >= entry.fetchedAt` → `>` | **survives** |
| `if (!price ...)` → `if (price == null ...)` | **survives** |
| remove `.replace(/,/g, '')` | **survives** |

Neither `>=` boundary is probed: the stale case uses `NOW - 16 * 60_000` (a full minute past the 15-minute line) and the failure case uses `failedAt: NOW` vs `fetchedAt: NOW - 1_000` (strictly greater). So `failedAt === fetchedAt` and exactly-at-`PRICE_CACHE_MS` are both unverified.

The **recovery** branch is the one that worries me most. When a fetch fails and a later one succeeds, `mergeTokenPrices` writes a fresh `fetchedAt` while the old `failedAt` remains in the entry, so `failedAt < fetchedAt` must return the value. Nothing asserts this. If someone later "simplifies" line 168 to `if (!isNullish(entry.failedAt)) return null`, every token that has ever had a single failed fetch is permanently frozen into fail-open — guard silently dead forever — and the suite stays green.

**Suggested tests:**
```ts
expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - PRICE_CACHE_MS }, '3', NOW)).toBeNull();
expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - PRICE_CACHE_MS + 1 }, '3', NOW)).toBe(6);
expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - 1_000, failedAt: NOW - 1_000 }, '3', NOW)).toBeNull();
expect(getFreshTokenUsdValue({ usd: 2, fetchedAt: NOW - 1_000, failedAt: NOW - 90_000 }, '3', NOW)).toBe(6);
expect(getFreshTokenUsdValue(undefined, '3', NOW)).toBeNull();
expect(getFreshTokenUsdValue({ fetchedAt: NOW - 1_000 }, '3', NOW)).toBeNull();
```
Export `PRICE_CACHE_MS` and write the boundary against it rather than hardcoding 16 minutes — I verified that lowering `PRICE_CACHE_MS` to 5 min leaves the current test passing vacuously.

Also worth covering here specifically: the comma-stripping mutation survives, and this is the **enforcement** path — a `'1,234.56'` misparse to `1` would make a large transfer look like a 99.9% loss.

src/features/balances/utils.ts
```diff
@@ -65,6 +65,25 @@ export function getFeePercentage(totalFeesUsd: number, transferUsd: number): str
...
+  inputUsd: number | null,
+  outputUsd: number | null,
+): number | null {
+  if (
+    isNullish(inputUsd) ||
+    isNullish(outputUsd) ||
+    !Number.isFinite(inputUsd) ||
+    !Number.isFinite(outputUsd) ||
+    inputUsd <= 0 ||
+    outputUsd < 0
```
xeno097, line 81: **Two of these four guards are provably dead to the test suite.**

Verified empirically against `utils.test.ts` + `priceImpact.test.ts` (baseline 13/13 passing):

| Mutation | Result |
|---|---|
| delete `!Number.isFinite(outputUsd)` | **13 passed — survives** |
| delete `outputUsd < 0` | **13 passed — survives** |
| `inputUsd <= 0` → `inputUsd < 0` | 1 failed — caught |

`utils.test.ts:135-137` exercises `null` and `Infinity` only in the **input** position. The output position — the side fed from `formatUnits(BigInt(bestRoute.raw.outputMin))`, i.e. remote API data — has no test at all. Delete either guard and `getPriceImpactPercentage(100, Infinity)` returns `Infinity` and `(100, -50)` returns `-150`, both flowing straight into `isPriceImpactTooHigh`.

```ts
test('returns null for non-finite or negative output', () => {
  expect(getPriceImpactPercentage(100, Number.POSITIVE_INFINITY)).toBeNull();
  expect(getPriceImpactPercentage(100, Number.NaN)).toBeNull();
  expect(getPriceImpactPercentage(100, -50)).toBeNull();
});
```

Two smaller notes on this function:

- The `inputUsd < 0` / `outputUsd < 0` branches are unreachable in production (`BigIntString` is `^\d+$`, `decimals >= 0`, and `getTokenUsdValue` already rejects falsy prices) — harmless, but they read as coverage that isn't doing anything.
- The comment at :68-69 and the test name `returns the signed USD value change` both describe a **dollar delta**; the function returns a **percentage**. And "Missing prices stay unknown" undersells it — `null` is also returned for non-finite and non-positive inputs, which `utils.test.ts:136` asserts under a heading claiming otherwise. Suggest: *"Returns null when either side is missing, non-finite, or non-positive — no meaningful percentage exists."*

src/features/transfer/engine/priceImpact.ts
```diff
@@ -0,0 +1,8 @@
+import { isNullish } from '@hyperlane-xyz/utils';
+
+export const MAX_PRICE_IMPACT_PCT = 10;
```
xeno097, line 6: **Hardcoded threshold with no override, in a template designed to be forked.**

`src/consts/config.ts` is the established home for exactly this shape of value — it already carries `defaultSlippageBps` via `parseFiniteEnvNumber('NEXT_PUBLIC_DEFAULT_SLIPPAGE_BPS', 100)` and `permit2ExpirationSeconds`, both numeric transfer-economics thresholds. CLAUDE.md and CUSTOMIZE.md both point forkers at `config.ts` for tuning. As written, a team deploying this for an illiquid long-tail asset has to patch source and carry a fork delta.

**Suggested fix:** `maxPriceImpactPct: parseFiniteEnvNumber('NEXT_PUBLIC_MAX_PRICE_IMPACT_PCT', 10)` in `config`, documented in `.env.example`. Note `PRICE_IMPACT_BLOCK_MESSAGE` must then become a function rather than a module-level template literal.

Worth deciding deliberately on the hard-block policy too, since there is no override path at all. In favour: users click through warnings, bridging is irreversible in a way a DEX swap isn't, and >10% is usually a bug rather than intent. Against: this is a permissionless bridge, legitimate reasons to accept >10% exist in the tail (exiting an illiquid position, escaping a depeg), and — most importantly — the measurement itself isn't reliable enough to justify irreversibility, since it compares two independently-sourced CoinGecko prices and folds in a slippage term the user chose.

A hard block is defensible; a hard block on *this* number with no escape hatch is harder to defend. If the measurement gets tightened (see the `outputMin` comment), a friction-gated override — typed confirmation, not a one-click "proceed" — would be the conventional resolution.

src/features/transfer/engine/priceImpact.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    expect(isPriceImpactTooHigh(null)).toBe(false);
+  });
+
+  test('blocks when the executable minimum crosses the threshold', () => {
+    const expectedImpact = getPriceImpactPercentage(100, 91);
+    const minimumImpact = getPriceImpactPercentage(100, 88.27);
+
+    expect(isPriceImpactTooHigh(expectedImpact)).toBe(false);
+    expect(isPriceImpactTooHigh(minimumImpact)).toBe(true);
+  });
```
xeno097: **This test asserts nothing the two tests above it don't already cover.**

`(91-100)/100*100 = -9` and `(88.27-100)/100*100 = -11.73` — both **further** from the `-10` boundary than the existing `-9.99 → false` and `-10 → true` assertions, so it is strictly weaker. The name promises "the executable minimum", but there is no minimum, no `outputMin`, no route, and no fresh-vs-cached distinction here — just `isPriceImpactTooHigh` called twice with hand-picked constants. `88.27` in particular looks reverse-engineered to produce the expected boolean.

I verified the redundancy: with this test deleted, the mutations `<=` → `<` and `MAX_PRICE_IMPACT_PCT = 10` → `12` are both still caught by the tests above. It also is the sole reason this file reaches across a feature boundary to import `getPriceImpactPercentage` from `../../balances/utils`.

The behavior it was reaching for — that the gate uses `outputMin` and not `output` — is the actual point of f3507798 and is currently **untested**, because that derivation lives inside the `useMemo` in `TransferForm.tsx:210-220`. Extracting it would make it testable:

```ts
test('blocks on the executable minimum even when the expected output is fine', () => {
  expect(isPriceImpactTooHigh(getPriceImpactPercentage(100, 95))).toBe(false);
  expect(isPriceImpactTooHigh(getPriceImpactPercentage(100, 90))).toBe(true);
});
```

More broadly: the guard **wiring** — that the flag disables the button and short-circuits the handlers — has no coverage at all, and the repo has the infrastructure for it. `src/components/buttons/ConnectAwareSubmitButton.test.tsx` renders a Formik-wrapped component with `renderToStaticMarkup` and asserts on both text and `disabled=""`. `ButtonSection` would need exporting (it's module-local at :1380), but that is the whole lift.

vercel.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "installCommand": "npx --yes pnpm@11.20.0 install --frozen-lockfile",
+  "buildCommand": "npx --yes pnpm@11.20.0 run build"
```
xeno097, line 4: **This is a real fix in the wrong PR, and it undercuts the repo's own supply-chain posture.**

On scope: a CI/deploy change in a branch titled "prevent high slippage", with an empty commit body and no linked issue. The underlying problem looks genuine — `packageManager` moved to pnpm 11 and `pnpm-workspace.yaml:38` uses `allowBuilds:`, a pnpm-11-only key (pnpm 10 spells it `onlyBuiltDependencies`), while Vercel infers its pnpm version from `lockfileVersion: '9.0'` → pnpm 9/10, so the native-build allowlist would silently not apply. But main has carried that state since the branch point with no fix, and the commit records no failing-deploy evidence. It belongs in its own PR where a reviewer can see what broke.

I verified the pin itself is safe: `npm view pnpm@11.20.0 version` → `11.20.0`, real and published, and it matches `package.json:129`. `pnpm run build` does still fire the `prebuild` font hook, so `buildCommand` isn't dropping fonts.

Two real problems remain:

1. **Integrity is discarded.** `packageManager` carries a `+sha512` hash; `npx --yes pnpm@11.20.0` fetches a tarball with no such verification. `pnpm-workspace.yaml:1-2` sets `minimumReleaseAge: 10080` explicitly labelled "supply chain protection" — so the repo deliberately hardens against fresh-package attacks, then pulls its package manager unverified on every build. The idiomatic Vercel fix is the `ENABLE_EXPERIMENTAL_COREPACK=1` project env var, which honours the existing hashed `packageManager` field and needs no `vercel.json` at all.
2. **The version is now in three places** (`package.json:129`, and both lines here) with no sync mechanism. The next dependency bump moves `packageManager` and these drift silently — the same class of bug this commit is fixing.

src/features/tokens/useTokenPrice.tsx
Xaroz: Fixed in `afa1f1a4`. `useFreshTokenUsdValue` now schedules a rerender at `fetchedAt + PRICE_CACHE_MS` and recomputes freshness from wall-clock time, so a suspended tab fails open when the timer runs on resume. Added a fake-timer regression test proving expiry without a price-entry mutation.

src/features/transfer/engine/priceImpact.ts
Xaroz: Keeping the 10% constant local for this PR: it is the explicitly requested policy, and adding a new public environment/configuration API would expand scope beyond this safety change.

src/features/tokens/useTokenPrice.tsx
Xaroz: Fixed in `8a2225fe`. Added exact TTL and failed-at boundaries, recovery, missing-entry/price, zero-price, grouped-amount, and timer-expiry coverage using the exported cache duration.

src/features/tokens/useTokenPrice.test.ts
Xaroz: Fixed in `8a2225fe`. The hook test no longer mutates the persisted Zustand store; cache merging is now a pure reducer with store-level failed-refresh and recovery tests.

src/features/tokens/useTokenPrice.tsx
Xaroz: Fixed in `8a2225fe` without expanding the hook API: the call site now distinguishes unavailable enforcement pricing and renders `Price unavailable` / `Minimum: unavailable`, while preserving the requested fail-open behavior.

src/features/balances/utils.ts
Xaroz: Fixed in `8a2225fe`. Added non-finite and negative output cases and corrected the function/test wording to describe a signed percentage and its full null domain.

src/features/transfer/engine/TransferForm.tsx
Xaroz: Fixed in `8a2225fe`. Swap routes now display the enforced minimum-value percentage beside expected impact, the disabled action includes the actual loss, and unavailable enforcement pricing is labeled explicitly.

src/features/transfer/engine/TransferForm.tsx
Xaroz: Keeping the executable-minimum math by design: the original requirement is a 10% worst-case value-loss cap, so user-authorized slippage must count. In `8a2225fe`, the UI now states and displays the actual minimum-received loss instead of presenting it as pure market price impact.

src/features/transfer/engine/TransferForm.tsx
Xaroz: Fixed in `8a2225fe`. A price crossing the limit during review now records validation failure, shows a form warning, and returns to edit mode so route/slippage controls remount. The action guards remain as defense in depth.

src/features/transfer/engine/TransferForm.tsx
Xaroz: Fixed in `8a2225fe`. Removed the silent fallback; malformed schema-validated route output now fails loudly, with regression coverage.

src/features/transfer/engine/priceImpact.test.ts
Xaroz: Fixed in `8a2225fe`. Replaced the hand-picked-value test with route-output derivation from `output` and `outputMin`, then verifies the minimum crosses the gate and produces the user-facing loss message.

paulbalaji (approved):
Rereviewed exact head `8a2225fe`. No blocking findings. The wall-clock freshness issue is fixed, and the latest follow-up cleanly addresses the current cache, malformed-output, test, and loss-UX feedback while preserving the intended 10% worst-case value-loss cap. Local validation passed: 438/438 tests, 57 focused tests, typecheck, lint, formatting, production build, and diff check. Exact-head CI/E2E/security/Socket/Vercel checks have no pending or failing jobs.

vercel.json
paulbalaji: this is fine, i do this in my pnpm 12 pr

xeno097 (approved):

PR #1226, opened 2026-09-01
feat: support hiding configured warp routes

paulbalaji:
## Summary

- add an empty, configurable UI-only Warp route denylist
- hide denied-only tokens and available-route results
- remove denied routes and rejections from normal and max quotes
- validate both route connection and bridge-step IDs
- pin Vercel install/build commands to pnpm 11.20.0

This adds generic presentation gating only. No route is denied on `main`; URE and on-chain contracts remain public.

## Validation

- `pnpm test` (423 passed before final hardening)
- `pnpm vitest src/features/api/RouterClient.test.ts` (11 passed after final hardening)
- `pnpm typecheck`
- `pnpm lint` (2 existing console warnings)
- `pnpm exec oxfmt --check src/consts/config.ts src/features/api/RouterClient.ts src/features/api/RouterClient.test.ts vercel.json`

### DIFF
--- a/src/consts/config.ts
+++ b/src/consts/config.ts
@@ -45,6 +45,7 @@ interface Config {
   version: string; // App release version shown in metadata/analytics
   walletConnectProjectId: string; // Project ID provided by walletconnect
   walletProtocols: ProtocolType[] | undefined; // Wallet Protocols to show in the wallet connect modal. Leave undefined to include all of them
+  warpRouteDenylist: string[]; // Route IDs hidden from this UI only; the router API and contracts remain public
   rpcOverrides: string; // JSON string containing a map of chain names to an object with an URL for RPC overrides (For an example check the .env.example file)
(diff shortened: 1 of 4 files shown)


haggis-hyperlane (approved):
LGTM. UI-only presentation gating, cleanly scoped.

- Field shapes verified against zod schemas: `warpRouteIds` always present; `connection` nullable (handled via `?.`); `rejections` optional (guarded).
- `isDeniedToken` correctly hides tokens only when *all* their routes are denied, so shared tokens stay visible.
- `quote`/`maxQuote` filter both routes and their rejections.
- Denylist empty by default → no behavior change on `main`; URE + contracts remain public.

Good test coverage for the denied-only, shared-token, and rejection cases.

Source Haggis thread: https://haggis.services.hyperlane.xyz/threads/01M1D9E6K5CGY8YBWBN689HTBQ

PR #1228, opened 2026-09-01
feat: hide NES route on Nexus

paulbalaji:
Adds the NES/bsc Warp route to the Nexus UI denylist. The route remains available through URE and onchain; Nexus omits it from tokens, available routes, quotes, and quote rejections.

### DIFF
--- a/src/consts/config.ts
+++ b/src/consts/config.ts
@@ -84,7 +84,7 @@ export const config: Config = Object.freeze({
     ProtocolType.Tron,
     ProtocolType.Aleo,
   ],
-  warpRouteDenylist: [],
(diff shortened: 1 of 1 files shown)


haggis-hyperlane (approved):
LGTM. Populates the nexus `warpRouteDenylist` with `NES/bsc`, activating the UI-only hiding wired up in the denylist code. `shouldDisableChains: true` is correctly preserved. Route stays public via URE + onchain, as intended.

Source Haggis thread: https://haggis.services.hyperlane.xyz/threads/01M1D9E6K5CGY8YBWBN689HTBQ

